PR #4, opened 2022-01-12
Added examplify() which generates a String by example.

amitsol:
Added examplify() which generates a String by example. The output string will have the same pattern as the input string.


bodiam:
Thanks @amitsol for this PR. We're honored that this was your first PR ever, great work!

PR #7, opened 2022-02-04
Add support of methods with varargs and #{options.option ...}

snuyanzin:
There 2 issues:
1. Current code can not resolve the code with varargs e.g. `Options.option`
2. Options does not have `option` method with `String` varargs which is implied by `FakeValuesService`

This PR adds check if there is varargs in method args and supports it, also adds `Options.option` with String varargs to make it working like
``` 
#{options.option '12','345','89','54321'}
```

the PR is same as https://github.com/DiUS/java-faker/pull/703

### DIFF
--- a/src/main/java/net/datafaker/Options.java
+++ b/src/main/java/net/datafaker/Options.java
@@ -21,6 +21,16 @@ public class Options {
     }
 
     /**
+     * Returns a random String element from an varargs.
+     *
+     * @param options The varargs to take a random element from.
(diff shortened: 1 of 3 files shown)


bodiam (approved):

bodiam:
Hi @snuyanzin , thanks again. It took me a bit longer to review this one, since I've added similar code recently, though I added it to Faker itself. I'll move my code also to Options.

PR #9, opened 2022-02-04
Put dependency versions in properties

snuyanzin:
The PR extracts versions of dependencies into properties to simplify the process of dependency bumping as with this PR they are all in the same place


bodiam:
Nice one! Small question : do you have an IntelliJ license? If not, can I offer you one?

snuyanzin:
Thanks for the offer, currently I have it for open source projects

PR #13, opened 2022-02-07
 Enable build for other platforms and other jdks

snuyanzin:
The PR add a matrix with `windows-latest` and `macos-latest` and also LTS jdks (8, 11, 17) to make sure it could be built on any of them


bodiam:
Great work! I think there's a small test issue, but I think I fixed that yesterday.

PR #15, opened 2022-02-08
Add support of date time pattern for string representation of date an…

snuyanzin:
The problem is that calls like `#{date.past ... }` and others with `date` converts date and time based on default `toString` with no way to customize it.

The PR adds corresponding methods which allow to specify date/time pattern


bodiam:
Awesome. This pr reminds me to add Java 8 time support instead of the old Java Date.

PR #17, opened 2022-02-09
Remove .travis.yml

snuyanzin:
Since datafaker uses GitHub actions `.travis.yml` looks as a useless file


bodiam:
Cheers! I think we naar need it to get some code coverage info, not sure about that

PR #22, opened 2022-02-10
Add missing md2 and sha384

snuyanzin:
It looks like there are missing `md2` and `sha384`. The PR adds support for them

If there is no objections I'll merge it soon


bodiam:
Thanks! I don't think I've even heard of them!

PR #23, opened 2022-02-10
Add templatify function to fill templates

snuyanzin:
There are already present several functions to generate a string by template e.g.
`net.datafaker.Faker#letterify` or `net.datafaker.Faker#numerify`.

The problem is that `letterify` could fill template only with ascii symbols and `numerify` only with digits and it is impossible to fill one `?` with several symbols.

The idea of this PR is providing of `xrify` (not sure about naming, probably there might be better options) which takes a template and a set of string options which could be used to fill it e.g.

E.g. if I want to fill `?` with some Russian or Chinese strings
```
faker.templatify("12??34", '?', "тест", "test", "测试测试")
```

### DIFF
--- a/src/main/java/net/datafaker/Faker.java
+++ b/src/main/java/net/datafaker/Faker.java
@@ -4,6 +4,7 @@ import net.datafaker.service.FakeValuesService;
 import net.datafaker.service.RandomService;
 
 import java.util.Locale;
+import java.util.Map;
 import java.util.Random;
 
 /**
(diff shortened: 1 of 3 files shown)


bodiam:
Cheers! I'm not a big fan of the name, I  wouldn't remember what this method does. How about letting go of that "ify" postfix, and call it "template" (templatify?) or so instead?

snuyanzin:
makes sense, thanks, i will rename it

bodiam:
Looks good, but the documentation is not entirely correct here:

 https://github.com/datafaker-net/datafaker/pull/23/files#diff-9ada76c37e7f8e25858b7a3fbc87901093393a8a18e5a4ee6f32669d5240d88bR388

It doesn't replace?, it replaces the character you specify, is it not?

snuyanzin:
oops, you're right, thanks for the catch
will add a correction for that

bodiam:
Also, there are some <p> characters in the javadoc. I don't think they add much, but I think they might break the javadoc generation, not 100% sure. I had some javadoc issues before, just not 100% sure if this was the same issue.

bodiam:
Great PR btw, my comments is just minor feedback, and happy to fix myself

PR #32, opened 2022-02-13
Refactor EnFile to speed it up

snuyanzin:
The PR refactors `EnFile` to speed it up

benchmark before
```
Benchmark           Mode  Cnt     Score   Error   Units
JmhTest.initFaker  thrpt    5    44.192 ± 0.733  ops/ms
```
and after
```
Benchmark           Mode  Cnt    Score   Error   Units
JmhTest.initFaker  thrpt    5  138.761 ± 2.593  ops/ms
```

benchmark'c code
```java
@Benchmark
public void initFaker(Blackhole blackhole) {
    blackhole.consume(new Faker());
}
```

### DIFF
--- a/src/main/java/net/datafaker/service/FakeValuesService.java
+++ b/src/main/java/net/datafaker/service/FakeValuesService.java
@@ -29,7 +29,7 @@ public class FakeValuesService {
     private static final Pattern A_TO_Z = Pattern.compile("([A-Z])");
     private static final Pattern UNDERSCORE = Pattern.compile("_");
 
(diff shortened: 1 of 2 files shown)


bodiam:
Nice!!

bodiam:
For naming conventions, I don't mind that much, but I use this as my guide:

https://stackoverflow.com/questions/1417190/should-a-static-final-logger-be-declared-in-upper-case

snuyanzin:
Ok, thanks for letting know 
in fact the main thing I did here with logger is making it static since it was in profiler's output

PR #34, opened 2022-02-13
Add .gitattributes

snuyanzin:
The PR adds `.gitattributes` which simplifies handling of lf/cr normalization across different platforms.


epragtbeamtree:
Nice. Should yml also be included here maybe? Or is there no need?

snuyanzin:
yes, makes sense, added yaml as well

PR #36, opened 2022-02-14
Use static faker as global faker in tests and method local as faker w…

snuyanzin:
This will speed up tests.

Currently faker is initialized before each test and huge amount of time is spent for yaml parsing. 
The PR makes `faker` static and initialized `beforeclass` to cope with this


snuyanzin:
merged as dc3fdac1770258a4ffed838f93d4282e544fe609

PR #47, opened 2022-02-16
Add csv generation

snuyanzin:
The PR adds support for csv generation
number and names of columns could be specified
separator and quotes could be specified with `separator()` and `quote()`
number of lines could be specified via `limit()`
with or without header also could be specified with `header()`
e.g. 
```
System.out.println(new Csv.CsvBuilder()
                .columns(Csv.Column.of("first_name", () -> faker.name().firstName()),
                        Csv.Column.of("last_name", () -> faker.name().lastName()),
                        Csv.Column.of("address", () -> faker.address().streetAddress()))
                .header(true)
                .separator(" ; ")
                .limit(5).build().get());
```

will generate something like that
```
"first_name" ; "last_name" ; "address"
"Jonah" ; "Kovacek" ; "009 Wilkinson Summit"
"John" ; "Murphy" ; "379 McCullough Locks"
"Colby" ; "Bins" ; "93534 Stevie Gardens"
"Wade" ; "Herzog" ; "83108 Willy Road"
"Marg" ; "Effertz" ; "415 Gene Plaza"
```

### DIFF
--- a/src/main/java/net/datafaker/Csv.java
+++ b/src/main/java/net/datafaker/Csv.java
@@ -1,0 +1,119 @@
+package net.datafaker;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+public class Csv {
+    private final String separator;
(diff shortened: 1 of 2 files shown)


bodiam:
Love it! Really nice!

PR #54, opened 2022-02-20
feat: add PESEL number generator

zolv:
This is a generator of a [Polish PESEL number](https://en.wikipedia.org/wiki/PESEL).

### DIFF
--- a/src/main/java/net/datafaker/IdNumber.java
+++ b/src/main/java/net/datafaker/IdNumber.java
@@ -14,132 +14,134 @@ import net.datafaker.idnumbers.SvSEIdNumber;
 import net.datafaker.idnumbers.ZhCnIdNumber;
 
 public class IdNumber {
-  private final Faker faker;
+    private final Faker faker;
 
-  protected IdNumber(Faker faker) {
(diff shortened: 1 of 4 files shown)


bodiam:
Hi @zolv , thank you for your PR, great work!

And I'm sorry, but there's a really small issue with it:

```
 Error:  /home/runner/work/datafaker/datafaker/src/main/java/net/datafaker/IdNumber.java:141: error: reference not found
Error:       * @param gender    Person's gender. Null value means {@link Gender.ANY}
```

It seems the Javadoc generation is a little strict, and it doesn't seem to be able to find the Gender. It might need a fully qualified name here. I'm happy to fix it, but I'm not 100% sure what will happen to the PR then, so if you could fix it, that might be better, and I'm happy to accept it! 

Thanks again!

bodiam:
Hi @zolv , thanks for your PR, great work. There was a small issue with the javadoc generation `{@link Gender.ANY}` needs to be `{@link Gender#ANY}`, but I've fixed it. Again, thanks for your contribution, we really appreciate it!

PR #55, opened 2022-02-21
Use locales, charsets wherever possible

snuyanzin:
The PR forces to specify locales, charsets and forbids usage of some apis e.g. `System.gc`, `Runtime.exit` and etc. the full list is at `src/main/config/forbidden-apis/signatures.txt`

### DIFF
--- a/src/main/java/net/datafaker/DateAndTime.java
+++ b/src/main/java/net/datafaker/DateAndTime.java
@@ -1,9 +1,13 @@
 package net.datafaker;
 
 import java.text.SimpleDateFormat;
-import java.util.Calendar;
+import java.time.Clock;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoField;
+import java.time.temporal.ChronoUnit;
 import java.util.Date;
-import java.util.GregorianCalendar;
+import java.util.Locale;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -44,7 +48,7 @@ public class DateAndTime {
      * @return a string representation of a future date from now.
      */
     public String future(int atMost, TimeUnit unit, String pattern) {
-        return toString(future(atMost, unit), pattern);
+        return toString(future(atMost, unit), pattern, faker.getLocale());
     }
 
     /**
@@ -72,7 +76,7 @@ public class DateAndTime {
      * @return a string representation of a future date from now, with a minimum time.
      */
     public String future(int atMost, int minimum, TimeUnit unit, String pattern) {
-        return toString(future(atMost, minimum, unit), pattern);
+        return toString(future(atMost, minimum, unit), pattern, faker.getLocale());
     }
 
     /**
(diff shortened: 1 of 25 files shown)


src/main/java/net/datafaker/Internet.java
```diff
@@ -179,7 +179,7 @@ public String macAddress() {
      * @return a correctly formatted IPv4 address.
      */
     public String ipV4Address() {
-        return String.format("%d.%d.%d.%d",
+        return String.format(faker.getLocale(), "%d.%d.%d.%d",
```
bodiam, line 182: I don't understand why you'd want to use the locale to format an ip address.

src/main/java/net/datafaker/DateAndTime.java
```diff
@@ -220,15 +239,15 @@ public Date between(Date from, Date to) throws IllegalArgumentException {
...
+        return toString(between(from, to), pattern, faker.getLocale());
     }
 
     /**
      * Generates a random birthday between 65 and 18 years ago from now.
      *
      * @return a random birthday between 65 and 18 years ago from now.
      */
-    public Date birthday() {
+    public LocalDate birthday() {
```
bodiam, line 250: I'm not sure about this. This breaks the API. How about instead of changing it here, we have a `DateAndTime` faker, which is the old Java thing, which uses Dates, and a new Java 8 Time, maybe called `Time` or so, which does uses only the Java 8 api, and we make the DateAndTime deprecated? 

I know we don't have many users of this library yet, but ideally I'd make the transition as easy as possible, and keep breaking the API as limited as possible. My own projects are using this library, and it's now going to break things for me.

src/main/java/net/datafaker/DateAndTime.java
```diff
@@ -250,14 +269,13 @@ public String birthday(String pattern) {
...
      */
-    public Date birthday(int minAge, int maxAge) {
-        int currentYear = Calendar.getInstance().get(Calendar.YEAR);
-        int currentMonth = Calendar.getInstance().get(Calendar.MONTH);
-        int currentDay = Calendar.getInstance().get(Calendar.DAY_OF_MONTH);
-        Calendar from = new GregorianCalendar(currentYear - maxAge, currentMonth, currentDay);
-        Calendar to = new GregorianCalendar(currentYear - minAge, currentMonth, currentDay);
-
-        return between(from.getTime(), to.getTime());
+    public LocalDate birthday(int minAge, int maxAge) {
```
bodiam, line 272: Same here. I'd prefer this to be in a separate provider, and keep this one as it, or maybe make it Deprecated or so

src/main/java/net/datafaker/Internet.java
snuyanzin: finally it will be rewritten by https://github.com/datafaker-net/datafaker/pull/56

snuyanzin:
@bodiam thanks for your feedback, probably it's better to do these changes in several separate steps, so i will close this PR and submit several others later

epragtbeamtree:
@snuyanzin I'm sorry! That was not my intention at all, I'm fine with any approach you choose.

snuyanzin:
no worries
I agree that there are some things to be improved within this PR.
From the other side there are issues with locale dates and `forbidden apis` which are not easy to be solved within this PR.
That's why i think it is better to go with smaller steps

bodiam:
Okay, that's fine. I don't want to stop your great work, or make your life harder in any way, it's just that at this moment I'd still really like to have Datafaker a drop in replacement for Javafaker, but better. 

If you say we should let go of that, I'm fine with that too, but my original intent was to just replace the dependency, change a few imports, and go. Maybe this intent is no longer applicable or relevant.

PR #56, opened 2022-02-22
Generate InetAdresses for ipv4 and ipv6

snuyanzin:
The PR adds generation of `InetAddress` objects both for `IPv4` and `IPv6`


bodiam:
Nice one!

PR #59, opened 2022-02-23
Add flags to maven to show timestamps during build

snuyanzin:
Could be helpful to see where time is consuming

the result could be seen in build logs at https://github.com/datafaker-net/datafaker/runs/5303678498?check_suite_focus=true#step:4:7

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -9,6 +9,9 @@ on:
   pull_request:
     branches: [ master ]
 
+env:
+  mvn_options: -B -Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss.SSS  
+
 jobs:
   build:
     strategy:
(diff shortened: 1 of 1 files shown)


bodiam:
Just curious, any reason why you needed this?

snuyanzin:
not right now, I picked from another project.
However it could speed up the process to identify the reason if at some certain point build starts consuming more time

bodiam:
Yeah, good idea. I was actually thinking of making the build fail if it ever takes longer than a minute to build. I'm not sure if that's an option with maven though

snuyanzin:
not sure about the whole `mvn` build... At least there are options to fail tests if timeout was reached e.g.
https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#parallelTestsTimeoutForcedInSeconds
and 
https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#parallelTestsTimeoutInSeconds

PR #70, opened 2022-03-03
David Ogar

The-True-Hooha:
Added Nigeria to the list of providers

### DIFF
--- a/src/main/java/net/datafaker/Nigeria.java
+++ b/src/main/java/net/datafaker/Nigeria.java
@@ -9,7 +9,7 @@ public class Nigeria {
 
     protected Nigeria(Faker faker) {
         this.faker = faker;
-        faker.fakeValuesService().addPath(Locale.ENGLISH, Paths.get("target/classes/en/nigeria.yml"));
+        faker.fakeValuesService().addPath(Locale.ENGLISH, Paths.get("src/main/resources/en/nigeria.yml"));
     }
 
     public String places() {
--- a/src/main/java/net/datafaker/Nigeria.java
+++ b/src/main/java/net/datafaker/Nigeria.java
@@ -1,0 +1,35 @@
+package net.datafaker.service;
+
+import net.datafaker.Faker;
+
+public class Nigeria {
+
+    private final Faker faker;
+
+    public Nigeria(Faker faker) {
+        this.faker = faker;
+    }
+
+    public String places() {
+        return faker.resolve("nigeria.places");
+    }
+
+    public String name() {
+        return  faker.resolve("nigeria.name");
+    }
+
+    public String food() {
+        return faker.resolve("nigeria.food");
+    }
+
+    public String schools() {
+        return faker.resolve("nigeria.schools");
+    }
+
+    public String celebrities(){
+        return faker.resolve("nigeria.celebrities");
+    }
+}
+
+
+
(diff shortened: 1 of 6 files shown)


src/test/java/net/datafaker/AustraliaTest.java
```diff
@@ -9,8 +9,7 @@
 public class AustraliaTest extends AbstractFakerTest {
 
     @Test
-    public void locations() {
-        assertThat(faker.australia().locations(), not(is(emptyOrNullString())));
+    public void locations() { assertThat(faker.australia().locations(), not(is(emptyOrNullString())));
```
alesaudate, line 12: Misformatted

src/main/java/net/datafaker/Nigeria.java
```diff
@@ -0,0 +1,37 @@
...
+import java.nio.file.Paths;
+import java.util.Locale;
+
+public class Nigeria {
+    private static final String KEY = "nigeria";
+    private final Faker faker;
+
+    protected Nigeria(Faker faker) {
+        this.faker = faker;
+        faker.fakeValuesService().addPath(Locale.ENGLISH, Paths.get("target/classes/en/nigeria.yml"));
```
alesaudate, line 12: Wrong location

alesaudate (changes requested):

src/main/java/net/datafaker/Nigeria.java
The-True-Hooha: Enlighten me please, wrong location for the class, or the yml file path?

src/test/java/net/datafaker/AustraliaTest.java
The-True-Hooha: Done

The-True-Hooha: I made the corresponding format changes locally, please check my last commit to see if it's corrected. 

Apologies.

bodiam:
@The-True-Hooha Thanks David, PR has been merged, most likely will be released over the weekend!!

src/main/java/net/datafaker/Nigeria.java
```diff
@@ -0,0 +1,37 @@
...
+import java.nio.file.Paths;
+import java.util.Locale;
+
+public class Nigeria {
+    private static final String KEY = "nigeria";
+    private final Faker faker;
+
+    protected Nigeria(Faker faker) {
+        this.faker = faker;
+        faker.fakeValuesService().addPath(Locale.ENGLISH, Paths.get("src/main/resources/en/nigeria.yml"));
```
alesaudate, line 12: @bodiam @The-True-Hooha I know this PR has already been merged, but this location is still wrong. I guess we need to come up with a test that actually prevents building a JAR whenever this shows up, because the test will succeed but the actual JAR will not.

alesaudate (commented):
See latest comment

src/main/java/net/datafaker/Nigeria.java
bodiam: Ah, thanks for letting me know, I even added an extra test to Faker It for this. No worries, I'll fix it.

bodiam: It's fixed, I removed this code, I already registered the yml file in the central spot.

PR #72, opened 2022-03-04
Add generation of duration

snuyanzin:
The PR adds generation of Duration for units: `day`, `hour`, `minute`, `second`, `millis`, `nanos`

e.g.
```
faker.date().duration(123, "day");
faker.date().duration(123, "days");
faker.date().duration(123, "hour");
faker.date().duration(123, "hours");
faker.date().duration(123, "minute");
faker.date().duration(123, "minutes");
```
and also possible to specify min boundary

```
faker.date().duration(3, 123, "day");
faker.date().duration(12, 123, "days");
faker.date().duration(32, 123, "hour");
faker.date().duration(43, 123, "hours");
faker.date().duration(34, 123, "minute");
faker.date().duration(121, 123, "minutes");
```

### DIFF
--- a/src/main/java/net/datafaker/DateAndTime.java
+++ b/src/main/java/net/datafaker/DateAndTime.java
@@ -1,9 +1,12 @@
 package net.datafaker;
 
 import java.text.SimpleDateFormat;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.GregorianCalendar;
+import java.util.Locale;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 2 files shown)


bodiam:
@snuyanzin Just wondering, but would the signature with the ChronoUnit make more sense than a stringy version?

snuyanzin:
It could be a separate method for that.
The problem of `ChronoUnit` is that it is impossible to use `ChronoUnit` in expressions like 
```
faker.expression("#{date.duration '15','SECONDS'}"),
```

bodiam:
Ah, yes, that makes sense. Then how about also exposing the typed version next to the stringy version? I think I'd be more inclined to use:

`faker.date().duration(123, ChronoUnit.DAYS);`

snuyanzin:
yes, that makes sense.
i will add it as well

PR #74, opened 2022-03-04
ChronoUnit could be used as an argument while Duration generation

snuyanzin:
The PR addresses feedback for https://github.com/datafaker-net/datafaker/pull/72


bodiam:
Thanks!!

PR #82, opened 2022-03-10
Lazy cache all locale values

bubuntux:
right now only English is preloaded into memory and prevent re-read the files, 
this change would make to lazy cache all locales so we have the performance gains when using multiple languages.


bodiam:
@bubuntux great, thanks for your PR!

bubuntux:
@bodiam  any chance you can release 1.2.X that includes this fix?

PR #88, opened 2022-03-18
Remove some caching to avoid same result for several calls of same ex…

### LINKED ISSUE #87: The same value is returned for random expressions within single run
**Describe the bug**
The same value is returned for random expressions within single run

**To Reproduce**
```java
import net.datafaker.Faker;

public class Main
{
    {
        Faker faker = new Faker();

        System.out.println(faker.expression("#{regexify '[a-z]{5}[A-Z]{5}'}"));
        System.out.println(faker.expression("#{regexify '[a-z]{5}[A-Z]{5}'}"));
        System.out.println(faker.expression("#{Address.city}"));
        System.out.println(faker.expression("#{Address.city}"));
    }
}
```

Actual result:
```
brmbnQKCAJ
brmbnQKCAJ
Port Quinn
Port Quinn
```

**Expected behavior**
Random data should be generated for each expression invocation.

**Versions:**
 - Faker Version: `1.2.0`

**Additional context**
This is a regression issue. It's not reproduced for faker `1.1.0`

snuyanzin:
The PR removes caching for functions as they lead to #87 
Also it adds a corresponding testcase.

fixes #87 

@bodiam I fear it could require to have a kind of fix release

### DIFF
--- a/src/main/java/net/datafaker/FakeCollection.java
+++ b/src/main/java/net/datafaker/FakeCollection.java
@@ -129,6 +129,7 @@ public class FakeCollection<T> {
         }
     }
 
+    @Deprecated // use net.datafaker.fileformats.Format.toCsv
     public static class CsvBuilder<T> {
         private String separator = ",";
         private char quote = '"';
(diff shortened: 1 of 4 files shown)


bodiam:
@snuyanzin hmmm, you might be right, though people could use the snapshot version too. Thanks for fixing it!

PR #92, opened 2022-03-22
Document binary repository URL for snapshot builds access

valfirst:


bodiam:
The documentation can also be found here: https://www.datafaker.net/documentation/getting-started/#snapshot-versions, but maybe mentioning the explicit building there is better indeed

PR #100, opened 2022-03-31
Add singleton to fake collections

snuyanzin:
The PR adds `singleton` method for collections to generate just one object


snuyanzin:
merged as https://github.com/datafaker-net/datafaker/commit/3be2a5e1eedf60e556040647f88e13204fd6ee7e

PR #101, opened 2022-04-04
Fix ru company.prefix names

Divinenickname:
`ЗАО` and `ОАО` are deprecated in russia now. Correct use in legally are `АО` and `ПАО`


bodiam:
Awesome, thanks for your PR!

PR #109, opened 2022-04-07
Move to assertj

snuyanzin:
The  PR migrates tests from hamcrest to assertj


bodiam:
Lot of changes! You didn't do that by hand right? I'm very okay with this pr

snuyanzin:
partially was by hands =), however most of it was done with help of tools

PR #123, opened 2022-04-21
Solve https://github.com/DiUS/java-faker/issues/707.

xiye2333333:
Please check it and notice us if there is any problem. Thanks a lot.
---- SE_SUSTech, group: Lanrand

### DIFF
--- a/src/main/java/net/datafaker/EldenRing.java
+++ b/src/main/java/net/datafaker/EldenRing.java
@@ -23,7 +23,7 @@ public class EldenRing {
         return faker.resolve("elden_ring.spell");
     }
 
-    public String NPC(){
-        return faker.resolve("elden_ring.NPC");
+    public String npc(){
+        return faker.resolve("elden_ring.npc");
     }
 }
--- a/src/main/java/net/datafaker/EldenRing.java
+++ b/src/main/java/net/datafaker/EldenRing.java
@@ -1,0 +1,29 @@
+package net.datafaker;
+
+public class EldenRing {
+    private final Faker faker;
+
+    protected EldenRing(Faker faker) {
+        this.faker = faker;
+    }
+
+    public String location() {
+        return faker.resolve("elden_ring.location");
+    }
+
+    public String weapon() {
+        return faker.resolve("elden_ring.weapon");
+    }
+
+    public String skill(){
+        return faker.resolve("elden_ring.skill");
+    }
+
+    public String spell(){
+        return faker.resolve("elden_ring.spell");
+    }
+
+    public String NPC(){
+        return faker.resolve("elden_ring.NPC");
+    }
+}
--- a/src/main/resources/en/elden_ring.yml
+++ b/src/main/resources/en/elden_ring.yml
@@ -1,8 +1,8 @@
 en:
   faker:
     elden_ring:
-      location: ["LIMGRAVE","WEEPING PENISULA", "LIURNIA OF THE LAKES","CAELID","AlTUS PLATEAU","MT.GELMIR","DRAGONBARROW","MOUNTAINTOPS OF THE GIANTS","CONSECRATED SNOWFIELD"]
+      location: ["Limgrave","Weeping Penisula", "Liurnia Of The Lakes","Caelid","Altus Plateau","MT.Gelmir","Dragonbarrow","Mountaintops Of The Giants","Consecrated Snowfield"]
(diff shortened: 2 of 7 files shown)


xiye2333333:
@bodiam Please check it and notice us if there is any problem. Thanks a lot.

bodiam:
Awesome, thanks for the PR. There's a few small issues with the PR. If you could address them that would be great,  otherwise happy to accept and I'll fix it myself.

I'd prefer it if you would rename the NPC method to npc (so, lowercase). That's a bit more in line with Java conventions.

Also, the yml file has quite a few strings in all caps, and I was wondering if it would be possible to turn them into capital + lowercase, so APPLE becomes Apple. 

Otherwise, no issues at all, and thank you for even updating the Readme!

xiye2333333:
@bodiam I use uppercase because the name in wiki (https://eldenring.wiki.fextralife.com/NPCs) are all in uppercase, but I think it's totally fine and better to convert them to lowercase. Thanks!

bodiam:
I think that's just the layout on the page, if you click on the characters, for example https://eldenring.wiki.fextralife.com/Boc+the+Seamster, you see that the name is no longer all caps there, which I think would be more appropriate, and more in line with the other yml files we have.

xiye2333333:
Ok, I get it. Thank you a lot!

xiye2333333:
I have changed the letters to lowercase！

src/main/java/net/datafaker/EldenRing.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public String skill(){
+        return faker.resolve("elden_ring.skill");
+    }
+
+    public String spell(){
+        return faker.resolve("elden_ring.spell");
+    }
+
+    public String NPC(){
```
bodiam, line 26: Please change this to npc (lowercase)

src/main/resources/en/elden_ring.yml
```diff
@@ -0,0 +1,8 @@
+en:
+  faker:
+    elden_ring:
+      location: ["Limgrave","Weeping Penisula", "Liurnia Of The Lakes","Caelid","Altus Plateau","MT.Gelmir","Dragonbarrow","Mountaintops Of The Giants","Consecrated Snowfield"]
+      weapon: ["Dagger","Cinquedea","Short Sword","Sword of Night and Flame","Claymore","Dark Moon Greatsword","Grafted Blade Greatsword","Rapier","Carian Glintstone Staff"]
+      skill: ["Lion's Claw","Impaling Thrust","Piercing Fang","Spinning Slash","Charge Forth","Stamp","Blood Tax","Repeating Thrust","Wild Strikes","Spinning Strikes","Double Slash","Prelate's Charge"]
+      spell: ["Adula's Moonblade","Cannon Of Haima","Collapsing Stars","Comet","Glintstone Cometshard","Great Oracular Bubble","Loretta's Mastery","Ranni's Dark Moon"]
+      NPC: ["Sorceress Sellen","Smithing Master Hewg","Sorcerer Rogier","D","Rennala","Renna","Fia","Irina","Queen Marika"]
```
bodiam, line 8: Lowercase here too please

src/test/java/net/datafaker/EldenRingTest.java
```diff
@@ -0,0 +1,32 @@
...
+        assertThat(faker.eldenRing().skill()).matches("[A-Za-z' ]+");
+    }
+
+    @Test
+    public void spell() {
+        assertThat(faker.eldenRing().spell()).matches("[A-Za-z' ]+");
+    }
+
+    @Test
+    public void NPC() {
```
bodiam, line 29: Lowercase please

bodiam (changes requested):

src/main/resources/en/elden_ring.yml
Lance-Azrael: Do we need to lower all the letters include names?

bodiam:
Ah, sorry for confusing you : no, the values are fine, it's just the key (NPC) which needs to be lower case (npc)

xiye2333333:
I have changed NPC to npc!

bodiam (approved):

bodiam:
Awesome. Thanks for implementing these suggestions, looks great to me now! Approved, and thanks again for contributing, much appreciated!

xiye2333333:
Thank you very much!

PR #127, opened 2022-04-22
Finish migration to AssertJ

snuyanzin:
The PR is intended to make use assertj wherever possible


bodiam:
It seems to also happen on Linux:

https://github.com/codecov/codecov-action/issues/586

Maybe we should revert to version 1?

snuyanzin:
Looks it was a temporary issue, now it starts working again....

Let's keep an eye to see how stable it will behave

PR #133, opened 2022-04-23
add OscarMovie.java and OscarMovie.yml

ak-maker:
add OscarMovie.java and OscarMovie.yml https://github.com/DiUS/java-faker/issues/712 and https://github.com/DiUS/java-faker/issues/713


ak-maker:
Hello @bodiam , are you willing to merge this PR. If you do it, our group will be fairly thrilled!!!

bodiam:
Absolutely! As a general feedback, there's a small issue, which is the casing of the yml file. I think we only use lowercase for that, so it would be great to keep it consist. 

However, I think this is something we can fix later, so I see nothing stopping us from  merging this PR! Great work, thank you! Merged!

ak-maker:
Thanks, our group really appreciate it!!!

PR #135, opened 2022-04-23
complete the weapon feature

Uncle-Road:


Uncle-Road:
hello @bodiam i am ready to contribute PR for weapon : ) Hope it can merge hahaha.
If have any question please chat with me.

bodiam:
@Uncle-Road PR looks great, but seems there are a few small test failures. If you address them, then I'm happy to merge this PR!

Uncle-Road:
@bodiam thanks you advance , and i check the error information and find i have bug in yml file and i already fix it.

Uncle-Road:
@bodiam thanks your advice and i check the error information and find i have bug in yml file and i already fix it.

bodiam:
@snuyanzin  Sergey, this yml file looks a bit strange. Could you please have a look at it?

snuyanzin:
Agree currently I see some functionality to extract `provenance`, `user`, `function`, `target`, `type`.

However there is nothing using `Light weapons`, `Crew served weapons` and etc.

@Uncle-Road could you please clarify for what reason do we need to have it there?

Uncle-Road:
@snuyanzin  hello, my idea is divide the weapon into **Three  Big Type** , so we choose type to choose three type. And Each  big type have some small class. And Each class have some Typical example. (the provenance is Universal random generante）
    
And you can see in the weapon.java i reslove all about it to get a weapon.
``` java
 public String getWeapon(){
        String type = faker.resolve("weapon.type");
        String types = faker.resolve("weapon.".concat(type));
        String res = faker.resolve("weapon.".concat(types));
        return res;
    }
```

Type
|__ user 
|      |_____Light weapon
|      |_____Personal weapons
|__function
|      |_____Antimatter weapons
|      |_____Non-lethal weapons
|__target
|      |_____Anti-aircraft
|      |_____Anti-fortification weapons

snuyanzin:
Ok, I see.
If I understand your idea right then it is the following
1. pick random `type`
2. pick random `types` based on `type` from point 1
3. pick random `res` based on `types` from point 2

However let's try to see at this from user's point of view who does not have any idea about it's internal implementation. 
For this user he/she gets a random `res`. So why do we need 3 random generations instead of just one?

Uncle-Road:
Actually，it is convience for api method call.      
If a user want a weapon. and it only need call **one time** method.  He just need write    
`faker.weapon.getWeapon()`       
and then he get the weapon, they donnot need kown the internal implementation. He not care the method call progress.     
And if somebody wanna fill my weapon.yml. They can easy add the example weapon.     
i do many functionality is to make divide weapon more clearly.     
If once it seem like    
```
name: [a,b,c,d,e,.....]
```
It doesn't look intuitive

snuyanzin:
Well, since this approach is unusual in compare with other `yaml`s I would suggest 
1. It would make sense to have more tests to simplify future debug. For example: currently there is no tests for case when new `type` or `user` are added or renamed then code will fail (e.g. rename `target` to `target2` in line for `type`) with `java.lang.RuntimeException: weapon.target2 resulted in null expression`. With such error it is not obvious what is wrong here and without deep diving in this implementation there is no way to understand the issue.
2. It should be also well covered with documentation to simplify maintenance efforts.
3. Probably it does not make sense to name each type `* weapon` since filename itself already says that everything here is about weapon. So I think word `weapon` could be omitted in most of the lines.
4. Nice to have: provide api to generate weapon of specified type e.g. `faker.weapon().getWeapon("Artillery")`

Uncle-Road:
ok, i will continue approve those  problems. Its very helpful. : )

bodiam:
I think you need to pull the latest master into this branch @Uncle-Road

bodiam:
Hi @Uncle-Road , we're planning on releasing 1.4.0 soon. Any chance to include this PR in the release?

bodiam:
@Uncle-Road hi, any update on this PR? I feel that it's really close, and it would be great if it could be part of this release.

bodiam:
This PR has been closed because of no activity.

Uncle-Road:
@bodiam Sorry for taking so long to reply, I've been preparing for my final exams lately. I will resubmit a PR during the summer

PR #136, opened 2022-04-23
Add ElectricalComponents faker

panilya:


bodiam:
Thanks @panilya for your PR, looks great!

PR #143, opened 2022-04-30
add ph-locale: internet, email

Divinenickname:
Add basic philippines locale. Internet and email


bodiam:
Seems like a good start. I'm in the Philippines as we speak, so this is quite appropriate! Thanks for the PR!

PR #158, opened 2022-05-15
Replace regex with cycle for javaNameToYamlName

snuyanzin:
Replacement of this regex speeds up for about 10% generation based data in yaml files
before
```
Benchmark                Mode  Cnt    Score   Error   Units
JmhTest._nameExpression  thrpt    5  181.514 ± 3.935  ops/ms
```
after
```
Benchmark                Mode  Cnt    Score    Error   Units
JmhTest._nameExpression  thrpt    5  206.220 ± 15.695  ops/ms
```

content of `JmhTest._nameExpression` is
```java
    @Benchmark
    public void _nameExpression(Blackhole blackhole) {
        Faker faker = new Faker();
        blackhole.consume(faker.name().name());
    }
```

### DIFF
--- a/src/main/java/net/datafaker/service/FakeValuesService.java
+++ b/src/main/java/net/datafaker/service/FakeValuesService.java
@@ -34,7 +34,6 @@ import java.util.stream.Collectors;
 
 public class FakeValuesService {
     private static final Pattern LOCALE = Pattern.compile("[-_]");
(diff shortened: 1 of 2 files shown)


bodiam:
Looks good!

snuyanzin:
Merged as https://github.com/datafaker-net/datafaker/commit/d549f57b0da9b98712693742c2aeefdb2bfb28a9

PR #162, opened 2022-05-16
Use valid void class for PRIMITIVE_WRAPPER_MAP

snuyanzin:
It looks like for the `void` type it is used a valid type. This PR fixes this


snuyanzin:
Merged as https://github.com/datafaker-net/datafaker/commit/f48b2946c48970201ce823b219cc116352072460

PR #166, opened 2022-05-22
Add philippines phone_number masks

divinenickname:
Add masks for philippines phones. Correct masks added from https://en.wikipedia.org/wiki/Telephone_numbers_in_the_Philippines#:~:text=As%20such%2C%20a%20typical%20telephone,63%2D2%2D8123%2D4567

image

Also, added a test


bodiam:
Thanks, nice PR!

PR #171, opened 2022-05-25
Added Famous Last Words faker

Tahanima:


bodiam:
Seems you're on a roll! Nice one again, thanks. 

Maybe a small hint for next time : we use the RepeatedTest annotation quite a bit, which repeatedly executes the test. A value somewhere between 10-100 repetitions is fine, since that detects incorrect assertions a bit earlier. But this PR is fine, thanks!

Tahanima:
> Seems you're on a roll! Nice one again, thanks.

You're welcome.

> Maybe a small hint for next time : we use the RepeatedTest annotation quite a bit, which repeatedly executes the test. A value somewhere between 10-100 repetitions is fine, since that detects incorrect assertions a bit earlier. But this PR is fine, thanks!

Thank you so much. I'll keep this in mind from next time onward.

PR #173, opened 2022-05-26
Add doc about perf comparison

snuyanzin:
The PR is aiming to show comparison of performance between datafaker and javafaker,  jFairy, kotlin-Faker across different jdks.
There is a separate project https://github.com/snuyanzin/datafaker-benchmark which could be used for some measurements

### DIFF
--- a/docs/documentation/performance.md
+++ b/docs/documentation/performance.md
@@ -1,0 +1,251 @@
+# Performance benchmarks
+This page is trying to go through some performance metrics and see how better/worse Data Faker is in compare with Java Faker and other similar projects.
+
+## Hardware & Software
+All the tests are done with help of JMH at Fedora 36 for different JDKs (mentioned in tables below).
+The laptop has 32Gb of RAM, AMD Ryzen 7 PRO 5850U with Radeon Graphics.
+
+## Data Faker (1.4.0) vs Java Faker (1.0.2) vs Kotlin-faker (1.11.0) vs JFairy (0.6.5)
+Originally Data Faker has been started as a fork of java-faker. In internet there are complaints about java-faker poor performance.
+Now there has been already 1.4.0 version released. Let's see what we can say about performance improvements.
+For that we are going to use JMH benchmarks. All the [code](https://github.com/snuyanzin/datafaker-benchmark) is available as a separate project because of license reasons.
+Since it's hardly possible and reasonable to compare each method's performance let's see what exactly we are going to compare within this article.
+As it could be seen below Data Faker is about 10x-100x times faster for some cases.
+Also, there is an interesting fact: in case of Data Faker moving from jdk8 to jdk18 performance is improving up to 25%, while for Java Faker it is not happening.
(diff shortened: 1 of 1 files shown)


bodiam:
@snuyanzin Good idea. What also might be interesting is to compare it other libraries, like kotlin-faker:

https://github.com/serpro69/kotlin-faker

They have a pretty nice table, and already a benchmark it seems. Would be nice to know how Datafaker is performing in this maybe?

snuyanzin:
Yes, I thought about it. The problem here is that there are not so many points by which we could compare since different libs provide different functionalities.... 
I add Kotlin Faker to this page

bodiam:
True, though at least kotlin faker seems to be using the same source yml files as Datafaker.net. I read your article, very interesting. Seems like kotlin faker has the same initialization problem as Java faker. It's just crazy to load all the yml files on startup.

docs/documentation/performance.md
```diff
@@ -1,74 +1,231 @@
...
+Unfortunately there is not so much information what kind of benchmarks have been executed. After some search across the project I was able to find only [this page](https://github.com/serpro69/kotlin-faker/blob/master/docs/src/orchid/resources/pages/faker-comparisons.md)
+So, it looks like the only test which were done is checking `Faker.name().name()` performance. Ok, let's start with the similar test here. Similar, because we are going to use JMH which was not used in their test.
+If applicable we try to execute same tests we did for previous section. So let's start with the original test from Kotlin Faker
+
+Original Test:
+=== "Java"
+
+```java
+    @Benchmark
+    @BenchmarkMode(Mode.Throughput)
```
bodiam, line 183: No results of this test?

snuyanzin: in the table below.
The only reason I put here 2 tests in one is that they have different clasnames and `getName`/`name` methods
like `io.github.serpro69.kfaker.Faker faker = new io.github.serpro69.kfaker.Faker();` vs `net.datafaker.Faker faker = new net.datafaker.Faker();` and `faker.getName().name();` vs `faker.name().name();`

snuyanzin:
Added a project https://github.com/snuyanzin/datafaker-benchmark and some other metrics.

bodiam:
Would it be an idea to move that repo to datafaker-net/datafaker-benchmark ? I'm happy to make repo if you can't?

snuyanzin:
No objections from my side
However I don't have enough grants to do that

bodiam:
Sorry about that, I'll see if I can fix that. In the meantime, I created a repo for it

snuyanzin:
no problem, thanks for creating it.
Meanwhile I noticed that Kotlin Faker supports some expressions (they call it random string from templates) https://serpro69.github.io/kotlin-faker/wiki/extras/#random-strings-from-templates

Thus, I will try to add comparison for that as well

bodiam:
want me to deploy this documentation?

snuyanzin:
Not yet, still in updating it

PR #174, opened 2022-05-26
Add MBTI Feature

ak-maker:


bodiam:
@ak-maker would it be possible for you to remove the testng dependency? I think having Junit 5 is enough.

bodiam:
Thanks for this PR! Nice work.

ak-maker:
My pleasure.

PR #183, opened 2022-05-27
Solve https://github.com/DiUS/java-faker/issues/377

Lance-Azrael:
This issue asked for some  last names with ñ for "es" and "es-mx" language.
We add few last name with ñ in name.yml and make it possible to generate last name with ñ and full name with these last name.
Please check it and notice us if there is any problem. Thank you a lot!
---- SE_SUSTech, group: Lanrand


Lance-Azrael:
last modification(https://github.com/datafaker-net/datafaker/pull/183/commits/afe8da136433335f8add2ccd5b20775dac1f136d) is for name like "Mr. Aaron De Romaña" with 4 parts in name.

snuyanzin:
I didn't get why are we going to introduce something ES specific in EN localle?
Why could not we use ES specific yamls for that?

bodiam:
Hi @Lance-Azrael , thanks for your work on this. 

However, I'm afraid this is a little bit of a non-standard approach for Datafaker, and goes against a few design principles in the library. 

We appreciate that you addressed the issue, but If someone wants to create names with multiple libraries, we'd recommend that theyuse multiple locales as described here: https://www.datafaker.net/documentation/usage/#multiple-locales

Lance-Azrael:
> Hi @Lance-Azrael , thanks for your work on this.
> 
> However, I'm afraid this is a little bit of a non-standard approach for Datafaker, and goes against a few design principles in the library.
> 
> We appreciate that you addressed the issue, but If someone wants to create names with multiple libraries, we'd recommend that theyuse multiple locales as described here: https://www.datafaker.net/documentation/usage/#multiple-locales
Thank you for your advice! I have studied the structure of the project again these days, and found that the functions mentioned in this issue can be realized with the data in es.yml, and only the locale language needs to be specified when using it.
I noticed that this project extracted the characteristics of specific region, such as the language and communication format unique to a region. I don't think this is a very good practice, because every time features are added, theoretically all regions should be modified (I also noticed that some features in some regions are not in others, presumably because they are not completely modified). I guess there is no particularly good solution to this, or it requires quite a bit of a change in the structure of the entire project. But anyway, is the current design better and clearer than what I did (listing regions in features), so I agree to reject this PR.
We really appreciate your patience and this wonderful project, which has taught us so much!(●'◡'●)

Lance-Azrael:
> Hi @Lance-Azrael , thanks for your work on this.
> 
> However, I'm afraid this is a little bit of a non-standard approach for Datafaker, and goes against a few design principles in the library.
> 
> We appreciate that you addressed the issue, but If someone wants to create names with multiple libraries, we'd recommend that theyuse multiple locales as described here: https://www.datafaker.net/documentation/usage/#multiple-locales

Thank you for your advice! I have studied the structure of the project again these days, and found that the functions mentioned in this issue can be realized with the data in es.yml, and only the locale language needs to be specified when using it.
I noticed that this project extracted the characteristics of specific region, such as the language and communication format unique to a region. I don't think this is a very good practice, because every time features are added, theoretically all regions should be modified (I also noticed that some features in some regions are not in others, presumably because they are not completely modified). I guess there is no particularly good solution to this, or it requires quite a bit of a change in the structure of the entire project. But anyway, is the current design better and clearer than what I did (listing regions in features), so I agree to reject this PR.
We really appreciate your patience and this wonderful project, which has taught us so much!(●'◡'●)

PR #184, opened 2022-05-27
Fix https://github.com/DiUS/java-faker/issues/379

xiye2333333:
add a function to generate name with a gender


xiye2333333:
We have implemented the function in issue: https://github.com/DiUS/java-faker/issues/379
Please check it! Thanks! 
                                              --SE_SUSTech, group: Lanrand

src/main/resources/en/name.yml
```diff
@@ -6,6 +6,9 @@ en:
       first_name:
         - "#{female_first_name}"
         - "#{male_first_name}"
+      first_name_with_gender:
+        - "#{female_first_name} #{last_name}, female"
```
snuyanzin, line 10: It's called `first_name_with_gender`, however in is configured having both first and last name.... why it is like that?

xiye2333333: because the issue asks to return a whole name and gender. Also the gender need to according to first name. So I name the function like this.

snuyanzin: The point is that name should match the action under the hood.
Currently it says about only first name and gender, however generates first name + last name. So the name does not match the action

snuyanzin: And another point is that this solution seems to be too specific to being included into the library.
Imagine another user wants similar functionality but he/she wants to have semicolon as separator instead of comma... What should this user do? New method?
Or there are some countries where not only male and female genders are possible to specify... It seems this scenario is also not covered here and no way to cover without a new method

bodiam: I agree with @snuyanzin here. I appreciate the effort you put into this and your support to make Javafaker and Datafaker better libraries, but for some issues in the issue tracker we just have to decide that we're not going to fix them, and this is one of them. 

However, if you do want to improve the code, and this also makes this feature easier, is if you can expose the female_first_name and male_first_name properties on the Name. That might be a welcome addition, especially since recently more and more language files have this distinction.

xiye2333333:
Ok, I understand. Sorry for trouble you. Thanks a lot.

src/main/resources/en/name.yml
xiye2333333: Ok, I understand. Sorry for trouble you. Thanks a lot.

Issue #159, opened 2022-05-16
Dark Soul Feature Request


NowhereTo:
**Is your feature request related to a problem? Please describe.**
[From] Java-faker's Feature Request about Dark Soul.

**Describe the solution you'd like**
Add stuff in Dark Soul to datafaker.

SickDawn:
@NowhereTo Ok, I see your issue now. I'm glad to solve this!

bodiam:
@NowhereTo @SickDawn love it! Thanks for your input, I'm happy to accept a PR related to this.  I don't consider this an issue at the moment as such, so I'll close it for now, but happy to accept a PR to support such a feature!

bodiam:
For inspiration, these a 2 great PRs which address a similar issue:

https://github.com/datafaker-net/datafaker/pull/155
https://github.com/datafaker-net/datafaker/pull/156

Thanks again!

PR #187, opened 2022-05-28
Add DarkSoul element

SickDawn:
issue: https://github.com/datafaker-net/datafaker/issues/159


bodiam:
Thank you! Really like it!

There's a few small issues with this PR however, and if you can fix them, that would be great, otherwise happy to do it myself.

One of the things is the filename and the yml key, it doesn't match the convention if the other files
Another issue is that the test will sometimes fail, since you're using () in the values, but the regex doesn't allow that.

If you could fix the above, that would be great!

bodiam:
Oh, and fixing that compilation error might be nice too :)

(you're missing () in the method in the faker.)

If you can run "./mvnw clean install" maven will tell you the build doesn't compile.

SickDawn:
> 

oh, my fault! It was really ashamed to make mistake like that. I have fix that problem now

SickDawn:
> 

about the test-failing problem, maybe I need take a little bit more time to fix it

SickDawn:
well
I notice that mismatch ing part might be weapon part. But I can't find any difference between the key in yml and in function. So I try remove it and if this pass the test, I maybe add this part back later

bodiam:
> well I notice that mismatch ing part might be weapon part. But I can't find any difference between the key in yml and in function. So I try remove it and if this pass the test, I maybe add this part back later

Sorry, I don't know what you mean with this.

SickDawn:
> One of the things is the filename and the yml key, it doesn't match the convention if the other files

in your comment, you mentioned filename and yml key was mismatched. So I try to find the mismatched one. I notice that in the .yml file, the weapon key shows different color from other key. So I think this key might be the mismatched one. But i check it for many times and still can't find the difference with the correct one. So i just remove that part. And if I could pass the test with what i have now, that means the part I removed does make wrong, right?
If I didn't misunderstand your comment, this is what I do to fix it.

SickDawn:
And it seems that all test can be passed now.

bodiam:
@SickDawn Thanks for making the changes. As you can see in the list of files, all files are written with underscores. Your file is not. Another thing is that the yml file contains the key `darkSoul`, but it should be `dark_soul` to be consistent with the other files.

SickDawn:
@bodiam oh, it looks like i misunderstand your words, sorry for that. I fix the name now.

bodiam:
No worries, I probably was a little unclear on my wording, my apologies for that! Current PR looks good, thanks for making the changes!

SickDawn:
@bodiam Thanks for your guidance！

PR #197, opened 2022-05-31
Added Big Bang Theory Faker

Tahanima:


Tahanima:
> Thanks!

You're welcome.

PR #205, opened 2022-06-07
Simplify api for collections, add initial sizes for string builders

snuyanzin:
The PR simplifies usage of collection API to generate collections.
E.g. to generate a list of numbers between `123` and `12345` of length between 3 and 10 the following code could be used
```java
        List<Integer> numbers = faker
            .collection(() -> faker.number().numberBetween(123, 12345))
            .len(3, 10).generate();
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,13 +92,11 @@ also more examples at https://www.datafaker.net/documentation/expressions/
 ### Collections
 ```java
 Faker faker = new Faker();
-List<String> names = faker.<String>collection()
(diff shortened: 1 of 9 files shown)


snuyanzin:
@bodiam probably you might be interested in having a look at this

bodiam:
@snuyanzin that's great. Still not 100% sure if that's what I need though, since I need to shorten the list, not expand it. I'll send you a sample

PR #210, opened 2022-06-09
Use Pattern instead of String for matching

snuyanzin:
For perf reasons it's better to use precompiled `Pattern` than compiling `String` each time

### DIFF
--- a/src/test/java/net/datafaker/PassportTest.java
+++ b/src/test/java/net/datafaker/PassportTest.java
@@ -2,12 +2,15 @@ package net.datafaker;
 
 import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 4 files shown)


src/test/java/net/datafaker/PassportTest.java
```diff
@@ -2,12 +2,15 @@
...
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * @author CharlotteE67
  */
+@Execution(ExecutionMode.CONCURRENT)
```
bodiam, line 13: sweet!

PR #215, opened 2022-06-11
Data generation repeatability feature enhancements

jpeffer:
feat: Added capability of setting and acquiring the seed used by a given Faker instance.
refactor: Removed Faker and RandomService instantiation by Random instance in favor of providing a seed.

### DIFF
--- a/src/main/java/net/datafaker/Faker.java
+++ b/src/main/java/net/datafaker/Faker.java
@@ -9,6 +9,7 @@ import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.Callable;
 import java.util.function.Supplier;
 
 /**
--- a/src/main/java/net/datafaker/Faker.java
+++ b/src/main/java/net/datafaker/Faker.java
@@ -9,7 +9,6 @@ import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.Random;
 import java.util.function.Supplier;
 
 /**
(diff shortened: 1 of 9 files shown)


bodiam:
@jpeffer thanks for the PR. This breaks the API, and I'm usually not a big fan of that. I'm not sure how many people are using the faker with Random constructor.

@snuyanzin  what do you think?

snuyanzin:
aagree and there are still some tests commented out... May be we could think in a direction of having two separate `RandomService` like classes. One is the old one and the second is with more `seed` control implementing same interface e.g.... This is just thoughts out loud...

jpeffer:
Yea I was fully aware it would break the API, though that was somewhat purposeful as the original design was inherently problematic. That was a carry over from the original library you forked from I believe. 

There are a few potential approaches:

1. This becomes a breaking change and therefore a major version update
2. We create a new Faker and RandomService at different package coordinates, then deprecate and modify the existing Faker and RandomService as a shim to the new ones
3. We simply deprecate the old constructors

The tests were purposefully commented out with a TODO section that contained additional questions pertaining to those tests. The main question I had was whether they covered a use case that is actually necessary or even desirable. It's encouraging a pass by reference design, which I would typically discourage. 

I left this as a draft PR simply to start the discussion. I'm happy to make further changes that meet your goals for the project. I think I favor option 2 above if you don't want to break the API now.

jpeffer:
@snuyanzin, I did run across the stackoverflow mentioning this option, but it seemed somewhat brittle. I think not relying on this workaround with be the wisest thing to do. I'll be opening another PR with an approach that aims to not break the existing API.

PR #216, opened 2022-06-14
Add DcComics Faker

panilya:


bodiam:
Thanks @panilya , small issue in the Faker, but I've fixed it, looks good to me!

PR #220, opened 2022-06-19
Set spotless checking unused imports

snuyanzin:
Add autoformatter `spotless` (https://github.com/diffplug/spotless) embedded into build
Apply google format (version 7) format. Version 7 since it still supports java8.

**Currently it is set to check unused imports only**

Now in case of violations it could being automatically fixed (including removal of unused imports) with `mvn spotless:apply` (this will be also mentioned in error message).

Here the format fix is the result of `mvn spotless:apply`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,6 +29,7 @@
         <commons-text.version>1.9</commons-text.version>
         <commons-validator.version>1.7</commons-validator.version>
         <generex.version>1.0.2</generex.version>
+        <google.javaformat.version>1.7</google.javaformat.version>
         <jacoco-maven-plugin.version>0.8.8</jacoco-maven-plugin.version>
         <jbanking.version>3.2.0</jbanking.version>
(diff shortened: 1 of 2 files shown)


bodiam:
I'm not a fan of the formatting. The line breaks seems a bit arbitrary, the javadoc code comments are now  weird, and it seems this formatting is very incompatible with my IntelliJ settings. What are you trying to solve here?

snuyanzin:
@bodiam 
I made some changes, now it checks only for unused imports

bodiam:
I don't mind spotless, but I mind the overly aggressive formatting. I'm not sure about all the options, but maybe there's a compromise to be found here

snuyanzin:
The main idea idea is automation: instead of complaining only(like e.g. checkstyle ) it could fix violations across the project

bodiam:
Let's do it

PR #221, opened 2022-06-20
Feature/core api refactor

jpeffer:
fix: Corrected failing integration unit tests, which exposed a defect in the handling of dot notation data provider properties. Previously, these properties were case-sensitive and the unit tests were failing.

feat: Introduction of core Faker interface and BaseFaker implementation, which is intended to support the decoupling of data providers, while still providing a means to shim the existing Faker API.

feat: Added seed constructors to the existing Faker implementation.

refactor: Modified existing data providers to require a net.datafaker.core.Faker implementation.


src/main/java/net/datafaker/service/FakeValuesService.java
```diff
@@ -720,7 +720,7 @@ private Object resolveExpression(String directive, String[] args, Object current
         // did first but FIRST we change the Object reference Class.method_name with a yml style internal reference ->
         // class.method_name (lowercase)
         if (dotDirective) {
-            supplier = () -> safeFetch(directive, null);
+            supplier = () -> safeFetch(directive.toLowerCase(), null);
```
jpeffer, line 723: This was needed to address failures within certain provider configurations that used title-case for the classname, like Address.street_title instead of address.street_title. The config classes could be updated or we could simply make it case-insensitive, as I have here.

src/test/java/net/datafaker/RandomFakerTest.java
```diff
@@ -20,47 +17,47 @@ public void before() {
...
-
-        resetRandomSeed();
-        final String secondInvocation = faker.internet().emailAddress();
-        assertThat(firstInvocation).isEqualTo(secondInvocation);
-    }
-
-    private void resetRandomSeed() {
-        random.setSeed(CONSTANT_SEED_VALUE);
-    }
+//    @Test
```
jpeffer, line 20: Commented out as I was questioning whether being able to change the seed by using the referenced object is a desirable thing you'd want to support. That seems like it could simply lead to unexpected behavior. The new approach allows for a new faker instance to be acquired with whatever seed is desired. 

An alternative would be to add a setSeed method inside RandomService to give someone the option to change the seed, while making the action more explicit in intent.

bodiam:
Closing this due to staleness

PR #233, opened 2022-07-18
Add 'Driving License' Faker

panilya:


bodiam:
Thanks! That's a lot of tests btw!

panilya:
@bodiam I did my best, glad that you like it. Maybe there is something I can work on?

bodiam:
@panilya  well, if you want, there's an open issue regarding the quality of the generated phone numbers. If you want to have a look at those, that might be a worthwhile improvement? Thanks for offering!

Issue #235, opened 2022-07-19
Incorrect ZhCN ID number generation


panilya:
**Describe the bug**
Chinese (zh_CN) ID number generator does not match the correct format. 'Date of Birth' in the generated number comes first, however, according to format, rather should be 'Address Code'.

**To Reproduce**
```
Faker faker = new Faker();
faker.idNumber().validZhCNSsn();
```

**Expected behavior**
![image](https://user-images.githubusercontent.com/42717485/179824049-d0cd98cd-24c0-4b73-9618-5e30319d660f.png)
(Resource: https://en.wikipedia.org/wiki/Resident_Identity_Card#Identity_card_number)

According to the format, the first 6 digits are 'Address code', the next 8 are 'Date of Birth' in the form YYYY-MM-DD, the next 3 are 'Order Code' and the last symbol is 'Checksum'. 

**Versions:**
 - OS: Linux Ubuntu
 - JDK: 11
 - Faker Version: 1.4.0

PR #236, opened 2022-07-19
Fix zh_CN ID number generator to match correct format

panilya:
Issue: #235


bodiam:
Nice writeup on the bug report!

PR #250, opened 2022-07-25
Faker for Bulgarian street name broken in 1.5.0

eikemeier:
See included regression test

### DIFF
--- a/src/test/java/net/datafaker/RegressionTest.java
+++ b/src/test/java/net/datafaker/RegressionTest.java
@@ -1,0 +1,15 @@
+package net.datafaker;
+
+import java.util.Locale;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class RegressionTest extends AbstractFakerTest {
+
+    @Test
+    void bulgarianStreetName() {
+        Faker localFaker = new Faker(new Locale("bg"));
+        assertThat(localFaker.address().streetName()).isNotEmpty();
+    }
+}


panilya:
@eikemeier Fixed. Thank you for informing about the problem.

snuyanzin:
I do not think it is fixed...
The issue is more generic
currently it seems the WA is applied only for `bg` while the same is actual for `by`, `ru`, `uk`.

I fear there could be something else (not only address related)
Moreover I would prefer do not change configuration files since they worked for previous releases...

I could have a look at it a bit later or any kind of help is also appreciated

bodiam:
I think (part of) the problem is that FakerIT doesn't run. Running it results in many broken tests. I'm not sure why yet, but I'll have a look.

snuyanzin:
@bodiam probably the fix is relatively simple, I added it to your PR #270

snuyanzin:
it looks it should be fixed within https://github.com/datafaker-net/datafaker/pull/270
I added there a bit more generic test there (for several countries)
@eikemeier could you please double check if your issue is fixed?

bodiam:
@snuyanzin I've reverted the other "fix" in the bg.yml file

snuyanzin:
I guess the issue is fixed (at least it is not reproduced at ci anymore)

eikemeier:
> it looks it should be fixed within #270
> I added there a bit more generic test there (for several countries)
> @eikemeier could you please double check if your issue is fixed?

As far as my test are concerned, the issue is fixed in 1.6.0-SNAPSHOT c02d418a7f1bbbcec0c8b331cba1ce4dca2f7e10

PR #251, opened 2022-07-25
Allow configure required and optional symbols for password generation

snuyanzin:
The PR allows configuration of optional and required symbols for password generation.
For instance, this code snippet leads to generation of passwords containing at least 1 lowercase letter, 1 uppercase letter, 1 digit and 1 "special" symbol.
```java
final Password.PasswordRuleConfig config = Password.PasswordSymbolsBuilder.builder()
    .withMinLength(6)
    .withMaxLength(10)
    .with(EN_LOWERCASE, 1)
    .with(EN_UPPERCASE, 1)
    .with(DIGITS, 1)
    .with(DEFAULT_SPECIAL, 1).build();
String password = faker.password().password(config);
```
in case we want to generate a password containing at least 1 lowercase letter, 2 uppercase letters, 3 digits and 4 "special" symbols we need to adopt `times`
```java
final Password.PasswordRuleConfig config = Password.PasswordSymbolsBuilder.builder()
    .withMinLength(6)
    .withMaxLength(10)
    .with(EN_LOWERCASE, 1)
    .with(EN_UPPERCASE, 2)
    .with(DIGITS, 3)
    .with(DEFAULT_SPECIAL, 4).build();
String password = faker.password().password(config);
```
In case number of required symbols is larger than a length of pass to generate than there are 2 options: 
* return an empty string (old behavior) and it is default behavior
* generate an `IllegalArgumentException` to do this invoke `.throwIfLengthSmall(true)` for `Password.PasswordSymbolsBuilder`

Furthermore now it allows specify other symbols to generate password
for instance, this code will allow to generate a password containing small ru and en letters and custom special symbols:
```java
final Password.PasswordRuleConfig config = Password.PasswordSymbolsBuilder.builder()
    .with(EN_LOWERCASE, 1)
    .with("абвгдеёжзийклмнопрстуфхцчшщъыьэюя", 1)
    .with("!@#$%^&*;'][{}", 1).build();
String password = faker.password().password(config);
```


bodiam:
I like it at lot, but do you think it should be part of Lorem? I know the underlying implementation is using Lorem, but wouldn't the builder be more at home in Internet?

snuyanzin:
Yes it looks also a bit strange for me...
I would extract it into a separate `PasswordProvider` with wider selection of predefined symbols sets and make reuse it from both `Lorem` and  `Internet`

snuyanzin:
I moved the logic to `net.datafaker.Password`.
@bodiam could you please have a look?
Once you do not have objections I'll merge it

bodiam:
Just had a look, looks fine to me!

Issue #252, opened 2022-07-25
Street name faker for 'bg' locale broken


panilya:
**Describe the bug**
PR that describes bug: #250 

**To Reproduce**
```
Faker localFaker = new Faker(new Locale("bg"));
assertThat(localFaker.address().streetName()).isNotEmpty();
```

**Versions:**
 - OS: Linux Ubuntu
 - JDK
 - Faker Version 15

PR #253, opened 2022-07-25
Fix broken 'Address.streetName' faker for 'bg' locale

panilya:
Issue related to the problem: #252
PR that shows this problem: #250


snuyanzin:
I wouldn't say that changing a yaml file is a solution. Moreover that configuration worked before...
We need to guarantee that same configuration should continue working if it worked for 1.4.0 and earlier.
Furthermore it looks like the same issue is actual for other locales like by, ru, uk, may be something else not related to address...

bodiam:
Shall we revert https://github.com/datafaker-net/datafaker/commit/702973bffffb4739966a887b83e12cc6b3b23398?

panilya:
> Shall we revert [702973b](https://github.com/datafaker-net/datafaker/commit/702973bffffb4739966a887b83e12cc6b3b23398)?

I think yes. Sorry about that

bodiam:
Don't worry about it, no problem at all!

PR #258, opened 2022-07-29
Make delimiter for latLon and lonLat configurable

snuyanzin:
The PR makes delimiter for `latLon` and `lonLat` configurable. Defaults to `,`


bodiam:
Just out of interest, did you need this?

snuyanzin:
This is a part of int/fraction delimiter's in different locales issue.
The problem is `ru` locale and probably some others.
In this locale comma is a delimiter between int and fraction parts of numbers.
So it's better to have another delimiter in that case.
Also I'm going to fix latitude/longitude to use corresponding delimiter between int and fraction parts of numbers

PR #261, opened 2022-07-29
Add 'Computer' faker

panilya:


src/main/java/net/datafaker/Computer.java
```diff
@@ -0,0 +1,42 @@
...
+    public String platform() {
+        return faker.resolve("computer.platform");
+    }
+
+    public String operatingSystem() {
+        List<String> operatingSystems = new ArrayList<>();
+        operatingSystems.add("linux");
+        operatingSystems.add("macos");
+        operatingSystems.add("windows");
+        return faker.resolve("computer.os." + operatingSystems.get(faker.random().nextInt(0, 2)));
```
panilya, line 28: @bodiam Is there a better way to implement this?

snuyanzin:
```java
return faker.resolve("computer.os." + faker.options().option("linux", "macos", "windows"));
```

bodiam:
Nice one!! Thanks!

bodiam:
(Sorry, I passed the wrong button!)

src/main/java/net/datafaker/Computer.java
bodiam: You could also use an Enum here

PR #264, opened 2022-07-30
Add space between providers' names

panilya:
I noticed that in the README some providers' names have spaces and some don't, so I corrected them to more readable ones.


bodiam:
Nice! What we actually need is a small tool to regenerate this list automatically, and also ideally create the providers.md file, which I think is getting a bit out of date, since it's quite tricky to update it every time

PR #273, opened 2022-08-02
Make concurrent tests more stable

snuyanzin:
~~The strange thing about passports is that it generates something strange and then in a loop tries to check if the result of generation is ok or not.
For instance,  in `net.datafaker.passportnumbers.ChPassportNumber#getValidCh` it tries to generates something with regex `[A-Z][0-9A-Z][0-9]{7}` and then checks result against `E\\d\\d{7}` and `G\\d{8}`. It is much faster to generate based on expected data.~~

Update fix concurrent issue at https://github.com/datafaker-net/datafaker/runs/7640952734?check_suite_focus=true

### DIFF
--- a/src/main/java/net/datafaker/service/FakeValuesService.java
+++ b/src/main/java/net/datafaker/service/FakeValuesService.java
@@ -39,17 +39,17 @@ public class FakeValuesService {
     private static final String DIGITS = "0123456789";
     private static final String[] EMPTY_ARRAY = new String[0];
     private static final Logger LOG = Logger.getLogger("faker");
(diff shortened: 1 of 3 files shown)


bodiam:
Hi @snuyanzin , I got rid of all that code here : 

https://github.com/datafaker-net/datafaker/commit/c0291fb1facb8fc1ddebe591c286eff4d1f22bbd

snuyanzin:
@bodiam 
ok, great.
However the funny thing is that the initial commit leads to test concurrent issue...

bodiam:
@snuyanzin  I saw that indeed, but I didn't look into a solution for that yet. Seems it's still failing?

snuyanzin:
At least now it seems more stable

Issue #163, opened 2022-05-20
Improve the quality of phone numbers generated


bodiam:
At the moment, the quality of some generated phone numbers is quite poor. To demonstrate this, there's a test called `PhoneNumberValidityFinderTest.testAllPhoneNumbers`, which generates the following output:

```
en_NZ=53
fr_CH=57
lv_LV=60
pt_PT=66
en_MS=69
tr_TR=84
zh_CN=86
hu_HU=93
en_PH=93
by_BY=94
ar_AR=97
no_NO=99
zh_TW=100
sv_SV=100
es_AR=100
bg_BG=100
```

This means that for the locales `zh_TW`, `sv_SV`, `es_AR` and `bg_BG` currently all phone numbers are generated incorrectly. 

It would be great if we can improve the quality here a little bit.

bodiam:
Bulgarian (BG) phone numbers are fixed now.

divinenickname:
@bodiam Should we divide each locale problem to different issues? As for me, it would be better to solve this big problem one-by-one.

bodiam:
They can be resolved, one by one, just put a comment here. I didn't really want to create ~30 issues just for the phone numbers

panilya:
Current state of `PhoneNumberValidityFinderTest.testAllPhoneNumbers` test:
```
sk_SK=65
lv_LV=72
fr_CH=73
en_MS=74
pt_PT=76
tr_TR=84
zh_CN=89
hu_HU=94
by_BY=95
ar_AR=97
es_AR=98
sv_SV=100
```

panilya:
@bodiam `sv_SV` is this the correct one? Because I cannot find any information about it. Maybe it's `sv_SE`?

bodiam:
@panilya  I'm not 100% sure. There's a script which generates this output, there might be an issue in the script. I know there was some mixup or so in the locales, but not 100% sure, sorry!

panilya:
@bodiam [List of locales](https://stackoverflow.com/questions/3191664/list-of-all-locales-and-their-short-codes). Cannot find the `sv_SV` locale, I will try to fix this later.

panilya:
> Current state of `PhoneNumberValidityFinderTest.testAllPhoneNumbers` test:
> 
> ```
> sk_SK=65
> lv_LV=72
> fr_CH=73
> en_MS=74
> pt_PT=76
> tr_TR=84
> zh_CN=89
> hu_HU=94
> by_BY=95
[...]

PR #274, opened 2022-08-04
Fix phone number formats for en-GB

panilya:
Issue: #163 
Minimized error code to +-10


bodiam:
@snuyanzin seems the concurrent tests are still flaky

snuyanzin:
looks like i forgot one field
currently playing with https://github.com/datafaker-net/datafaker/pull/275
in case it will not fail after a bunch of runs i'll merge it

PR #275, opened 2022-08-04
Stabilize build for the case of concurrent tests

snuyanzin:
~~Make `mapOfMethodAndCoercedArgs` non static to avoid concurrency issues~~
~~add description for tests~~
Stabilize build for the case of concurrent tests


snuyanzin:
ok, now there are 20 successful builds in a row https://github.com/datafaker-net/datafaker/actions/runs/2906808090

bodiam:
> ok, now there are 20 successful builds in a row https://github.com/datafaker-net/datafaker/actions/runs/2906808090

Awesome, thanks for your work on this. Let's hope this was the final bit.

snuyanzin:
probably this change was the most important
https://github.com/datafaker-net/datafaker/pull/275/files#diff-ee97fe2f9fc8eaa7521b384156ad4c142059c3bb6f692e1698067aafc647f55bL36-R36

I didn't expect that in `currency_code` for `uk` there could be empty value and from another side a mandatory check for absence of empty currency codes...

bodiam:
The heck :) that's weird indeed. You'd think by now most of the crazy parts are gone, but seems there's still some hidden gems in the code!

PR #282, opened 2022-08-08
Error when generation passport without specifying locale

panilya:
There is "passport.valid resulted in null expression" error when generating passport number without specifying Locale. 

Edit: 
There is `Locale.ENGLISH` - `public static final Locale ENGLISH = createConstant("en", "");` without specifying country and there is `Locale.US` - `public static final Locale US = createConstant("en", "US");`. Datafaker uses `Locale.ENGLISH` and there are a lot of problems involving this choice, for example: incorrect resolving of `phone_number` formats. When changing `Faker` to `Locale.US` everything works fine. Should I change default `Locale` to `Locale.US` in `Faker`?


bodiam:
Hmm, if you change the default locale, not a single test breaks? Cause if that's the case, I'd say go for it

panilya:
This change has side effects, so I will not change the default locale to `Locale.US`.

panilya:
@bodiam Now it seems okay, is it a good approach to solve this bug?

bodiam:
I think so, I see no issue with this, and it's much better than what we had before, so let's use this.

Issue #285, opened 2022-08-09
Add uuidv3


andrinmeier:
**Is your feature request related to a problem? Please describe.**
There is currently no way of generating repeatable/reproducible UUIDs. They are always random, even if a constant random seed is provided.

**Describe the solution you'd like**
Add internet().uuidv3() which is based on a parameter we can control. We could simply generate random bytes using the random service (which in turn uses the constant random seed). The output would be of the same structure as the existing uuid() method. It would merely be uuid v3 instead of v4.

This has been discussed in #278. Created a new issue to give this issue/solution a bit more "exposure/visibility".

andrinmeier:
PR is merged.

PR #286, opened 2022-08-09
#285: Add internet().uuidv3()

andrinmeier:
First PR to this project, feel free to provide feedback on what to change.

Added:
* internet().uuidv3()
* nextRandomBytes()
* Two tests: one whether the format is correct and one to test whether constant random seeds do what they're supposed to (which is the entire reason for this PR)

Notes:
* Using 16 bytes for generating UUID since it covers the entire UUID
* maven site didn't work locally, I hope the provided javadocs are okay
* It's also unclear how versioning works, should I increase the minor version?
* Should I add something to the changelog?

### DIFF
--- a/src/main/java/net/datafaker/Internet.java
+++ b/src/main/java/net/datafaker/Internet.java
@@ -363,9 +363,19 @@ public class Internet extends AbstractProvider {
     }
 
     /**
+     * Returns a UUID (type 3) as String.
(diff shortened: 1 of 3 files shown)


bodiam:
This PR looks good to me! Small note on the Javadoc, but nothing big. Thanks for your contribution!

Issue #278, opened 2022-08-06
Unclear on how to use Format.toCsv(FakeCollection<T> collection)


andrinmeier:
This is neither a bug nor a feature request, that's why I'm writing in free form, I hope that's okay.

I'm trying to do the following:

1. Generate valid CSV data (using datafaker)
2. Give CSV string to a parser (the one under test)
3. Compare parsed entities with the generated CSV data

How could one go about doing this? I saw that there is a Format.toCSV() overload that accepts a collection but it's unclear to me how you'd use that method.

bodiam:
No worries at all. I'm happy to help you more, but as a starting point, did you have a look at this documentation?

https://www.datafaker.net/documentation/file-formats/#csv

If so, is there anything specially unclear?

andrinmeier:
The documentation is a bit unclear on how to use a collection builder for generating CSV files. An example like the test shown here would have been nice: https://github.com/datafaker-net/datafaker/blob/94d99553f0356b2c5b74ef5bf6ef607a85a27262/src/test/java/net/datafaker/FakeCollectionTest.java#L193.

I've basically done the following right now:
1. Create a custom class that has methods for each faked value (e.g. one for generating uuids, one for generating dates)
2. Create a custom faker that uses that custom class and initializes the Faker-superclass with the same random seed each time
3. Use two different custom faker instances: one builds a collection for comparing the parsed entities and the other one for generating the CSV file

I'd be willing to submit a PR if you'd like. Three PRs, actually. One explaining how to use a collection builder for generating CSV files, another one for an end-to-end example on how to compare the custom generated entities with domain objects. The third PR would have a certain nullRate to inject errors. What do you think?

I've also noticed that generating UUIDs ignores the Faker's random seed, thus generating a "truly" random UUID each time. What do you think of changing that? A simple hack could be to simply call UUID.nameUUIDFromBytes() and passing a randomly generated number each time, using the RandomService / user provided random seed.

bodiam:
I really like all your ideas, and I'd accept any PR if you'd be willing to submit one. Thanks for the great feedback, that's really useful!

andrinmeier:
No problem, will do. Would you like me to create new issues for all the things I've just mentioned or shall I simply reference this issue in each PR?

snuyanzin:
just fyi: there is also a separate doc about collections https://www.datafaker.net/documentation/collections/
may be the issue is missing link between collection's and csv's doc pages?

[...]

PR #287, opened 2022-08-09
Add CSV format example using a fake collection builder

andrinmeier:
This PR adds a small CSV format example using a fake collection builder. To be honest, I just copy-pasted the example from the repository's README. But this should make it easier for people to understand how it works without having to go to the GitHub repository. Let me know what you think.

See #278.


bodiam:
If it helped you, then for sure we should have it in the documentation. Thanks!

PR #289, opened 2022-08-11
Feature/garments sizes

p4pupro:
---
name: Add Garments Sizes Faker
about: new feature to use Garments Sizes. see example: XS, S, M, L, XL, XXL, XXXL
title: 'Garments Sizes'
labels: 'Garments Sizes'
assignees: ''

---

**Is your feature request related to a problem? Please describe.**
No, is just a feature.

**Describe the solution you'd like**
None.

None.

**Additional context**
None.

### DIFF
--- a/src/main/java/net/datafaker/Faker.java
+++ b/src/main/java/net/datafaker/Faker.java
@@ -647,6 +647,10 @@ public class Faker {
         return getProvider(GameOfThrones.class, () -> new GameOfThrones(this));
     }
 
+    public GarmentSize garmentSize() {
+    return getProvider(GarmentSize.class, () -> new GarmentSize(this));
+  }
+
     public Gender gender() {
         return getProvider(Gender.class, () -> new Gender(this));
     }
--- a/src/main/java/net/datafaker/Faker.java
+++ b/src/main/java/net/datafaker/Faker.java
@@ -647,7 +647,9 @@ public class Faker {
         return getProvider(GameOfThrones.class, () -> new GameOfThrones(this));
     }
 
-    public GarmentSize garmentSize() { return getProvider(GarmentSize.class, () -> new GarmentSize(this)); }
+    public GarmentSize garmentSize() {
+      return getProvider(GarmentSize.class, () -> new GarmentSize(this));
+    }
 
     public Gender gender() {
         return getProvider(Gender.class, () -> new Gender(this));
--- a/src/main/java/net/datafaker/Faker.java
+++ b/src/main/java/net/datafaker/Faker.java
@@ -647,9 +647,7 @@ public class Faker {
         return getProvider(GameOfThrones.class, () -> new GameOfThrones(this));
     }
 
-    public GarmentSize garmentSize() {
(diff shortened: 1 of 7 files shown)


src/main/java/net/datafaker/GarmentSize.java
```diff
@@ -0,0 +1,76 @@
...
+  public String size() {
+    return faker.fakeValuesService().fetchString("garments.sizes");
+  }
+
+  /**
+   * This method returns a garment XS size
+   *
+   * @return a XS string of garment size
+   */
+  public String anXS() { return faker.fakeValuesService().fetchString("garments.xs"); }
```
bodiam, line 28: I don't really understand this. Why would you want this?

src/main/resources/en/garments_sizes.yml
```diff
@@ -0,0 +1,11 @@
+en:
+  faker:
+    garments:
+      sizes:  ["XS", "S", "M", "L", "XL", "XXL", "XXXL"]
+      xs: ["XS"]
```
bodiam, line 5: Same for the mapping here, why would you map xs to XS?

bodiam:
Hi @p4pupro , thanks for the PR! I've left a few comments, could you have a look? Also, it seems the build is failing, I'm not sure if that's caused by this PR, but could you please have a look at it? Thanks!

src/main/java/net/datafaker/GarmentSize.java
p4pupro: I just want a `size` method to get a random size, and a specific method for each size to get a specific size.

bodiam: But if you know which size you want, why do even want to call the method?

p4pupro: totally true... make no sense... sorry abut that.. i did a commit to solved this.

src/main/resources/en/garments_sizes.yml
p4pupro: I remove senseless mapping!

src/main/java/net/datafaker/Faker.java
```diff
@@ -647,9 +647,7 @@ public GameOfThrones gameOfThrones() {
         return getProvider(GameOfThrones.class, () -> new GameOfThrones(this));
     }
 
-    public GarmentSize garmentSize() {
-    return getProvider(GarmentSize.class, () -> new GarmentSize(this));
-  }
+    public GarmentSize garmentSize() { return getProvider(GarmentSize.class, () -> new GarmentSize(this)); }
```
panilya, line 650: I think multi-line would be better in order to maintain codestyle

p4pupro: okey, but you have many examples with one line

panilya: @p4pupro No, IDE collapses code to make classes more readable

p4pupro: Okey, I did the change to multi line.

PR #295, opened 2022-08-13
Change error count threshold value in PhoneNumberValidityFinderTest.testAllPhoneNumber

panilya:
I want to propose lower error count in the test, all the locales (except for 2, they are not actually phone numbers locale, I will raise an issue later on this) are now under 50 error count threshold, so I think it would be better to lower error count in order to be able to detect some incorrect phone number formats, and improve them more.


bodiam:
That's fine too 👍

Issue #265, opened 2022-07-30
Generator of unique values for file based generators


snuyanzin:
After thinking about https://github.com/datafaker-net/datafaker/issues/232 I came to the idea which potentially should work without issues mentioned in https://github.com/datafaker-net/datafaker/issues/232 however with the assumption that we consider only bounded set of values like data from files, enums

During resolution of provider in `net.datafaker.service.FakeValuesService#safeFetch` it retrieves a list of possible values and then picks a random.
So the idea is keeping track of retrieved values each time removing picked one. So, it means we should not retrieve it each time. Since each time we remove the one which was already generated then we guarantee that we have only those which are allowed to be generated. So there is no need to do multiple retries. Once we generate all unique values we could throw an exception 

Also the downside is that it requires deep dive into core implementation and probably will consume time to implement

bodiam:
Great! Just closing a few issues before 1.6.0 release.

PR #296, opened 2022-08-14
Add faker for pulling unique values from files

giunto:
This PR should solve issue #265 by pulling a list of all values for the given key, and popping off a random value every time. Please let me know if there are any issues with the PR.


src/main/java/net/datafaker/Unique.java
```diff
@@ -0,0 +1,60 @@
...
+        if (object == null) {
+            throw new RuntimeException(String.format(
+                "No values found for key %s",
+                key
+            ));
+        }
+
+        if (!(object instanceof List)) {
+            throw new RuntimeException(String.format(
+                "Object found for key %s is not in list format",
```
snuyanzin, line 53: In fact this exception text will not help end user to understand what he/she has done wrong

src/main/java/net/datafaker/Unique.java
```diff
@@ -0,0 +1,60 @@
...
+                key
+            ));
+        }
+
+        return new ArrayList<>((List<String>) object);
```
snuyanzin, line 58: There is no check that it is a list o strings and not a list of lists....
So probably in case it is a list of lists it makes sense to throw an exception like `there is not value for such a key`

src/main/java/net/datafaker/Unique.java
```diff
@@ -0,0 +1,60 @@
...
+        if (values.isEmpty()) {
+            throw new RuntimeException(String.format(
+                "All possible values have been generated for key %s under locale %s",
+                key,
+                locale
+            ));
+        }
+
+        int index = faker.random().nextInt(0, values.size() - 1);
+        String value = values.remove(index);
```
snuyanzin, line 32: To speed up removal of elements it is better to do in 2 steps:
1. exchange element with index `index` with the latest element
2. remove the latest element.
This will be much faster (O(1) complexity vs current O(n)) just because in that case there is no need to move elements left.
The downside  is no ordering however in this case we don't need any order

src/main/java/net/datafaker/Unique.java
```diff
@@ -0,0 +1,60 @@
...
+
+        Map<String, List<String>> valuesByKey = valuesByKeyAndLocale.getOrDefault(locale, new HashMap<>());
+        List<String> values = valuesByKey.get(key);
+
+        if (values == null) {
+            values = fetchValues(key);
+        }
+
+        if (values.isEmpty()) {
+            throw new RuntimeException(String.format(
```
snuyanzin, line 24: Probably `NoSuchElementException` fits here better

snuyanzin:
Thanks for the contribution, I left some comments.

giunto:
Thanks for the feedback! I do have one additional question. Since it's possible for YAML files to contain other data types such as integers, how should those be handled? Is it okay if we just return them as strings, or should they be handled differently?

snuyanzin:
> Since it's possible for YAML files to contain other data types such as integers, how should those be handled? Is it okay if we just return them as strings, or should they be handled differently?

this is a good question... currently i do not have a good answer for that...

src/main/java/net/datafaker/Unique.java
giunto: I added a check that filters out any lists and throws the same "No values found for key" exception. Is RuntimeException okay for that or is there a more suitable exception for this case?

src/main/java/net/datafaker/Unique.java
giunto: Done

src/main/java/net/datafaker/Unique.java
giunto: Done. Thanks for the advice.

src/main/java/net/datafaker/Unique.java
giunto: I think this issue has been resolved. I completely removed that exception and reused the "No values found for key" exception if it failed to convert to a list.

giunto:
Along with the other fixes, I updated the fetchValues method to automatically convert non-string values into strings. It didn't looks like there were many file based fakers that returned something other than a string. I think if we want to account for other datatypes we could add another method like fetchIntegerFromYaml. Let me know if there are any issues with that.

src/main/java/net/datafaker/Unique.java
```diff
@@ -0,0 +1,82 @@
...
+
+        valuesByKey.put(key, values);
+        valuesByKeyAndLocale.put(locale, valuesByKey);
+
+        return value;
+    }
+
+    private String removeAtIndex(List<String> values, int index) {
+        int lastIndex = values.size() - 1;
+        swapValuesAtIndexes(values, index, lastIndex);
```
snuyanzin, line 43: Use `java.util.Collections#swap(java.util.List<?>, int, int)` instead

src/main/java/net/datafaker/Unique.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    private void swapValuesAtIndexes(List<String> values, int firstIndex, int secondIndex) {
+        String firstValue = values.get(firstIndex);
+        values.set(firstIndex, values.get(secondIndex));
+        values.set(secondIndex, firstValue);
+    }
+
+    private List<String> fetchValues(String key) {
+        Object object = faker.fakeValuesService().fetchObject(key);
```
snuyanzin, line 54: it's better to check for `instanceof` rather than catch exceptions at least because of performance reasons

src/main/java/net/datafaker/Unique.java
```diff
@@ -0,0 +1,82 @@
...
+            throw noValuesFoundException(key);
+        }
+    }
+
+    private boolean isValidDatatype(Object value) {
+        return !(value instanceof List);
+    }
+
+    private RuntimeException noValuesFoundException(String key) {
+        return new RuntimeException(String.format(
```
snuyanzin, line 77: Why for some cases it throws `NoSuchElementException` and for others `RuntimeException` ?

giunto: I wasn't sure what the preferred exception type was. I'll change all occurrences to NoSuchElementException.

giunto: Done

src/main/java/net/datafaker/Unique.java
giunto: Done

src/main/java/net/datafaker/Unique.java
giunto: Done

src/test/java/net/datafaker/UniqueTest.java
```diff
@@ -0,0 +1,141 @@
+package net.datafaker;
+
+import net.datafaker.service.RandomService;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+import java.util.*;
+
+import static java.util.Arrays.asList;
+import static org.junit.jupiter.api.Assertions.*;
```
snuyanzin, line 10: please use assertJ assertions instead of junit's

src/main/java/net/datafaker/Unique.java
```diff
@@ -0,0 +1,72 @@
+package net.datafaker;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class Unique extends AbstractProvider {
```
bodiam, line 6: Would it be possible to add a bit of documentation to the class and maybe public methods? It's a little light on documentation as it stands, which makes it a bit harder to use for some people

bodiam:
Hi all, what's the state of this PR? I see a few open remarks related to the testing, but I don't think they're blockers for including this PR. Is there anything missing? I'd love to make a new release this week or so, but I'm not sure if this one should be included too?

src/test/java/net/datafaker/UniqueTest.java
giunto: This should be resolved now.

src/main/java/net/datafaker/Unique.java
giunto: I added some brief documentation to the class itself and the method. Let me know if anything can be phrased differently or if more explanation is necessary. I'll also work on getting some actual documentation added into the markdown docs.

giunto:
I just pushed up a couple commits that should resolve most of the outstanding issues. Assuming that those commits look good, the only thing that needs to be added are markdown docs. If those are necessary for this PR to be part of the release I can try to get those done this week.

bodiam:
Great, thank you. While the markdown documentation would be a nice to have, we can always add it later in a separate pr. It just helps people to find features like this a bit easier I guess. Thanks again!

PR #297, opened 2022-08-15
Utility for generating 'providers.md'

panilya:
At that moment, it doesn't support the descriptions. I will add later


src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+
+    private static final Pattern pattern = Pattern.compile("\\d\\.\\d\\.\\d");
+    private final JavaParser parser = new JavaParser();
+
+    // Specify destination of 'providers.md' file
+    private final String destinationPlaceOfProvidersFile = "src/test/java/net/datafaker/script/providers.md";
+
+    // Need to update this list when adding a new class in 'src.main.java.net.datafaker' which is not Faker.
+    // Store files with '.java' extension in order to avoid unnecessary transformation (substring '.java')
+    private final List<String> fakersToExcludeFromGeneration = new ArrayList<>(Arrays.asList("Faker.java", "CreditCardType.java", "AbstractProvider.java", "FakeCollection.java"));
```
snuyanzin, line 26: for contains it's better to use Set because of O(1) vs O(n)

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+                }
+
+                String tag;
+                try {
+                    tag = extractTagFromJavadoc(dirPath + nameOfFile, fakersWithoutSinceTag);
+                } catch (FileNotFoundException e) {
+                    throw new RuntimeException(e);
+                }
+
+                String fileNameWithoutExtension = nameOfFile.substring(0, nameOfFile.length() - 5);
```
snuyanzin, line 58: could you please elaborate what is majic number `5` here?

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+        final File path = new File(dirPath);
+
+        File[] files = path.listFiles();
+
+        StringBuilder providersResultString = new StringBuilder();
+
+        List<String> fakersWithoutSinceTag = new ArrayList<>();
+
+        for (File file : files) {
+            if (!file.isDirectory()) {
```
snuyanzin, line 44: May be `Files.isRegularFile()` ?

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -0,0 +1,188 @@
...
+
+    private static final Pattern pattern = Pattern.compile("\\d\\.\\d\\.\\d");
+    private final JavaParser parser = new JavaParser();
+
+    // Specify destination of 'providers.md' file
+    private final String destinationPlaceOfProvidersFile = "src/test/java/net/datafaker/script/providers.md";
+
+    // Need to update this list when adding a new class in 'src.main.java.net.datafaker' which is not Faker.
+    // Store files with '.java' extension in order to avoid unnecessary transformation (substring '.java')
+    private final Set<String> fakersToExcludeFromGeneration = new HashSet<>(Arrays.asList("Faker.java", "CreditCardType.java", "AbstractProvider.java", "FakeCollection.java"));
```
snuyanzin, line 24: In fact it is possible to use reflections lib from here https://github.com/ronmamo/reflections
then just to get all real fakers
```java
Set<Class<?>> subTypes =
  reflections.get(SubTypes.of(SomeType.class).asClass());
```
and no need to do the manual work mentioned in comments

bodiam: does doing that give you access to all the Javadoc? Or do you mean just for this list of Fakers?

snuyanzin: list of classes which could be used to read source code.

As I understand the initial idea it tries to through all java files within the folder except those mentioned in `fakersToExcludeFromGeneration` which should be constantly updated manually as stated in comments.

so it's better to get all the fakers classes as children of `AbstractProvider` and use this list to read files. Package info could be used to construct real path however it's as one of the ways

bodiam:
That looks very interesting. Nice work, this would be really nice to have to update the documentation with!

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
panilya: @snuyanzin Yes, you understood everything correctly. It is not very pleasant to add filenames that are not providers to the set every time. Good idea, never thought of that. But, if I understand correctly, there will be a lot of rewriting, or not?

bodiam: @panilya Look, while I agree with with @snuyanzin , for a simple utility class like this, I wouldn't bother to go this far. The chance of adding a new class to this package isn't that big, and even if it happens, it's a small fix to add it to the list of exclusions here. Personally, I'd just keep it like this, it's already incredibly helpful to be able to generate the correct documentation even once.

snuyanzin: not really, i can have a closer look at it later today to see if could be done with minimal changes

panilya: @bodiam I will gladly rewrite, first of all, I try to learn something new from it, and thank you for the feedback. And I always try to do everything as best as possible, so, I will try at least

panilya: @snuyanzin 
```java
Set<Class<?>> subTypes =
  reflections.get(SubTypes.of(AbstractProvider.class).asClass());
```
This will give a Set of all classes that extends `AbstractProvider`?

snuyanzin: yes, it should, at least it is stated in their doc

panilya: It's working now, but Set doesn't guarantee an ordering, so, `providers.md` is unsorted thus looking kind of unreadable. Is it okay to use TreeSet in this case?

snuyanzin: yes, if you need just lexicographic ordering, otherwise also need a corresponding comparator

bodiam: Absolutely, a TreeSet would be a great choice!

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * Writes header and table header to new 'providers.md'
+     *
+     * @see #writeToFile(String)
+     */
+    private void constructHeaderInProvidersFile() {
+        final String header = "# Fake Data Providers\n" + "\n" + "Datafaker comes with the following list of data providers:" + "\n";
+        final String tableHeader = "\n" + "| Name" + Strings.repeat(" ", 88) + "|" + " " + "Description" + Strings.repeat(" ", 120) + "|" + " " + "Since" + " " + "|" + "\n";
+        final String tableHeaderDelimiter = "|" + Strings.repeat("-", 93) + "|" + Strings.repeat("-", 132) + "|" + Strings.repeat("-", 7) + "|";
```
snuyanzin, line 107:
```suggestion
    public String generateColumn(String name, char padSymbol, int length) {
        if (name.length() >= length) return name;
        return padSymbol + name + Strings.repeat(padSymbol, length - name.length() - 1);
    }
    /**
     * Writes header and table header to new 'providers.md'
     *
     * @see #writeToFile(String)
     */
    private void constructHeaderInProvidersFile() {
        final int nameColLength = 93;
        final int descColLength = 132;
        final int sinceColLength = 7;
        final String header = "# Fake Data Providers\n" + "\n" + "Datafaker comes with the following list of data providers:" + "\n";
        final String tableHeader = "\n|" + generateColumn("Name", ' ', nameColLength) + "|" + generateColumn("Description", ' ', descColLength) + "|" + generateColumn("Since", ' ', sinceColLength) + "|\n";
        final String tableHeaderDelimiter = "|" + generateColumn("", '-', nameColLength) + "|" + generateColumn("", '-', descColLength)  + "|" + generateColumn("", '-', sinceColLength) + "|";
```

panilya:
How about introducing a file with descriptions of providers `providers_description.yml`, and getting descriptions of providers from that file when generating providers' docs with `ProvidersDocsGenerator`?

snuyanzin:
Why can't it be done in same javadoc with something like `{@description your description here}` ?

panilya:
yes, your idea is better

panilya:
Writing description in `{ }` inlines description with `@since` tag in the same line, so it looks kind of unreadable. Do you know to fix it?

snuyanzin:
Didn't get what you mean.... could you elaborate please?

I mean something like this
```
/**
 * @since 0.8.0
 * {@description your
 * multiline
 * description
 * }
 */
```

panilya:
I mean this:

image
image

But if doesn't matter, I will implement this. I apologize for the lack of clarity.

bodiam:
No, ideally don't do that!

The javadoc as-is is fine, please just use that, and don't introduce something special for this.

panilya:
everything looks good to me, what do you think? I think it's ready for merge

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -0,0 +1,204 @@
...
+        // Do not add spaces in abbreviation
+        if (providersNameAbbreviation.contains(providerName)) {
+            return providerName;
+        }
+
+        StringBuilder providerBuilder = new StringBuilder();
+        providerBuilder.append(providerName.charAt(0));
+
+        for (int i = 1; i < providerName.length(); i++) {
+            if (Character.isUpperCase(providerName.charAt(i)) && providerName.charAt(i - 1) != ' ') {
```
snuyanzin, line 197:
```suggestion
            if (Character.isUpperCase(providerName.charAt(i)) && Character.isLowerCase(providerName.charAt(i - 1))) {
```
then you probably don't need `providersNameAbbreviation` at all

panilya: @snuyanzin Thank you for pointing at the problem

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -0,0 +1,197 @@
...
+
+import java.io.*;
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.reflections.scanners.Scanners.SubTypes;
+
+public class ProvidersDocsGenerator {
+    private static final Pattern pattern = Pattern.compile("\\d\\.\\d\\.\\d");
```
snuyanzin, line 18:
```suggestion
    private static final Pattern pattern = Pattern.compile("\\d\\.\\d+\\.\\d");
```
otherwise it sill stop working starting since version 1.10.1

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -0,0 +1,197 @@
...
+
+        final String completeNameSection = "\n" + inCompleteProviderEntry + Strings.repeat(" ", lengthOfNameSectionToFill) + "|";
+
+        final String providerDescriptionSectionEntry = Strings.repeat(" ", 132) ;
+
+        // Format of `@since tag` should be '#.#.#'
+        String providerSinceTagSectionEntry = "|" + " " + sinceTag + " " + "|";
+
+        try {
+            writeToFile(completeNameSection + providerDescriptionSectionEntry + providerSinceTagSectionEntry);
```
snuyanzin, line 141:
```suggestion
String nameEntry = "\n|" + generateColumn("[" + addSpaceBetweenNameOfProvider(providerName) + "]" +
            "({{ datafaker.javadoc }}/" + providerName + ".html)", ' ', 93);

        final String providerDescriptionSectionEntry = Strings.repeat(" ", 132) ;

        // Format of `@since tag` should be '#.#.#'
        String providerSinceTagSectionEntry = "|" + " " + sinceTag + " " + "|";

        try {
            writeToFile(nameEntry + providerDescriptionSectionEntry + providerSinceTagSectionEntry);
```

usage of `net.datafaker.script.ProvidersDocsGenerator#generateColumn` could simplify the code here
probably `93` could be extracted into constant since it is also used in `net.datafaker.script.ProvidersDocsGenerator#constructHeaderInProvidersFile`

panilya: Done

snuyanzin:
Regexp does not work properly
to check run and have a look at record with `CPF`, column `since`. 
To fix it regexp should be changed to something like `"@since\\s+\\d\\.\\d+\\.\\d"` 
and `net.datafaker.script.ProvidersDocsGenerator#matchSinceTag`
to 
```java
  private String matchSinceTag(String javaDocComment) {
        Matcher comment = pattern.matcher(javaDocComment);

        if (comment.find()) {
            return comment.group().substring("@since".length()).trim();
        }
        return "";
    }

```

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -0,0 +1,188 @@
...
+        String providerSinceTagSectionEntry = "|" + " " + sinceTag + " " + "|";
+        try {
+            writeToFile(nameEntry + providerDescriptionSectionEntry + providerSinceTagSectionEntry);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * When generating new file, the old one should be removed, because new data will be written in existing one.
```
snuyanzin, line 139: It seems the comments does not match functionality: it does not remove old file and always adds.
Moreover it seems that each time this method is called it opens file again and again for every row.
It would be better to open/recreate a file once somewhere before the cycle in `generateProvidersDocs` and then just write into it

panilya: I mean when you run the entire program more than once, it just will append the entire file structure in the existing one

snuyanzin: > the old one should be removed
does it imply manually?

panilya: yes

snuyanzin: ok, just created a pr to your branch https://github.com/panilya/datafaker/pull/1
it allows to forget about such manual work

snuyanzin:
ok from my point of view it looks more or less ok
there are 2 tiny things
1. Probably `src/test/java/net/datafaker/script/providers.md` should be changed to more suitable location
2. May be it makes sense to make build failing when `fakersWithoutSinceTag` is not empty however it could be done within a different issue

@bodiam do you have anything to add?

bodiam:
I'm thinking of making a new release, is there anything open on this PR?

bodiam:
@panilya Awesome, I'll fix that. The code is a bit strange at some places, for example the `extractTagFromJavadoc`. You might think it extracts tags from Javadoc, but it's doesn't. It actually returns comments, and not only that, it also keeps track of files which didn't have a `@since` field. Also, it doesn't extract things from JavaDoc, it reads files. So, this method should actually be called:

`extractJavaodcFromSourceFileAndKeepTrackOfFilesWhichDontHaveASinceField`

which is a bit of a mouth full. So, instead, it would probably be better to give this method task: either extract comments form the Javadoc, or keep track of files which don't have a `@since` field, but not both. 

There's really no need to change this, I really appreciate your work on this, but it's just some feedback for in the future, and part of SOLID/SRP(Single Responsibility Pattern), which might be good reading material if you feel like it.

PR #307, opened 2022-08-23
Correction of Country in de.yml

redhell:
Change "die Bermudas" to "Bermudas"


bodiam:
Thanks @redhell , nice fix!

PR #308, opened 2022-08-23
Add german license plates

redhell:
Base Pattern:
char{1-3} char{2} number {1-4}
Extension for E(electric) and H(historic)


bodiam:
Hi @redhell , really appreciate it, nice work!

PR #309, opened 2022-08-23
Vehicle update in german

redhell:
Manufactures, Colors, Fuel types

(sorry for the standalone commit).

Do you have a release cycle?


bodiam:
@redhell For the release cycle, we try to release a new version every month, but your changes are already available in the snapshot version, they get published automatically on every merge.

https://www.datafaker.net/documentation/getting-started/#snapshot-versions

redhell:
@bodiam At the end/beginning of a month is fine for me.

PR #314, opened 2022-08-23
Bring back StarWars

redhell:
I don't know why the Star Wars Provider was removed but here it is back in action.


bodiam:
Hi @redhell , thanks for bringing it back, I also have no idea why it was gone, but could you please add a test for this?

redhell:
yep, I'm on it. 

I'll commit them when finished. Please squash the commits after.

redhell:
@bodiam Test ✅ 

Please squash commits and merge :)

PR #315, opened 2022-08-24
Improved documentation and slightly improved documentation generator …

bodiam:
…to also generate descriptions.

### DIFF
--- a/src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
+++ b/src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
@@ -20,14 +20,14 @@ import java.util.HashSet;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import static org.reflections.scanners.Scanners.SubTypes;
 
 public class ProvidersDocsGenerator {
-    private static final Pattern sincePattern = Pattern.compile("@since\\s+\\d\\.\\d+\\.\\d");
-    private static final Pattern descriptionPatterns = Pattern.compile(".* \\* ([A-Z].+)");
     private final JavaParser parser = new JavaParser();
 
     // Specify destination of 'providers.md' file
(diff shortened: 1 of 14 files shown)


bodiam:
@panilya  I've now added the descriptions to the generator too. Thanks, that was quite easy after you created the generator, nice work!

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -79,38 +78,41 @@ public void generateProvidersDocs(BufferedWriter writer) {
...
-        }
+            if (!isComment) {
+                fakersWithoutSinceTag.add(filePath);
+            }
 
-        return comment;
+            return comment;
+        } catch (FileNotFoundException e) {
+            throw new UncheckedIOException(e);
+        }
```
panilya, line 115: I'm curious, what are the benefits of using `UncheckedIOException` over `IOException`? Get rid of the need of handling exceptions in try-catch?

bodiam: There's not really a right answer to this and I think it's mostly a matter of opinion, but most programming languages don't have checked exceptions, such as C#, Kotlin, Groovy, Scala, etc. There's a reason for this, which is that it makes the code a bit cluttered, and most exceptions you can't recover from anyway. Also, when you do something like:

```
list.map { x -> doSomething(x) }
     .filter { y -> .... }
```

This codes becomes quite complex when `doSomething` is a method which throws checked exceptions. Checked exceptions are basically incompatible with functional programming. Because of that, I usually convert all of my checked exceptions into runtime exceptions, and in the case of IOExceptions, there is already a good candidate, which is `UncheckedIOException`. 

But, partly it's opinion. There's people who really like checked exceptions, since it gives them structure, it makes you aware of what can go wrong in code, etc. I used think like that, but in my experience, checked exceptions mostly give a lot of noise while not providing much benefit. Your mileage might vary, but I'd say try both options out and see what works best for you. Just keep in mind that of the all the mainstream languages, Java is the only one which uses checked exceptions. I think that says something :-)

panilya: Thank you for the explanation, now, I understand the purpose of `UncheckedIOException`

bodiam: No worries at all. I'd also suggest reading this book: https://www.manning.com/books/grokking-simplicity

The reason for this is that this book teaches about functions, functional programming, and immutability. I think that some of the code in the documentation generator could have been slightly better by adhering to things like immutability, and implementing CQS (Command Query Separation, a fancy term for saying that methods either do something, or they return something, but not both). Combined with things like immutability and having functions with no side effects, this makes programming far easier. 

For example, if you have a method like `processItems(items)`, and the `items` argument is a list of things, never mutate the input. Instead, have the processItems return a new list (and don't worry about the extra bit of memory, it's in 99.9% of the cases not an issue). But if you pass items to a method, and it never changes, but instead returns something new, it helps a lot in predicting what the code does, and it becomes almost a function without side effects: this makes it incredibly easy to test it. It's as easy as testing a function like `z = x + y`: processing a list of items will always return a new list of processed items.

The `extractCommentFromJavadoc` is a good example of a method like this: it's mutating the input Set, making it harder to understand what this method is doing. I personally would split this method in 2 methods:

* one which extracts the comment
* one which creates a list of fakers without since tag

You could do it like this:

```
// option 1
Set<String> fakersWithoutSinceTag = new HashSet<>();

// for each file
String version =  extractVersionFromJavadoc("src/main/java/net/datafaker/" + name + ".java");
if(version == null) {
   // no version found
   fakersWithoutSinceTag.add(name);
}
```

This is slightly better, since it doesn't mutate the fakerWithoutSinceTag inside the `extractVersionFromJavadoc` method. What would be even better is:

```
// option 2
Set<String> fakersWithoutSinceTag = findFakersWithoutSinceTag(subTypes);
```

now the fakersWithoutSinceTag can even become immutable (not to be confused with final, that's something else), and this improve the predictability of your code, because if nobody can change the Set anymore, you can be sure that the contents of the set is what you made it to be. 

In the current implementation, that's hard to say, in `option 1`, this is slightly better, and in `option 2`, it tries to combine a few of the items I mentioned all toghether. I'll leave it at this, I wasn't planning on writing a book here :-)

snuyanzin:
I moved logic to `Column` enum and it makes it much easier to add new column
now in case there is a need then just add one line at https://github.com/datafaker-net/datafaker/pull/315/commits/eea396763e677f3ebd575ed2cba5e6344e4e642a#diff-829e8c91f3a1a76f619adccc4050e582f420b47530179dff7d50155d03928f19R126-R127
and a new column will be generated

also i left only one flush instead of flushing after each write

bodiam:
> I moved logic to `Column` enum and it makes it much easier to add new column now in case there is a need then just add one line at [eea3967#diff-829e8c91f3a1a76f619adccc4050e582f420b47530179dff7d50155d03928f19R126-R127](https://github.com/datafaker-net/datafaker/commit/eea396763e677f3ebd575ed2cba5e6344e4e642a#diff-829e8c91f3a1a76f619adccc4050e582f420b47530179dff7d50155d03928f19R126-R127) and a new column will be generated
> 
> also i left only one flush instead of flushing after each write

Thanks. I think there's a few other ways to improve it, but this could potentially be a good class for learning some solid coding or so in the future. Good news is that the code works well, no we only have to add 100s of descriptions to the fakers and the documentation will be even better.

Issue #268, opened 2022-08-01
Add more passport number generators


panilya:
**Is your feature request related to a problem? Please describe.**
At that moment, there are only 2 passport number generators. Am(American) and Ch(Chinese) passport number generator.

**Describe the solution you'd like**
Add support for more countries, at least for the EU.

bodiam:
Good idea!

bodiam:
Hi @panilya , I found this page: https://docs.microsoft.com/en-us/microsoft-365/compliance/sit-defn-greece-passport-number?view=o365-worldwide

I just added quite a few passport number to the code based on this.

PR #316, opened 2022-08-24
Add passport number format for en_GB locale

panilya:
Issue: #268


src/test/java/net/datafaker/PassportTest.java
```diff
@@ -46,4 +46,11 @@ void testValidCanada() {
         assertThat(new Faker(new Locale("en", "CA")).passport().valid())
             .hasSize(8);
     }
+
+    @RepeatedTest(10)
+    void testValidUnitedKingdom() {
+        assertThat(new Faker(new Locale("en", "GB")).passport().valid())
+            .matches("[0-9]{9}")
+            .hasSize(9);
```
bodiam, line 54: I've done this too somewhere in the code right? It's a little bit of a duplication: the regex already is checking it's always 9 in length, so there's not really a point of checking again for the size of 9, but oh well :)

PR #319, opened 2022-08-31
Add 'Locale' faker

panilya:


snuyanzin:
Why do we need it use `*.yml` instead of java's embedded `DateFormat#getAvailableLocales`?

panilya:
I thought it would be better to use `locale.yml` to maintain consistency (I mean that all fakers have a corresponding `.yml` ).

snuyanzin:
Not every faker uses yml, e.g. idnumbers generators, date time generators do not use them.
Right now java provides more than `1000` locales, i do not see a reason to not reuse it

src/test/java/net/datafaker/integration/FakerIntegrationTest.java
```diff
@@ -170,6 +170,7 @@ void testAllFakerMethodsThatReturnStrings(Locale locale, Random random) throws E
         testAllMethodsThatReturnStringsActuallyReturnStrings(faker.kaamelott());
         testAllMethodsThatReturnStringsActuallyReturnStrings(faker.leagueOfLegends());
         testAllMethodsThatReturnStringsActuallyReturnStrings(faker.lebowski());
+        testAllMethodsThatReturnStringsActuallyReturnStrings(faker.localeFaker());
```
snuyanzin, line 173: There is a contract for `java.util.Locale#getDisplayName()` saying that

> depending on which fields are specified in the locale. The field separator in the above parentheses, denoted as a comma character, may be localized depending on the locale. If the language, script, country, and variant fields are all empty, this function returns the empty string.

that's the reason of failing. We cannot  say that display name is always not empty

snuyanzin, line 173:
```suggestion
```

panilya:
I can't understand why in most test runs there is 1 `Locale.baseLocale()` fall

snuyanzin:
Could you give a name of test failing after 1000 iterations?

panilya:
`LocaleFakerTest.baseLocale()`

snuyanzin:
javadoc for the `java.util.Locale#toString` contains an answer
>    * <p>If both the language and country fields are missing, this function will return
>    * the empty string, even if the variant, script, or extensions field is present (you
>    * can't have a locale with just a variant, the variant must accompany a well-formed
>    * language or country code).

src/main/java/net/datafaker/LocaleFaker.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     *
+     * @return locale in form: "English (United States)"
+     */
+    public String displayName() {
+        int randomIndex = faker.random().nextInt(availableLocales.length - 1);
+        Locale locale = Arrays.stream(availableLocales)
+            .filter(lcl -> !lcl.getDisplayName().isEmpty())
```
snuyanzin, line 31: Why do we need to filter it each time during `displayName` call if it could be filtered once during filling of  `availableLocales` field?

panilya:
```
Croatian
Chechen (Cyrillic, Russia)
English (Malta)
Punjabi (Gurmukhi, India)
Albanian (Latin, Albania)
Northern Sami
Inari Sami
English (Macao SAR China)
Koyraboro Senni (Latin, Mali)
Cebuano (Philippines)
French (Belgium)
Sangu
Samburu (Latin, Kenya)
Bodo (India)
Serbian (Cyrillic, Bosnia & Herzegovina)
Arabic (Arabic, Egypt)
```

This is the result of `Locale.displayName()`.  I don't know if this result is ok

panilya:
Result of `LocaleFaker.baseLocale()`:
```
sr_ME
ms_MY
doi_IN
en_VC
el_GR
mg_MG
rm_CH
seh_MZ
nl_BE
mzn_IR
ne_NP
ar_JO
```

Result of `LocaleFaker.displayName()`:
```
Swedish
English (Antigua & Barbuda)
Chakma (Bangladesh)
Danish (Denmark)
Konkani (India)
English (Tanzania)
Vai
Lower Sorbian (Germany)
Spanish (Ceuta & Melilla)
French (France)
English (Sint Maarten)
Faroese (Faroe Islands)
Inari Sami (Finland)
```

snuyanzin:
By accident I noticed there is another class dealing with locales `net.datafaker.service.LocalePicker`
It's better to have one instead of adding more

panilya:
@snuyanzin If it's better to have one, how to implement generation of random locales. As I see, `LocalePicker` is a service, and isn't accessible via `Faker` instance, which I don't think is acceptable, the solution I see is to add `LocalePicker` as a provider in `Faker`, what do you think?

snuyanzin:
There is no a real reason to have it not a provider.
We could convert it to provider and make it having methods to pick locale based on what is supported by datafaker and methods based on what is supported by java

panilya:
Looking at such significant changes in the project, I think it is better to close this pull request and add this provider in the new one

PR #329, opened 2022-09-09
Introduce maven checkstyle plugin

yuokada:
- Introduce maven-checkstyle-plugin
- Reformat
- Reformat: fix import order


snuyanzin:
We are already using spotless which gives autoformatting.... while checkstyle can not do this.
Currently it is enabled only for imports and also could be enabled for other code...
So I do not see a real value to have checkstyle.... Or am I missing something?

yuokada:
@snuyanzin  I didn't know about spotless itself. 

> Currently it is enabled only for imports and also could be enabled for other code...

As far as I checked, source files have different import order style. For example, the 2 following files have different import styles.

- https://github.com/datafaker-net/datafaker/blob/0e326aac6fb325aa5cd786cb80b7490056f6abcb/src/main/java/net/datafaker/IdNumber.java#L3-L15
- https://github.com/datafaker-net/datafaker/blob/0e326aac6fb325aa5cd786cb80b7490056f6abcb/src/main/java/net/datafaker/Twitter.java#L3-L10

yuokada:
I see.

> however as far as I remember @bodiam had some concerns about formatting

Just my curious. I would like to know them if possible.

bodiam:
Oh, not real concerns, but I don't ever want to encounter a failing build or anything like it because I put a space to many in some file, or because I use a * for the imports or the other way around. 

The formatting is "whatever IntelliJ's defaults are", and that's mostly it.

I've probably been bitten too much about strict styles in corporate projects which almost add no value to a  project, so I guess I'm a bit put off by tools like this.

snuyanzin:
I would say this is where spotless could help with auto formatting
if you run `mvn clean install` then it could fail because of styling issues
however if you run `mvn spotless:apply && mvn clean install` then first command will autoformat according to rules and second will do the job you want. So this combination will not fail because of styling issues

yuokada:
Thank you for explanations. I understand policy for this repository.
I will close this.

bodiam:
Not really a policy, plus we absolutely appreciate your contribution, I like to keep it as low on rules and process as possible, that's mostly it.

snuyanzin:
The way we could continue is to double check if there is a way to configure google-java-format plugin in IntelijIdea and other IDEs to behave same way like it is mentioned in `pom.xml`. 
In case of success we could enable spotless since autoformatting from spotless and IDE will the same

PR #331, opened 2022-09-10
Added release action.

bodiam:


bodiam:
@snuyanzin  Do you think this would be enough to attach it to Github Releases? The previous setup didn't, since it didn't change the snapshot to a final release.

snuyanzin:
hm, this I don't know since never did it via GH actions... however we could try

bodiam:
I've also never done it like this, but last time I did it from my local machine, didn't go well (Java 11). I'll give this a shot.

PR #352, opened 2022-09-13
Add missing state (Flevoland) for Netherlands.

jaapcoomans:
The address faker for the Netherlands (nl) lists only 11 out of 12 provinces as a state. This PR adds the only one that was still missing: Flevoland.


bodiam:
Oops, that's a shameful omission, that's my birth province! Thanks for adding it!

Issue #346, opened 2022-09-12
How to generate JSON with different data providers


panilya:
1)
```
[{
    "firstName": "Tesha", 
    "lastName": "Wyman", 
    "address": "Algeria"
}]
```
2)
```
[{
    "firstName": "Brenton", 
    "lastName": "Schumm", 
    "address": ["Saint Pierre and Miquelon"]
}]
```
Currently, I found only way to generate JSON with different providers via approach like this `.set("phones", name -> faker.collection(() -> faker.phoneNumber().phoneNumber()).len(2).build().get())`, but this way outputs as a list (code snippet 2.). Is it possible to generate JSONs as in code snippet 1.? 

`"address": "Algeria"` and `"address": ["Saint Pierre and Miquelon"]`, I need to achieve the first result.

snuyanzin:
here are 2 pojos: `FirstPOJO` and `SecondPOJO` which could generate JSON for your first and second cases
```java

import net.datafaker.fileformats.Format;

import java.util.stream.Collectors;

public class Issue346 {
  private static class FirstPOJO {
    private final Name name;
    private final Address address;

    public FirstPOJO(Name name, Address address) {
      this.name = name;
      this.address = address;
    }

    public Name getName() {
      return name;
    }

    public Address getAddress() {
      return address;
    }
  }

  private static class SecondPOJO {
    private final Name name;
    private final FakeCollection<Address> addresses;

    public SecondPOJO(Name name, FakeCollection<Address> addresses) {
      this.name = name;
      this.addresses = addresses;
    }

    public Name getName() {
      return name;
    }

    public FakeCollection<Address> getAddresses() {
      return addresses;
    }
  }

  public static void main(String[] args) {
    Faker faker = new Faker();

    System.out.println(
        Format.toJson(
                faker.collection(() -> new FirstPOJO(faker.name(), faker.address())).len(2).build())
            .set("firstName", firstPOJO -> firstPOJO.getName().firstName())
            .set("lastname", firstPOJO -> firstPOJO.getName().lastName())
            .set("address", firstPOJO -> firstPOJO.getAddress().country())
            .build()
            .generate());

    System.out.println(
        Format.toJson(
                faker
                    .collection(
                        () ->
                            new SecondPOJO(
                                faker.name(), faker.collection(faker::address).len(2).build()))
                    .len(2)
                    .build())
            .set("firstName", firstPOJO -> firstPOJO.getName().firstName())
[...]

PR #354, opened 2022-09-13
Add new JSON generation example in file formats docs

panilya:
Issue: #346


snuyanzin:
Since it is example of simple JSON generation I think it would make sense to place it  at the very beginning of JSON section and then more complex examples

snuyanzin (approved):

PR #358, opened 2022-09-15
Fix flaky test

jaapcoomans:
Fixes flaky test introduced in #355.

The pattern expects a single word, while one of the possible values consists of 2 words. Used a different pattern for the test.


bodiam:
Funny that this fix showed a new flaky test... Thanks for fixing!

jaapcoomans:
@bodiam That's inception indeed 😁
Reported an issue instead of directly fixing the other flaky test because I'm unsure about what to fix: the test of the implementation.

bodiam:
Thanks for reporting it, I do appreciate that. I'll have a look later to see what the best approach is!

PR #360, opened 2022-09-15
Make sure that padding is used in Hashing

CucumisSativus:
In some cases it's possible to have MD5 hash (and possibly other as well) with the first byte set to 0. In this case it gets removed, while converting to BigInteger and the hex string has 31 characters. To make sure that it works as expected, we need to guarantee proper padding.

### DIFF
--- a/src/main/java/net/datafaker/Hashing.java
+++ b/src/main/java/net/datafaker/Hashing.java
@@ -15,35 +15,35 @@ public class Hashing extends AbstractProvider {
     }
 
     public String md2() {
-        return generateString("MD2");
+        return generateString("MD2", "%032x");
     }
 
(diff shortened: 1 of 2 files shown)


bodiam:
Nice work @CucumisSativus , appreciate the PR!

CucumisSativus:
@bodiam thanks a lot for a quick merge! How long could it take before the new version is released to maven central?

bodiam:
@CucumisSativus a new snapshot is released after every merge, and they're generally pretty stable, so if that's something you can use for the time being, I'd recommend that. Otherwise we usually release a version about once a month.

CucumisSativus:
@bodiam thanks a lot!

PR #364, opened 2022-09-17
Introduce FakerContext, extract Faker's changeable info into FakerContext

snuyanzin:
The main issue this PR is trying to solve is that so far `net.datafaker.Faker#providersMap` contains `AbstractProvider` as value and `AbstractProvider` itself contains link to `Faker` which contains information about `Locale` and `RandomService`.
The problem is that in that map there is one `AbstractProvider` per class and this is the main blocker before movement this map to static. To see issues just switch it to static and run tests.
Another issue is that RandomService is containing in 2 classes: `Faker` and `FakeValuesService`. The PR makes is located only on Faker's level by removal from `FakeValuesService`.

The PR introduces `FakerContext` containing `Locale` and `RandomService` and it allows to satisfy all the cases we have in tests

### DIFF
--- a/docs/documentation/custom-providers.md
+++ b/docs/documentation/custom-providers.md
@@ -44,7 +44,7 @@ Create your own custom faker, which extends `Faker`, and register the custom pro
     ``` java
     public static class MyCustomFaker extends Faker {
         public Insect insect() {
-            return getProvider(Insect.class, () -> new Insect(this));
+            return getProvider(Insect.class, Insect::new, this);
         }
     }
     ```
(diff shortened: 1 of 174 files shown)


snuyanzin:
ok finally it passed more than 20 ci tests in a row https://github.com/datafaker-net/datafaker/pull/364/checks

snuyanzin:
Merged as https://github.com/datafaker-net/datafaker/commit/b41062135af26c41fade24d610f80a22338d0bd7

PR #379, opened 2022-09-23
Move id number objects to final fields to decrease number of object creations

snuyanzin:
The PR reduces creation of id related objects

### DIFF
--- a/src/main/java/net/datafaker/base/IdNumber.java
+++ b/src/main/java/net/datafaker/base/IdNumber.java
@@ -19,6 +19,12 @@ import java.time.ZoneId;
  */
 public class IdNumber extends AbstractProvider<BaseProviders> {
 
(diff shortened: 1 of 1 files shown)


bodiam:
Wouldn't this increase the number of object creations? Also, is the number of object creations really relevant?

snuyanzin:
oops, sorry I noticed your comment after merging...
I thought about lazy... I will do it in a separate step where I want to move all id fakers into a separate package
Thanks for the feedback

bodiam:
> oops, sorry I noticed your comment after merging...
> I thought about lazy... I will do it in a separate step where I want to move all id fakers into a separate package
> Thanks for the feedback

No issue at all. I don't think it has to be lazy at all, it's only 6 objects, and creating them is super fast. Using lazy might result into concurrency issues etc, , so I don't think I'd go for that.

Issue #350, opened 2022-09-13
European makes and models for cars


jaapcoomans:
I'm currently using the Vehicle faker to generate car makes and models. What I noticed, is that currently primarily US models and makes are included in the default configuration.

Now I could go and add some more European makes and models to one or more locale files, but that does not feel right. While there are differences in markets at large (US / Europe / Asia, probably some more), this is a subject that's not strictly bound to a single locale. Also in our use case we use several locale's from around Europe.

What would be the best way forward here?
* Use the locales as they are right now and just reproduce makes and models over many locale files?
* Add specific methods to the Vehicle faker for European makes and models?
* Something completely different...?

To be clear: I'd be happy to contribute and file PR, I'm just wondering what direction I should / could take with this one.

bodiam:
Hi @jaapcoomans , great to see you here. I just had a look at `vehicle.yml`, and while the list of manufacturers is quite large, the makes and models are not. 

I think the  best way forward would be to add more makes and models to the `vehicle.yml` locale, I don't see an issue with that. PR on this certainly welcome! 

(I've just added a few Tesla models to the data)

snuyanzin:
By the way a few words about how these tasks could be solved
1. Added a new provider or update an existing one in upstream.
2. datafaker supports custom providers. That means that people don't have to wait for a new release to start use their providers and use it from same datafaker like any other provider. For more info about custom providers please have a look here https://www.datafaker.net/documentation/custom-providers/

And yes contributions are more than welcome

bodiam:
@snuyanzin in my understanding this is more about updating an existing provider, the `Vehicle` provider (which is lacking a bit of content) than adding a new provider.

jaapcoomans:
@bodiam Thanks for thinking along! My question was indeed about extending the `Vehicle` provider.
I think you're right, adding more makes and models to the `vehicle.yaml` is the best approach. Expect a PR soon 🙂 

Also, TIL that the things I used to call Fakers are actually called Providers. I'll update my vocabulary 👍

bodiam:
@jaapcoomans Well, you don't have to update your vocabulary, there's not really a clear agreement on them I think. I call Datafaker a faker, and something like a Vehicle thing a provider of data, or, provider in short. But if you want to call it a faker because it generates fake vehicles, I can understand that, I just find it slightly ambiguous when talking about a Faker if we mean the Faker class, or a Data faker, so, Provider it is.

bodiam:
[...]

PR #380, opened 2022-09-23
Add more European makes and models for the Vehicle provider

jaapcoomans:
The current lists of vehicle makes and models did not include many cars that are common in Europe. As discussed in #350.
Extended the lists in `vehicle.yml` based on some resources on best selling cars in Europe of the past few years. 

Changed test assertion to accept `Citroën` and `Mercedes-Benz` as valid values.


bodiam:
Awesome! Thanks. Seems the test doesn't like VW Id.4 though

jaapcoomans:
Indeed, I missed some test patterns. Fixed it 👨‍🔧

bodiam:
Thanks Jaap for this pr! Btw, seems our random testing isn't so bad ;-)

PR #384, opened 2022-09-24
Datafaker wasn't 100% identical when repeating with the same seed, no…

bodiam:
…w it is. Plus wrote test to prove this, including a test to show that the results are different when it's created with a new faker.


bodiam:
@jaapcoomans @snuyanzin FYI

snuyanzin:
lgtm
I added a couple of commits to make tests green

bodiam:
@snuyanzin thanks! I missed that the tests were failing, thanks for fixing!

PR #391, opened 2022-09-24
Add test to check all methods with same seed are working as expected

snuyanzin:
The PR adds test for same seed for all methods with no args of all providers


bodiam:
Nice, but how is this different from https://github.com/datafaker-net/datafaker/pull/384?

snuyanzin:
oops, didn't want to commit , just played around

bodiam:
No worries at all. I'm also happy if you want to replace my test with this one, it looks fine to me.

PR #397, opened 2022-09-25
Move 'Overwatch' faker to videogame group

panilya:


src/test/java/net/datafaker/base/OverwatchTest.java
```diff
@@ -1,10 +1,11 @@
 package net.datafaker.base;
```
snuyanzin, line 1: Also makes sense to move to `package net.datafaker.videogame;`

panilya: Fixed
Forgot about that

panilya: @snuyanzin  Feel free to merge if everything seems okay

PR #399, opened 2022-09-26
Rename botUserAgentAny method

yuokada:
Follow the following change.

- https://github.com/datafaker-net/datafaker/pull/394

----

- Let me add some bot user agents to `internet.yaml` in this PR.


bodiam:
@yuokada Any reason why this PR is here?

PR #401, opened 2022-09-26
Add 'Locality' faker

panilya:


panilya:
@snuyanzin Is it okay to do it this way?

README.md
```diff
@@ -303,6 +303,7 @@ Providers
 * League Of Legends
 * Lebowski
 * Lord Of The Rings
+* Locale
```
panilya, line 306: What is better, `Locale` or `LocalePicker`, because in the code, this faker is called `localePicker()`?

bodiam: I think it's better if this list is generated (like you did with the Markdown) and matches the code. However, Locale, while a good name, is a bit conflicting with `java.util.Locale`, so maybe LocaleProvider (or picker), or perhaps Internationalization? (but then again, a lot of the Faker is related to i18n). Tricky question!

panilya: How about renaming the class name to `LocaleProvider` and using this name throughout the project?

panilya: Or `LocaleFaker` is better. I see, that in the code there are `Base` and `Provider` suffixes.

snuyanzin: What about plural `Locales`?
`Base` and `Provider` suffixes are used for abstract classes/interfaces and there is no need for every class add such suffixes. It's clear that if any class extends/implements anything with `Base`/`Provider` then it will definitely be `Base`/`Provider`

bodiam: I thought about Locales too, but it's not like we have `Names`, `Addresses`, or even `Dogs`. Basically, all our providers are singular, so maybe that's not ideal. How about calling it `Locality` or so? I'm no native speaker, but it's a bit of a synonym in most cases, but yes, I know it generates `Locale`s, so that's not ideal either perhaps.

snuyanzin: i'm ok with `Locality` as well

panilya: Then, I will rename to `Locality`

panilya:
I can't find the cause of this test failure

snuyanzin:
I created a PR to your branch fixing it https://github.com/panilya/datafaker/pull/2

bodiam:
There might be a bug here, I'm not 100% sure. The test failed because of the repeatability of the locality:

Expected:

```
 "locality.displayName"="Russian",
 "locality.localeString"="en-ZA",
 "locality.localeStringWithoutReplacement"="fi-FI",
```

Actual:

```
 "locality.displayName"="Armenian",
 "locality.localeString"="vi",
 "locality.localeStringWithoutReplacement"="ru",
```

I had a quick look, but I couldn't see an obvious mistake. Maybe there's an issue with using a Set or so, which might have an unpredictable ordering, but could be something else too. But hey, the test works!

bodiam:
Ah, it seems all is good now! Thanks both of you!

panilya:
I think it's ready to merge

bodiam:
Then let's merge it :-)

PR #409, opened 2022-09-28
Add port provider

panilya:


src/test/java/net/datafaker/base/InternetTest.java
```diff
@@ -146,6 +146,11 @@ void testPasswordMinLengthMaxLengthIncludeUpperCaseIncludeSpecial() {
         assertThat(faker.internet().password(10, 25, true, true)).matches("[a-zA-Z\\d!@#$%^&*]{10,25}");
     }
 
+    @Test
+    void testPort() {
+        assertThat(faker.internet().port()).isLessThanOrEqualTo(String.valueOf(65535));
```
snuyanzin, line 151: -1, -2, -3 and etc are also less than 65535 however should not be accepted here

panilya: @snuyanzin What is a better way to compare?

snuyanzin: depends on use case if you want to compare numbers then you should compare numbers
if you want to compare strings then you should compare strings
right now you are comparing strings how it looks like you want to compare numbers

src/main/java/net/datafaker/base/Internet.java
```diff
@@ -151,6 +151,14 @@ public String password(int minimumLength, int maximumLength, boolean includeUppe
 
         return faker.password().password(config);
     }
+    
+    /**
+     * Returns a port number between 0 and 65535
+     * @return a port number
+     */
+    public String port() {
```
snuyanzin, line 159: why should be string?

panilya: I hastily did, thought for a moment that port should be a String in order to follow the format, after your comment realized that it's not true.

panilya: Changed to int

src/test/java/net/datafaker/base/InternetTest.java
panilya: Fixed

bodiam: Great! Btw, I'd recommend to use RepeatedTest instead of Test. Ideally with a parameter between 10 and 100, for example, RepeatedTest(10) or so. Anything above 100 will make the build a bit slower, but I find that for fast tests, 10-100 will find some issues quicker than just @Test. 

In any case, I think this code looks fine, even without the RepeatedTest

panilya: @bodiam Replaced `@Test` to `@RepeatedTest(100)`

snuyanzin: the range in port generation and in test should be same otherwise there is still a chance to be flaky for the test

src/main/java/net/datafaker/base/Internet.java
```diff
@@ -152,6 +152,15 @@ public String password(int minimumLength, int maximumLength, boolean includeUppe
         return faker.password().password(config);
     }
 
+    /**
+     * Returns a port number between 1 and 65535
+     *
+     * @return a port number
+     */
+    public int port() {
+        return faker.random().nextInt(1, 65535);
```
snuyanzin, line 161: Based on RFC 6335 https://www.rfc-editor.org/info/rfc6335
allowed range is `[0, 65535]` so it should be `return faker.random().nextInt(0, 65535);`

panilya: Fixed

bodiam:
Looks good to me!

PR #416, opened 2022-09-29
More reformat

bodiam:
Reformatted all the code using IntelliJ default settings.

### DIFF
--- a/src/main/java/net/datafaker/FakeCollection.java
+++ b/src/main/java/net/datafaker/FakeCollection.java
@@ -107,7 +107,7 @@ public class FakeCollection<T> {
             minLength = minLength < 0 ? maxLength : minLength;
 
             RandomService randomService;
-            if(faker == null) {
+            if (faker == null) {
                 randomService = new RandomService();
             } else {
                 randomService = faker.random();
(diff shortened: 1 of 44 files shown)


snuyanzin:
i wonder how did you get this format?
I  would suggest to have kind of spotless to allow format as embedded and have it the same for everyone

bodiam:
@snuyanzin I used IntelliJ default format.

bodiam:
And I don't mind any tools, but I'd rather not install extra plugins in my IDE, and I'd prefer never to see a build failure because someone added a space too many or something. But if `mvn clean install` or something does a reformat too (without breaking things) then I'm fine. 

Reformatting the code actually broke things in `Json.java`, seems (my) IntelliJ doesn't like the commented out code:

image

snuyanzin:
the problem that if it is open source we can not force to have everyone use only a specific IDE with always default settings...

bodiam:
How is that a problem? The code was mis-formatted like this for a few years. 

If someone is using any different IDE such as Netbeans or Eclipse, then I hope they use `.editorconfig`, and if they don't, then I'll fix it.

PR #422, opened 2022-10-04
Add 'Fullmetal Alchemist: Brotherhood' provider

panilya:


src/main/java/net/datafaker/movie/FullmetalAlchemist.java
```diff
@@ -0,0 +1,23 @@
+package net.datafaker.movie;
+
+import net.datafaker.base.AbstractProvider;
+
+public class FullmetalAlchemist extends AbstractProvider<MovieProviders> {
```
bodiam, line 5: It's missing a @since ......

panilya: Fixed

Issue #429, opened 2022-10-09
Add transformation schemas


snuyanzin:
The problem is that currently to produce JSON, CSV or any other format there should be configured rules independently for each of the formats and for each in its own way... It makes usage of that functionality more complicated.

The idea is to create a transformation schema which could define the rules of retrieval/supplying data and transformers which could define rules of transformations into specific formats.
e.g. there could be a schema
```java
final Schema.SchemaBuilder<Name, Object> schemaBuilder =
        new Schema.SchemaBuilder<Name, Object>()
            .of(
                field("first", Name::firstName),
                compositeField(
                    "second",
                    new Field[] {field("third", Name::username)}),
                field("fifth", Name::fullName)).build();
```

and transformer like
```java
public interface Transformer<IN extends AbstractProvider<?>, OUT> {
    OUT apply(IN input, Schema<IN, ? extends OUT> schema);
}
```
Then all the transformers should care how to transform to something based on schema without dealing about schema definition
Finally e.g. with `JsonTransformer` it could generate
```json
{"first":"Dania", "second":{"third":"lindsay.wiza"}, "fourth":"Kelvin West"},
{"first":"Tad", "second":{"third":"usha.greenholt"}, "fourth":"Ronald Lakin"},
{"first":"Evelyne", "second":{"third":"regan.maggio"}, "fourth":"Claire Wehner"},
{"first":"Jefferson", "second":{"third":"pam.bayer"}, "fourth":"Willie Jacobson"}
```
It will generify and simplify support of different transformations

snuyanzin:
done

PR #430, opened 2022-10-09
Introduce transformation schemas

snuyanzin:
The PR introduces schemas for transformations mentioned at https://github.com/datafaker-net/datafaker/issues/429
Also it introduce a new SQL transformer

an example of usage
```java
Faker faker = new Faker();
      System.out.println(
        new SqlTransformer.SqlTransformerBuilder()
            .build()
            .generate(
                faker.collection(faker::name).len(5).generate(),
                Schema.of(field("name", Name::fullName), field("username", Name::username))));
```

which results to
```sql
insert into "MyTable" ("name", "username") values ('Joel Adams', 'darren.wiza');
insert into "MyTable" ("name", "username") values ('Cameron Runolfsdottir', 'kayla.robel');
insert into "MyTable" ("name", "username") values ('Justin Ernser', 'michel.anderson');
insert into "MyTable" ("name", "username") values ('Mauro Pouros', 'jesus.schulist');
insert into "MyTable" ("name", "username") values ('Annie Cormier Jr.', 'roseline.vonrueden');

```

### DIFF
--- a/src/main/java/net/datafaker/transformations/CompositeField.java
+++ b/src/main/java/net/datafaker/transformations/CompositeField.java
@@ -1,0 +1,22 @@
+package net.datafaker.transformations;
+
+import net.datafaker.providers.base.AbstractProvider;
+
+public class CompositeField<MyObject extends AbstractProvider<?>, MyType> extends Schema<MyObject, MyType> implements Field<MyObject, MyType> {
+    private final String name;
+
+    public CompositeField(String name, Field<MyObject, MyType>[] fields) {
+        super(fields);
+        this.name = name;
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    @Override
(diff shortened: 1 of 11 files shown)


src/test/java/net/datafaker/FakerTest.java
```diff
@@ -36,6 +36,11 @@ void examplifyMixed() {
         assertThat(faker.examplify("abc123ABC1zzz")).matches("[a-z]{3}[0-9]{3}[A-Z]{3}[0-9][a-z]{3}");
     }
 
+    @Test
+    void test() {
+        assertThat(9).isBetween(1, 65535);
```
epragtbeamtree, line 41: That's a nice test :-) (I think that slipped in when testing the ports?)

snuyanzin: yes you're right, good catch, thanks

PR #435, opened 2022-10-14
FakeStream implementation

NarekDW:
This is the FakeStream simple implementation.
But I have a few doubts here:
1. Is it worth to move FakeStream and FakeCollection classes to separate folder (like util) ?
2. For now it doesn't support operations like
`Format.toCsv(FakeStream.Builder...)` and `Format.toJson(FakeStream.Builder...)`
as it (Csv and Json classes in particular) was designed for FakeCollection only and it will take much effort (and much redundant and similar code) to add such an ability. What do you think, how could we improve the code and avoid of writing a lot of similar code? Probably to add some new level of abstraction? Or adding some converter function, which could convert FakeStream to FakeCollection and the pass collection to the functions like `Format.toCsv(FakeStream.Builder...)` and `Format.toJson(FakeStream.Builder...)` ?
3. Also, do we need a common interface or abstract class for FakeCollection and FakeStream, as them contain quite similar API? From my point of view, here we don't need common interface, as in fact them belong to different subjects...


snuyanzin:
Hi @NarekDW great to see here and thanks for the contribution.
Agree about common classes.
First I would say we could completely reuse Builder and would say replace `build()` methods with `collection()` to build `FakeCollection` and `stream()` to build `FakeStream`... In theory for backward compatibility `build()` method could be still kept and marked as deprecated.
What do you think?

I will try to have a deeper look later

NarekDW:
Hi @snuyanzin glad to see you :blush: 
Thank you for the response :)
As I understood, you want to combine `FakeCollection` and `FakeStream` into one class and reuse common Builder, right?
If so, then I see some issues here:
1. Collection and Stream have different nature, Stream is not Collection and vice versa.
2. At first glance I see a number of challenges in implementation, for example: 
2.1 initialization issue.
 FakeCollection - is always a fixed sized data structure (with default value  = 10).
 FakeStream - could be an infinite size data structure.
2.2 Issue to distinguish which data structure was created by the client (Collection or Stream)... 

Correct me if I’m wrong please.
Don't you think that it would be better to keep them (FakeCollection and FakeStream) separately?

snuyanzin:
Not really. `FakeCollection` and `FakeStream` could be in separate classes
At the same time as you suggested above there could be an interface e.g. `Sequence` (probably there could be better name).
Both `FakeCollection` and `FakeStream` could implement it.
At the same time both builders are the same, so it could be extracted to this `Sequence` interface or even separately.

Initialization issue could be easily workarounded e.g.
```java
    public static class Builder<T> {
        private final List<Supplier<T>> suppliers;
        private int minLength = -1; // negative means same as maxLength
        private int maxLength = -1; // negative means infinite stream
        private double nullRate = 0d;
        private BaseProviders faker;
    ...

    public FakeStream<T> buildStream() {
      ...
    }

    public FakeCollection<T> buildCollection() {
        maxLength = maxLength == -1 ? 10 : maxLength;
        ...
     }
}
```

> 2.2 Issue to distinguish which data structure was created by the client (Collection or Stream)...

Didn't get this problem. In `Faker` there could be 2 different methods `collection()` and `stream()`. So depends on which was invoked

NarekDW:
@snuyanzin 
1. I've added an abstract class `FakeSequence`, which is common class for `FakeCollection` and `FakeStream` classes, with common methods/API/Builder (in the child classes we just overwrite the `build` method).
2. Also, I changed CSV and Json creation method to support that common  (`FakeSequence`) class, instead of using `FakeCollection` only. Added support to generate csv and json from finite/infinite stream.
3. I moved that classes to the separate folder - `util`. Is it fine from backward compatibility point of view?
4. I don't like some classes names like `CsvCollectionBasedBuilder` and `JsonFromCollectionBuilder` + `JsonForFakeCollection`. It could be better to name them `*Sequence*`,  but I didn't change them as they are public classes and the change could lead to some issues with client code (regarding backward compatibility)...

Could you review please? 
Is this the same way which you mentioned in your previous comment?

src/main/java/net/datafaker/formats/Json.java
```diff
@@ -280,6 +283,23 @@ private void value2String(T record, Object value, StringBuilder sb) {
...
             }
         }
+
+        private List<T> getValuesFrom(FakeSequence<T> sequence) {
+            if (sequence == null) {
+                return null;
+            }
+
+            if (sequence.isInfinite()) {
+                throw new IllegalArgumentException("Trying to generate a CSV from infinite sequence");
```
snuyanzin, line 293: I think it's better when exception message says what should be done to fix the issue, e.g.
`The sequence should be finite`

snuyanzin:
thanks, in general it looks good
there are a few cosmetic comments
1. >I moved that classes to the separate folder - util. Is it fine from backward compatibility point of view?

It's better to have some more self-explanatory names e.g. `sequence` rather than `util` 
2. Currently there are 2 classes `CsvTest` within this PR, the one from `fileformats` package should be removed together with `fileformats` package

src/main/java/net/datafaker/formats/Json.java
NarekDW: Yep, sounds reasonable, done.

NarekDW:
Done. Changed the folder name from `util` to `sequence` and changed the error message in case of generating csv or json file from infinite sequence.

src/main/java/net/datafaker/sequence/FakeSequence.java
```diff
@@ -0,0 +1,111 @@
...
+        public final FakeSequence.Builder<T> suppliers(Supplier<T>... suppliers) {
+            Objects.requireNonNull(suppliers);
+            this.suppliers.addAll(Arrays.asList(suppliers));
+            return this;
+        }
+
+        public abstract FakeSequence<T> build();
+
+        /**
+         * @param <S> [Stream<T>, List<T>]
```
snuyanzin, line 103: Looks like this is a reason of ci failure, also fails locally with `mvn clean install`
@NarekDW  fyi

NarekDW: Yep, my bad, fixed

snuyanzin:
Merged
thanks for the contribution  @NarekDW

bodiam:
@NarekDW This is a really great contribution! Love it! Just wondering: would it be possible for you to add something to our documentation, the markdown files I mean? If you don't have time, no problem, I'm happy to port some of your unit tests to the public website (I saw you added something to the README already) and publish it here or so:

https://www.datafaker.net/documentation/collections/

snuyanzin:
this page is a result of generation based on input here https://github.com/datafaker-net/datafaker/blob/main/docs/documentation/collections.md
so you can just edit this file

NarekDW:
Thanks, let me take a look.

bodiam:
Thanks! 

And sorry about that 404, I've probably configured something wrong, thanks for reporting!

NarekDW:
Please review - https://github.com/datafaker-net/datafaker/pull/443
Is that autogeneration (of docs) a part of CI?
How can I check the result? :)

PR #447, opened 2022-10-18
Rework Generics of Transformer interface

NarekDW:
@snuyanzin  could you review generics in Transformer interface, please?

I've found a number of issues here:
1. On the code below, `Schema<?, ? extends OUT>` means that `OUT transform(IN input)` function from `Field` interface could generate the value only of some specific type. For example for `CsvTransformer` class, `OUT` is defined as `CharSequence`, so it means that we can generate CSV only from CharSequence, we couldn't generate CSV (fields) from other objects (integer, double, boolean e.g.). 
Take a look on the test case `testCsvWithDifferentObjects` and `testCsvWithDifferentObjectsFunction`.
```java
public interface Transformer<IN extends AbstractProvider<?>, OUT> {
    OUT apply(Object input, Schema<?, ? extends OUT> schema);
...
```

2. Here, `IN` generic type should not extends AbstractProvider, because in fact that type corresponds to the  
`Schema<IN, ...>` interface  then it is passed to the function `OUT transform(IN input)`  from `Field` interface, that function takes an argument of IN type and feeds it to the functional interface, which generates the data based on the argument, therefore, the argument could be an object of any type. 
Take a look on the test case `testCsvWithDifferentObjects` and `testCsvWithDifferentObjectsFunction`.
```java
public interface Transformer<IN extends AbstractProvider<?>, OUT> {
    ...
    String generate(List<IN> input, final Schema<IN, ? extends OUT> schema);
    ...
}
```

3. I think all the functions from interface `Transformer` should return a data of type `OUT`

Otherwise, if we don't want to change the types definition, we have to try to adjust the code to that types.

### DIFF
--- a/src/main/java/net/datafaker/transformations/CsvTransformer.java
+++ b/src/main/java/net/datafaker/transformations/CsvTransformer.java
@@ -17,7 +17,7 @@ public class CsvTransformer<IN> implements Transformer<IN, CharSequence> {
   }
 
   @Override
-  public CharSequence apply(Object input, Schema<IN, ?> schema) {
+  public CharSequence apply(IN input, Schema<IN, ?> schema) {
     Field<IN, ?>[] fields = schema.getFields();
     if (fields.length == 0) {
       return "";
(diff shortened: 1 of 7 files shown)


snuyanzin:
awesome, thanks

PR #448, opened 2022-10-19
feat(adding_moves_to_pokemon_provider)

TGITS:
The list of moves exists in the resource file but was not exploited in the pokemon provider.
Small contribution to use it in order to allow the production of  move in the provider.


TGITS:
No problem. As a matter of fact, 1000 repetitions is maybe a little too much. I will lower the repetitions to 100 if it's ok with you.

bodiam:
Great, thank you!!

PR #453, opened 2022-10-20
Colorize Maven output in GitHub Actions

timtebeek:
Through -Dstyle.color=always just after --batch-mode


bodiam:
cool stuff! Thanks for that!

PR #454, opened 2022-10-20
Run AssertJ best practices OpenRewrite recipe

timtebeek:
Fairly simple changes automatically applied through OpenRewrite.
Now consistently uses static imports for assertions, and test method names dropped the no-longer-needed-test-prefix.

https://docs.openrewrite.org/reference/recipes/java/testing/assertj/assertj
https://docs.openrewrite.org/reference/recipes/java/testing/cleanup/bestpractices

### DIFF
--- a/.github/workflows/openrewrite.yml
+++ b/.github/workflows/openrewrite.yml
@@ -1,0 +1,27 @@
+# This workflow will build a Java project with Maven, and cache/restore any dependencies to improve the workflow execution time
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: OpenRewrite dryRun
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+env:
+  mvn_options: --batch-mode -Dstyle.color=always -Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss.SSS
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up JDK 8
+        uses: actions/setup-java@v3
+        with:
+          java-version: '8'
+          distribution: 'temurin'
+          cache: maven
+      - name: OpenRewrite dryRun
+        run: ./mvnw $mvn_options rewrite:dryRun --file pom.xml
--- a/pom.xml
+++ b/pom.xml
@@ -378,5 +378,31 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>openrewrite-check</id>
+            <build>
+                <plugins>
+                    <plugin>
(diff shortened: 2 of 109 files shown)


snuyanzin:
Thanks for your PR
from my point of view such changes should come together with some automatic rules which could prevent future methods name in similar way, otherwise it does not make big sense

snuyanzin (changes requested):
Probably some protecting rules are required

bodiam:
> Thanks for your PR
> from my point of view such changes should come together with some automatic rules which could prevent future methods name in similar way, otherwise it does not make big sense

While I don't mind the cleanup, I'm quite opposed to any rule which makes a fail for non-functional reasons, and I don't see myself changing my standpoint on this any time soon.

Also having another build would not be a great solution, it would just be a build which always fails, I'm not going to let a build fail because someone uses a "test" prefix as a method name for example, or has a space too many, or prefixes a test method with public, or any of those reasons, you can have prescriptive rules like this at your work if you fancy them, but not here.

bodiam:
I actually personally prefer methods starting with "test" (or my real preference is "should"), but I see zero reasons to strip the test prefix from methods, so I'm going to close this one.

timtebeek:
No worries; as I said, most of these changes were made automatically, and I understand that (not) using a test prefix is a matter of taste. The additional build step I only added after it was suggested, but understandable that you do not want to throw up any hurdles. In other projects I've seen such automated checks take off some of the review burden, but if it does not fit here that's fine.

You might want to cherry pick https://github.com/datafaker-net/datafaker/pull/454/commits/956e73056464c6f059bd683334586ffde5590095 only to get the static imports for AssertJ, even if it's not enforced through https://github.com/datafaker-net/datafaker/pull/454/commits/99f0f658a67fda5d5c15eee323701bbf33f3705e.
Or you can recreate those changes by executing this command:
```
mvn org.openrewrite.maven:rewrite-maven-plugin:4.36.0:run \
  -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-testing-frameworks:1.30.0 \
  -DactiveRecipes=org.openrewrite.java.testing.assertj.Assertj
```
I'll leave the fork & branch up for a while still, before I clean it up.

Happy to help; no problem if not all of my suggestions make it in. :)

bodiam:
Thanks for your understanding. I feel not great about rejecting PRs knowing they have been created with the best intentions. 

I probably will cherry pick your commits indeed (or feel free to send another PR if it's not too much effort), cause I do like consistency and the static imports would be great to incorporate, I'm just slightly hesitant to make this a standard.

Thanks again!

bodiam:
Ps: absolutely love Open Rewrite. Great tool, and I hope there will be support for Kotlin soon, since our (work) code base (Java+Kotlin) would greatly benefit from some of the migrations.

timtebeek:
Glad to hear you already know and like the tool! :) I've opened https://github.com/datafaker-net/datafaker/pull/456 to separately and only update those imports.

PR #464, opened 2022-10-22
feat/adding_type_to_pokemon_provider

TGITS:
Small feature proposal to add the type to the pokemon provider


bodiam:
Seems like a great proposal, I'll merge it. Are you using the Pokémon for a project or so?

TGITS:
@bodiam Sorry for the seemingly obsession with Pokemon :-) 

My son is a huge fan of Pokemons. Working on some examples for blog posts in French about Datafaker, I've used the Pokemon provider among others and found some minor problem, so I had a look in the source code... one thing leading to another, I have proposed some small fixes, for now quite focused on the Pokemon provider :-).

I have preferred to do very small PR focused on a small topic rather than bigger ones. 

I hope to contribute to the library on other parts than Pokemon if I can.

I must say you are welcoming and you process the PR quickly and kindly. Thanks a lot for that. It is a pleasure to contribute to Datafaker even for very humble fixes or features 👍 .

bodiam:
I love the obsession with Pokémon. Sometimes we have discussions / feedback about that it should be more focused on business topics, but I really like that it's not, and it's great to see that people are using the more exotic fakers too! 

Part of the reason why I started Datafaker was that I was a bit disappointed to see all these open source projects with many unmerged PRs, no feedback, where people puterious effort to fix things and nothing happens with those contributions. 

I'm trying to make it as easy to contribute as possible, and we accept almost any PR to the library. There's a few exceptions to it, but I try to motivate as best as I can why a certain PR might not be a great fit, but I sometimes feel a bit guilty for wasting people's time, so the best I can do is try to give a hopefully clear explanation in a timely fashion.

Thanks again for your contributions, really any contribution is more than welcome!

TGITS:
I agree with you for the use of more exotic provider. Sometimes it is a way to introduce some fun and variability in your test data. 
Of course the main need is for the more focused fake data providers but the exotic ones have value too for testing.

In any case, thanks for datafaker, the work on it and the way the project is managed.

PR #468, opened 2022-10-22
Add icon for Intellijidea toolbox

snuyanzin:
now in toolbox it could be seen as
Screenshot from 2022-10-22 14-47-08
or in recent projects in idea
Screenshot from 2022-10-22 14-48-24


snuyanzin:
@bodiam in case you have better icon we can use it (this one is just a result of compression of existing one to decrease file size)

PR #470, opened 2022-10-23
Add MariaDB sql dialect

panilya:


src/main/java/net/datafaker/transformations/SqlDialect.java
```diff
@@ -12,6 +12,7 @@ public enum SqlDialect {
     LUCIDDB("\""),
     MSSQL("[]"),
     MYSQL("`", Casing.UNCHANGED),
+    MARIADB("`"),
```
snuyanzin, line 15: Please keep abc order

snuyanzin: I guess it should have same casing as Mysql or am I wrong?

panilya: This is why I'm requesting your review. The documentation says that it depends on OS. For Windows, it's not case-sensitive. For Linux, Database, table, table aliases and trigger names are affected by the systems case-sensitivity, while index, column, column aliases, stored routine and event names are never case sensitive.

snuyanzin: Do you have a link to documentation?

panilya: https://mariadb.com/kb/en/library/identifier-case-sensitivity/

snuyanzin: it is stated that
>The [lower_case_table_names](https://mariadb.com/kb/en/server-system-variables/#lower_case_table_names) server system variable plays a key role. It determines whether table names, aliases and database names are compared in a case-sensitive manner. If set to 0 (the default on Unix-based systems), table names and aliases and database names are compared in a case-sensitive manner. If set to 1 (the default on Windows), names are stored in lowercase and not compared in a case-sensitive manner. If set to 2 (the default on Mac OS X), names are stored as declared, but compared in lowercase.

it looks like in that case it is either lowercase or case sensitive => `Casing.To_LOWER` should be applied. Even if this parameter is not set on db side name will be quoted and it is ok

panilya: Fixed

snuyanzin (approved):

PR #479, opened 2022-10-24
Make FakeSequence Iterable

NarekDW:
## Make FakeSequence Iterable

To allow to iterate by fake sequence like this:
```java
FakeSequence<String> digits = faker
        .stream(() -> faker.number().digit())
        .len(10)
        .build();

for (String digit : digits) 
    System.out.println(digit);
```

### DIFF
--- a/src/main/java/net/datafaker/sequence/FakeStream.java
+++ b/src/main/java/net/datafaker/sequence/FakeStream.java
@@ -2,6 +2,7 @@ package net.datafaker.sequence;
 
 import net.datafaker.service.RandomService;
 
+import java.util.Iterator;
 import java.util.List;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
@@ -23,6 +24,11 @@ public class FakeStream<T> extends FakeSequence<T> {
 
     public boolean isInfinite() {
         return maxLength < 0;
+    }
+
+    @Override
+    public Iterator<T> iterator() {
+        return get().iterator();
     }
 
     public static class Builder<T> extends FakeSequence.Builder<T> {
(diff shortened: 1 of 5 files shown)


src/main/java/net/datafaker/sequence/FakeStream.java
```diff
@@ -25,6 +26,11 @@ public boolean isInfinite() {
         return maxLength < 0;
     }
 
+    @Override
+    public Iterator<T> iterator() {
+        return get().iterator();
+    }
+
```
snuyanzin, line 33: Probably it's better to make such change on FakeSequence level like
```java
    public abstract <Sequence extends Iterable<T>> Sequence get();

    public final Iterator<T> iterator() {
        return get().iterator();
    }
```
and `FakeStream`, `FakeCollection` keep untouched

NarekDW: @snuyanzin we can do that, but in that case we have to use Reflection API, as Stream doesn't extend Iterable interface. (It has it's own method with the same name...)
Something like this:
```java
    @Override
    public Iterator<T> iterator() {
        try {
            Object iterable = get();
            Method iterator = iterable.getClass().getMethod("iterator");
            iterator.setAccessible(true);
            //noinspection unchecked
            return (Iterator<T>) iterator.invoke(iterable);
        } catch (Exception e) {
            throw new RuntimeException(e);
        }
    }
```

Is this fine?

snuyanzin: true, ok then let's go with the initial version

NarekDW: Ok, I'm leaving this unchanged

snuyanzin (approved):

PR #483, opened 2022-10-24
[bug-fix] generate separated SQL statements

NarekDW:
## Fix the bug with generation of separated SQL statements.
For example, such code:
```java
BaseFaker faker = new BaseFaker(new Random(10L));
            field("Text", () -> faker.name().firstName()),
            field("Bool", () -> faker.bool().bool())
        );

        SqlTransformer<Object> transformer = new SqlTransformer.SqlTransformerBuilder<>().build();
        String sql = transformer.generate(schema, 2);
```
Computes such a result:
`INSERT INTO "MyTable" ("Text", "Bool") VALUES ('Willis', false)INSERT INTO "MyTable" ("Text", "Bool") VALUES ('Carlena', true);`

But I'm expecting to get:
`INSERT INTO "MyTable" ("Text", "Bool") VALUES ('Willis', false);`
`INSERT INTO "MyTable" ("Text", "Bool") VALUES ('Carlena', true);`

Also, added some little improvement in API of Transformer interface

### DIFF
--- a/src/main/java/net/datafaker/transformations/SqlTransformer.java
+++ b/src/main/java/net/datafaker/transformations/SqlTransformer.java
@@ -1,6 +1,7 @@
 package net.datafaker.transformations;
 
 import java.util.List;
+import java.util.StringJoiner;
 
 public class SqlTransformer<IN> implements Transformer<IN, CharSequence> {
     private static final String INSERT_INTO_UP = "INSERT INTO ";
@@ -45,6 +46,11 @@ public class SqlTransformer<IN> implements Transformer<IN, CharSequence> {
             }
         }
         return false;
+    }
+
+    @Override
+    public CharSequence apply(IN input, Schema<IN, ?> schema) {
+        return apply(input, schema, 0);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


bodiam:
Seems there's a new flaky test in the phone number generators. I'll have a look. I retriggered the build

src/main/java/net/datafaker/transformations/SqlTransformer.java
```diff
@@ -168,19 +174,35 @@ public String generate(List<IN> input, Schema<IN, ?> schema) {
...
-                    sb.append(SqlDialect.getLastRowSuffix(dialect, keywordUpperCase));
-                }
+                sb.append(SqlDialect.getLastRowSuffix(dialect, keywordUpperCase));
                 sb.append(";");
             }
         }
         return sb.toString();
     }
 
+    public String generateSeparatedStatements(Schema<IN, ?> schema, int limit) {
```
snuyanzin, line 196: it's better to have it private otherwise people could use it and we have less freedom for refactoring

src/main/java/net/datafaker/transformations/SqlTransformer.java
```diff
@@ -168,19 +174,35 @@ public String generate(List<IN> input, Schema<IN, ?> schema) {
...
     @Override
     public String generate(Schema<IN, ?> schema, int limit) {
+        if (withBatchMode) {
+            return generateBatchModeStatements(schema, limit);
+        } else {
+            return generateSeparatedStatements(schema, limit);
+        }
+    }
+
+    public String generateBatchModeStatements(Schema<IN, ?> schema, int limit) {
```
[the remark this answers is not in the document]
NarekDW: Yep, my bad

src/main/java/net/datafaker/transformations/SqlTransformer.java
NarekDW: Yep, my bad

NarekDW: fixed

PR #486, opened 2022-10-24
Add compass provider

panilya:


src/main/java/net/datafaker/providers/base/Compass.java
```diff
@@ -0,0 +1,110 @@
+package net.datafaker.providers.base;
+
+/**
+ * @since 1.7.0
+ */
+public class Compass extends AbstractProvider<BaseProviders> {
+
+    enum Point { CARDINAL, ORDINAL, HALF_WIND, QUARTER_WIND }
+
+    private Point point;
```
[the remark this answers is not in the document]
panilya: Then how about `Direction`?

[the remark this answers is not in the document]
snuyanzin: Would be great if you first a bit elaborate the meaning of this enum and what for it's intended to use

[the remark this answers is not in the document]
panilya: The compass rose has different directions, such as Cardinal, Ordinal, Half wind and Quarter wind. Cardinal is four main points of a compass: north, south, east, and west. Ordinal directions refer to the direction found at the point equally between each cardinal direction, and so on. 

I introduced an enum in order to not have a lot of methods like this:
```
   public String cardinalDirectionWord() {
        return resolve("compass.cardinal.word");
    }

    public String cardinalDirectionAbbreviation() {
        return resolve("compass.cardinal.abbreviation");
    }

    public String cardinalDirectionAzimuth() {
        return resolve("compass.cardinal.azimuth");
    }

    public String ordinalDirectionWord() {
        return resolve("compass.ordinal.word");
    }

    public String ordinalDirectionAbbreviation() {
        return resolve("compass.ordinal.abbreviation");
    }

    public String ordinalDirectionAzimuth() {
        return resolve("compass.ordinal.azimuth");
    }
```

And instead of this, just specify the direction. I hope I've explained my point of view.

src/main/java/net/datafaker/providers/base/Compass.java
```diff
@@ -0,0 +1,110 @@
...
+     * @return the full-size direction word according to the specified {@link CompassPoint}. Otherwise, the point of direction will be selected randomly.
+     */
+    public String word() {
+        if (compassPoint == null) {
+            return resolve("compass.direction");
+        }
+
+        switch (compassPoint) {
+            case CARDINAL:
+                return resolve("compass.cardinal.word");
```
bodiam, line 43: I think it would be better to get remove these switch statements, and create the resolve string by templating this string, something like:

`"compass." + compassPoint.key + ".word`

panilya: Fixed

bodiam: Nice. But now the default fallback you had before is gone. Was that not needed?

panilya: This is not necessary because each method checks for null, and if the user has not specified a direction, the null check should work and use a random direction

PR #496, opened 2022-10-25
Remove redundant bank.yml file

panilya:
In `resources.en` there is a file `bank.todo.yml`. This file basically contains IBAN formats for countries. But, there is already implemented logic for this (in `Finance` faker), so there is no point to have this file in the project.


epragtbeamtree:
Makes sense to me to remove it, nice work!

PR #506, opened 2022-11-01
Add Gaussian (normal) distribution to the RandomService

NarekDW:
Hi guys .

What do you think about adding different types of distributions to the `RandomService`, which could be used further by different providers?

For example you can review the one which I've added, this is a Gaussian (normal) distribution model, which could generate values with such a properties:
~ 68 % of generated values will belong to the segment [mean - deviation, mean + deviation]
~ 95 % of generated values will belong to the segment [mean - 2 * deviation, mean + 2 * deviation]
~ 99 % of generated values will belong to the segment [mean - 3 * deviation, mean + 3 * deviation]

by default, mean = 0 and deviation = 1.0

Also we could add:
 - [Bernoulli distribution](https://en.wikipedia.org/wiki/Bernoulli_distribution)
 - [Discrete uniform distribution](https://en.wikipedia.org/wiki/Discrete_uniform_distribution)
 - and so on...

### DIFF
--- a/src/main/java/net/datafaker/service/RandomService.java
+++ b/src/main/java/net/datafaker/service/RandomService.java
@@ -20,7 +20,6 @@ public class RandomService {
         this.random = random != null ? random : SHARED_RANDOM;
     }
 
-    @SuppressWarnings("unused")
     public int nextInt() {
         return random.nextInt();
     }
@@ -33,7 +32,6 @@ public class RandomService {
         return random.nextInt((max - min) + 1) + min;
     }
 
-    @SuppressWarnings("unused")
     public float nextFloat() {
         return random.nextFloat();
     }
@@ -67,6 +65,36 @@ public class RandomService {
 
     public double nextDouble(double min, double max) {
         return min + (nextDouble() * (max - min));
+    }
+
+    /**
+     * Gaussian or normal distribution.
+     * @see <a href="https://en.wikipedia.org/wiki/Normal_distribution">Normal distribution</a>
+     * ~ 68 % of generated values will belong to the segment [mean - deviation, mean + deviation]
+     * ~ 95 % of generated values will belong to the segment [mean - 2 * deviation, mean + 2 * deviation]
+     * ~ 99 % of generated values will belong to the segment [mean - 3 * deviation, mean + 3 * deviation]
+     *
+     * @return Returns the next pseudorandom, Gaussian ("normally") distributed double value with mean 0.0
+     * and standard deviation 1.0 from this random number generator's sequence.
+     */
+    public double nextGaussian() {
+        return random.nextGaussian();
+    }
+
+    /**
+     * Gaussian or normal distribution with predefined mean and standard deviation.
+     * @see <a href="https://en.wikipedia.org/wiki/Normal_distribution">Normal distribution</a>
+     * ~ 68 % of generated values will belong to the segment [mean - deviation, mean + deviation]
+     * ~ 95 % of generated values will belong to the segment [mean - 2 * deviation, mean + 2 * deviation]
+     * ~ 99 % of generated values will belong to the segment [mean - 3 * deviation, mean + 3 * deviation]
+     *
+     * @param mean      the mean
+     * @param deviation the standard deviation
+     * @return Returns the next pseudorandom, Gaussian ("normally") distributed double value with mean {@code mean}
+     * and standard deviation {@code deviation} from this random number generator's sequence.
+     */
+    public double nextGaussian(double mean, double deviation) {
+        return mean + deviation * nextGaussian();
     }
 
     public boolean nextBoolean() {
--- a/src/main/java/net/datafaker/service/RandomService.java
+++ b/src/main/java/net/datafaker/service/RandomService.java
@@ -67,31 +67,14 @@ public class RandomService {
         return min + (nextDouble() * (max - min));
     }
 
-    /**
-     * Gaussian or normal distribution.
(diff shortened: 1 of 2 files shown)


bodiam:
That's actually on my list for a long time. For example, it would make a lot of sense when generating ages or so. At the moment, there will be the same amount of 2 year olds as 40 year olds, and something like this would be really great. 

I haven't checked yet why the build is failing, need some help?

NarekDW:
@bodiam the first run of CI failed because of `net.datafaker.providers.base.CNPJTest.invalid_multiBranchIsTrue_shouldGenerateCNPJWithBranchNumberGreaterThan0001` test. Seems it is a  flaky test. 1 out of 1k or 10k runs it fails for me on my local machine...

Could you relaunch  CI please?

src/main/java/net/datafaker/service/RandomService.java
```diff
@@ -69,6 +67,36 @@ public double nextDouble(double min, double max) {
...
+    /**
+     * Gaussian or normal distribution.
+     * @see <a href="https://en.wikipedia.org/wiki/Normal_distribution">Normal distribution</a>
+     * ~ 68 % of generated values will belong to the segment [mean - deviation, mean + deviation]
+     * ~ 95 % of generated values will belong to the segment [mean - 2 * deviation, mean + 2 * deviation]
+     * ~ 99 % of generated values will belong to the segment [mean - 3 * deviation, mean + 3 * deviation]
+     *
+     * @return Returns the next pseudorandom, Gaussian ("normally") distributed double value with mean 0.0
+     * and standard deviation 1.0 from this random number generator's sequence.
+     */
```
snuyanzin, line 79: We do not master this... and we do not track if anything is changing between jdk's versions
That's why it seems it's better to refer to jdk's javadoc

src/test/java/net/datafaker/service/RandomServiceTest.java
```diff
@@ -92,6 +92,57 @@ void testDoubleInRange(RandomService randomService) {
...
+
+        for (int i = 0; i < totalAmount; i++) {
+            double value = randomService.nextGaussian(testMean, testDeviation);
+            if (value >= testMean - testDeviation && value <= testMean + testDeviation) oneSigmaCount++;
+            if (value >= testMean - 2 * testDeviation && value <= testMean + 2 * testDeviation) twoSigmaCount++;
+            if (value >= testMean - 3 * testDeviation && value <= testMean + 3 * testDeviation) threeSigmaCount++;
+        }
+
+        double oneSigmaProbability = (oneSigmaCount * 1.0) / totalAmount;
+        double twoSigmaProbability = (twoSigmaCount * 1.0) / totalAmount;
```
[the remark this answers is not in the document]
NarekDW: From my point of view, this is just to be sure, that all the JDKs are producing correct results, but we can get rid of this tests, no problem for me.

bodiam:
So, I don't mind adding this to Datafaker,  but what I was thinking of was a random generator which uses the current providers, but using a Gaussian distribution. With a `nextGaussian` method I can't really influence the generation of the fake values right? For example, in my earlier example: if I want to generate an age, I currently get an evenly distributed generated age, but I'd rather use a normal distribution for this.

src/main/java/net/datafaker/service/RandomService.java
```diff
@@ -69,6 +67,19 @@ public double nextDouble(double min, double max) {
         return min + (nextDouble() * (max - min));
     }
 
+    public double nextGaussian() {
```
bodiam, line 70: What's the use case here, and why is this better than using `getRandomInternal().nextGaussian()` ?

NarekDW:
> So, I don't mind adding this to Datafaker, but what I was thinking of was a random generator which uses the current providers, but using a Gaussian distribution. With a `nextGaussian` method I can't really influence the generation of the fake values right? For example, in my earlier example: if I want to generate an age, I currently get an evenly distributed generated age, but I'd rather use a normal distribution for this.

Yep, for now it doesn't affect on generation of fake values, it is just an additional function, which could be used in some specific providers directly, as an additional option...
For example if we want to generate height of a person we can use it directly in provider, something like this:
```java
public Double height() {
    return faker.random().nextGaussian(175, 10);
}
```

snuyanzin:
but in fact right now it also could be used like 
```java
public Double height() {
        return faker.random().getRandomInternal().nextGaussian(175, 10);
    }
```

src/main/java/net/datafaker/service/RandomService.java
NarekDW: I think, this class is just another level of abstraction where all the types of generator functions are gathered. Why don't we use `getRandomInternal().nextInt()` instead of `nextInt()` method? :)

bodiam: Because `getRandomInternal().nextInt()` is new, and `nextInt()` was already there for several years, so removing it would have been hard. 

`getRandomInternal()` was introduced so it can be used to expose things which aren't currently exposed through Faker, such as, say, `nextGaussian()` ;-)

snuyanzin: no idea why, may be need to go to git history to answer that question, however what is the benefit here or what kind task will it solve ?
However based on code methods from this class are usually used internally while `faker.number` is more like public..
Since `RandomService` is public the rule is likely not 100% satisfied

NarekDW: I see, that from your point of view guys, these changes don't make much sense for now, will close PR then, until it might be required.

PR #512, opened 2022-11-01
Support transformation schema for YAML

RVRhub:
The PR introduces yaml transformer.
An example of usage

```
Faker faker = new Faker();
System.out.println(
    new YamlTransformer<Name>().generate(
        faker.collection(faker::name).len(1).build(),
        Schema.of(field("name", Name::fullName), field("username", Name::username))));
```

Result:
```
name: Mabelle Prohaska
username: clara.lemke
```

### DIFF
--- a/src/main/java/net/datafaker/transformations/YamlTransformer.java
+++ b/src/main/java/net/datafaker/transformations/YamlTransformer.java
@@ -1,0 +1,100 @@
+package net.datafaker.transformations;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.StringJoiner;
+
+import net.datafaker.sequence.FakeSequence;
+
+public class YamlTransformer<IN> implements Transformer<IN, CharSequence> {
+
+    private static final String INDENTATION = "  ";
+
+    @Override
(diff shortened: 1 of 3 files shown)


src/main/java/net/datafaker/transformations/YamlTransformer.java
```diff
@@ -0,0 +1,100 @@
...
+        return data.toString();
+    }
+
+    @Override
+    public String generate(Schema<IN, ?> schema, int limit) {
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < limit; i++) {
+            sb.append(apply(null, schema));
+            if (i < limit - 1) {
+                sb.append("\n");
```
snuyanzin, line 46: Probably here should be `LINE_SEPARATOR` as well
```suggestion
                sb.append(LINE_SEPARATOR);
```

snuyanzin:
hi @RVRhub nice to see you here
thanks for your contribution
in general lgtm, i left a nit comment

src/main/java/net/datafaker/transformations/YamlTransformer.java
RVRhub: Replaced line separator

PR #513, opened 2022-11-01
Add java object transformer

snuyanzin:
Another transformer which could fill java objects
e.g. this will generate a collection of Person objects according to schema
```java
JavaObjectTransformer jTransformer = new JavaObjectTransformer();
            field("firstName", () -> faker.name().firstName()),
            field("lastName", () -> faker.name().lastName()),
            field("birthDate", () -> faker.date().birthday()),
            field("id", () -> faker.number().positive()));

        Collection<Person> persons = new ArrayList<>();
        for (int i = 0; i < 10; i++) {
            persons.add(jTransformer.apply(Person.class, schema));
        }
```

cc @bodiam since you were looking for something like that afair


bodiam:
Nice. I'm still a bit more a fan of what EasyRandom does, eg if you have a structure like this:

```
data class Person(val addresses: List<Address>)

data class Address(val street: String, val city: String, val country: String)    
```

That we can generate something by doing:

```
val person = faker.populate(Person())
```

bodiam:
@snuyanzin  what do you want to do with this PR?

snuyanzin:
let's merge it

PR #515, opened 2022-11-02
Add tests for uncovered method in Time provider

panilya:
Coverage of `Time` provider is 100% now

### DIFF
--- a/src/test/java/net/datafaker/providers/base/TimeTest.java
+++ b/src/test/java/net/datafaker/providers/base/TimeTest.java
@@ -84,6 +84,15 @@ public class TimeTest extends BaseFakerTest<BaseFaker> {
     }
 
     @Test
+    void testBetweenWithMask() {
+        String pattern = "mm:hh:ss";
+        LocalTime now = LocalTime.now();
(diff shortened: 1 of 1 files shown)


src/test/java/net/datafaker/providers/base/TimeTest.java
```diff
@@ -75,6 +83,16 @@ void testBetweenThenLargerThanNow() {
             .hasMessage("Invalid time range, the upper bound time is before the lower bound.");
     }
 
+    @Test
+    void testBetweenWithPattern() {
```
snuyanzin, line 87: what's difference between this test and existing `testBetween`?

panilya: This test for `between` with pattern

snuyanzin: There are 2 tests below testing this for future and past
It would make sense to similar test naming like `betweenWithMask`
and similar style for code of the test. It would simplify reading and understanding for others in future

panilya: Fixed

snuyanzin (approved):

PR #523, opened 2022-11-03
Add more tests

panilya:


src/test/java/net/datafaker/service/FakerIDNTest.java
```diff
@@ -2,12 +2,21 @@
...
     @Test
     void toASCIINoError() {
         assertThat(FakerIDN.toASCII("hello")).isEqualTo("hello");
     }
+
+    @Test
+    void toASCIIEmptyInput() {
+        assertThatThrownBy(() -> FakerIDN.toASCII(""))
+            .isInstanceOf(RuntimeException.class);
+    }
```
panilya, line 20: What input to pass in order to get `RuntimeException`?

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
```diff
@@ -241,6 +235,38 @@ void testLocalesChainGetterRu() {
...
+        } else {
+            assertThat(fc1).isNotEqualTo(fc2);
+        }
+    }
+
+    static Stream<Arguments> fakerContexts() {
+        return Stream.of(
+            Arguments.of(new FakerContext(new Locale("en"), new RandomService(new Random(1))), new FakerContext(new Locale("uk"), new RandomService(new Random(1))), false),
+            Arguments.of(new FakerContext(new Locale("en"), new RandomService(new Random(1))), null, false),
+            Arguments.of(new FakerContext(Locale.US, new RandomService(new Random(1))), new FakerContext(Locale.US, new RandomService(new Random(1))), true)
```
panilya, line 266: I don't know why, but this one fails. Says that it's not equal

snuyanzin: randomservice and random do not have equals method

panilya: Oh. In that case, maybe there's another way of implementing such test?

snuyanzin: compare by references instead of value

panilya: Take a look please if this is good approach

src/test/java/net/datafaker/FakerTest.java
```diff
@@ -283,6 +286,8 @@ void fakerInstanceCanBeAcquiredViaUtilityMethods() {
         assertThat(new Faker(Locale.CANADA)).isInstanceOf(BaseFaker.class);
         assertThat(new Faker(new Random(1))).isInstanceOf(BaseFaker.class);
         assertThat(new Faker(Locale.CHINA, new Random(2))).isInstanceOf(BaseFaker.class);
+        assertThat(new Faker(Locale.US, new RandomService())).isInstanceOf(BaseFaker.class);
+        assertThat(new Faker(new FakeValuesService(), new FakerContext(Locale.JAPAN, new RandomService()))).isInstanceOf(BaseFaker.class);
```
snuyanzin, line 290: what is the reason of such tests?
any `new Faker` as an instance of `BaseFaker`

src/test/java/net/datafaker/formats/SqlTest.java
```diff
@@ -281,6 +297,21 @@ void batchSqlTestForSqlTransformerPostgres() {
...
+    @Test
+    void sqlCasingCheck() {
+        BaseFaker faker = new BaseFaker();
+        Schema<String, String> schema =
+            Schema.of(field("firstName", () -> faker.name().firstName()),
+                field("lastName", () -> faker.name().lastName()));
+
+        SqlTransformer<String> transformer =
+            new SqlTransformer.SqlTransformerBuilder<String>()
+                .casing(Casing.TO_UPPER)
```
snuyanzin, line 309: in this test case this method call does not make any sense

src/test/java/net/datafaker/FakerTest.java
snuyanzin: test coverage is not a good metric.
There is a much better metric to answer question if a test is a good one or not.
A good test is test which is failing otherwise it does not bring much value ...

panilya: Now I get it, test is deleted

snuyanzin:
@panilya could you please next time squash your commits before merging to main? 
It would simplify searching through git history to understand where(what issue/pr) which change was done

panilya:
Okay, sorry for that

PR #524, opened 2022-11-04
 Adds Microsoft Azure provider

agoncal:
Based on some of the most common Azure services naming convention (https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-abbreviations)


bodiam:
Hi @agoncal , great to see you here! Thanks for your contribution!

src/main/java/net/datafaker/providers/base/Azure.java
```diff
@@ -0,0 +1,147 @@
+package net.datafaker.providers.base;
+
+
+/**
+ * Generates data for Azure services. This is based on the Azure best practices of naming conventions:
+ * <a href="https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-naming">...</a>
```
bodiam, line 6: Small mistake here, I think you used IntelliJ to autofix the urls? (I had a similar issue.)

agoncal: Yes, Intellij :o) Thanks for fixing it

PR #529, opened 2022-11-04
Add One Piece provider

panilya:


src/main/java/net/datafaker/providers/movie/MovieProviders.java
```diff
@@ -140,6 +140,10 @@ default OscarMovie oscarMovie() {
         return getProvider(OscarMovie.class, OscarMovie::new);
     }
 
+    default OnePiece onePiece() {
```
snuyanzin, line 143: keep it in abc order

Issue #547, opened 2022-11-10
Baseball feature request


Blackcows:
**Is your feature request related to a problem? Please describe.**
 We expected a provider which could be frequently used, Baseball but could not find on this repository. 

**Describe the solution you'd like**
To solve this problem, we created Baseball Provider under Sports category which will provides team names, coaches, positions and player names. 

**Additional context**
We hope our code being contribute to improve data-faker project. Thank you in advance, and thanks for maintaining this project.

PR #548, opened 2022-11-10
Add Baseball provider

Blackcows:
Added Baseball provider under Sport package with issue #547

### DIFF
--- a/src/main/java/net/datafaker/providers/sport/Baseball.java
+++ b/src/main/java/net/datafaker/providers/sport/Baseball.java
@@ -1,0 +1,59 @@
+package net.datafaker.providers.sport;
+
+import net.datafaker.providers.base.AbstractProvider;
+
+/**
+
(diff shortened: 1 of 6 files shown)


src/main/java/net/datafaker/providers/sport/Baseball.java
```diff
@@ -0,0 +1,59 @@
+package net.datafaker.providers.sport;
+
+import net.datafaker.providers.base.AbstractProvider;
+
+/**
+
+ * Generate random components of baseball game, e.g. teams, coaches, positions and players.
+ *
+ * @author unknown and irakatz
+ * @since 0.8.0
```
bodiam, line 10: Small issue here with the version number, this should be 1.7.0 I think?

Blackcows: Yeah, I mean 1.7.0. I'll change the version number.

PR #561, opened 2022-11-27
Generate boolean with probability

NarekDW:
## Add ability to generate boolean value with some predefined probability of true

Example: 
```java
Faker faker = new Faker().withProbabilityConfig(new ProbabilityConfig().withBool(0.0));
boolean falseValue = faker.bool().bool(); // always false

faker = faker.withProbabilityConfig(new ProbabilityConfig().withBool(1.0));
boolean trueValue = faker.bool().bool(); // always true

faker = faker.withProbabilityConfig(new ProbabilityConfig().withBool(0.7));
boolean value = faker.bool().bool(); // will generate true with probability 70 % and false with probability 30 %
```

### DIFF
--- a/src/main/java/net/datafaker/configuration/ProbabilityConfig.java
+++ b/src/main/java/net/datafaker/configuration/ProbabilityConfig.java
@@ -1,0 +1,22 @@
+package net.datafaker.configuration;
+
+public class ProbabilityConfig {
+    public final static double DEFAULT_BOOLEAN_PROBABILITY = 0.5;
+    private final double boolProbability;
+
+    public ProbabilityConfig() {
+        this(DEFAULT_BOOLEAN_PROBABILITY);
+    }
+
+    private ProbabilityConfig(double boolProbability) {
+        this.boolProbability = boolProbability;
+    }
+
+    public ProbabilityConfig withBool(double probability) {
+        return new ProbabilityConfig(probability);
+    }
+
+    public double getBoolProbability() {
+        return boolProbability;
+    }
+}
--- a/src/main/java/net/datafaker/configuration/ProbabilityConfig.java
+++ b/src/main/java/net/datafaker/configuration/ProbabilityConfig.java
@@ -3,32 +3,27 @@ package net.datafaker.configuration;
 import java.util.*;
 
 public class ProbabilityConfig {
-    public final static double DEFAULT_BOOLEAN_PROBABILITY = 0.5;
-    private final double boolProbability;
-    private final Map<Integer, Double> intProbability = new HashMap<>();
+    private final Map<Object, Double> probabilities = new HashMap<>();
 
     public ProbabilityConfig() {
-        this(DEFAULT_BOOLEAN_PROBABILITY);
     }
 
-    private ProbabilityConfig(double boolProbability) {
-        this.boolProbability = boolProbability;
-    }
-
-    public ProbabilityConfig withBool(double probability) {
-        return new ProbabilityConfig(probability);
-    }
-
-    public ProbabilityConfig withInt(int value, double probability) {
-        intProbability.put(value, probability);
+    public ProbabilityConfig with(boolean value, double probability) {
+        probabilities.put(value, probability);
         return this;
     }
 
-    public double getBoolProbability() {
-        return boolProbability;
+    public ProbabilityConfig with(int value, double probability) {
+        probabilities.put(value, probability);
+        return this;
     }
 
-    public Map<Integer, Double> getIntProbability() {
-        return intProbability;
+    public ProbabilityConfig with(Object value, double probability) {
+        probabilities.put(value, probability);
+        return this;
+    }
+
+    public Map<Object, Double> getProbabilities() {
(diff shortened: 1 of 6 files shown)


bodiam:
Hi @NarekDW . Thanks for this PR, but I'm not really sure about it. This PR is for booleans, but what about numbers? What about ages? What about enums? Or even what about coin flips?

This would require different versions of every method out there, and I'm not sure if that's ideal. What might be better if to change the underlying RandomService instead, and control the behavior through that.

I'm not sure how easy this would be, but it might be a better approach than this. What do you think?

NarekDW:
Hi @bodiam , I'm not sure if I understood you, but do you mean to move this method:
```java
    public boolean bool(double probability) {
        if (probability < 0.0 || probability > 1.0) {
            throw new IllegalArgumentException("Probability parameter should be between [0.0, 1.0]: " + probability);
        }

        return faker.random().nextDouble() < probability;
    }
```
to the `RandomService` (as `nextBoolean(double probability)` method) class, and use it from different classes, right? (including Bool class)

bodiam:
@NarekDWar well, I think we can't change the signature of every method out there just to change the randomness, but I was hoping that maybe by using the RandomService somehow for this, we  wouldnt need to change every method, but the logic could be configured in the Service instead

NarekDW:
Then it could be added to the RandomService class as an additional (overloaded) method, like this:
```java
    public boolean nextBoolean() {
        return random.nextBoolean();
    }

    public boolean nextBoolean(double probability) {
        if (probability < 0.0 || probability > 1.0) {
            throw new IllegalArgumentException("Probability parameter should be between [0.0, 1.0]: " + probability);
        }

        return random.nextDouble() < probability;
    }
```
And used from Bool class, like
```java
    public boolean bool(double probability) {
       return faker.random().nextBoolean(probability);
    }
```

snuyanzin:
Probably to rephrase @bodiam 's idea if i understand it correctly it would make sense to have something like `ProbabilityConfig` which should be taken into account. E.g. 
```java
ProbabilityConfig.with(1, 0.5).with(2, 0.1).with(3, 0.2);
...
```
should mean that in case of numbers it should generate 1 with probability 0.5, 2 with probability 0.1, 3 with probability 0.2 and any other number with probability 0.2.
Similar logic could be applied to other entities

NarekDW:
> Probably to rephrase @bodiam 's idea if i understand it correctly it would make sense to have something like `ProbabilityConfig` which should be taken into account. E.g.
> 
> ```java
> ProbabilityConfig.with(1, 0.5).with(2, 0.1).with(3, 0.2);
> ...
> ```
> 
> should mean that in case of numbers it should generate 1 with probability 0.5, 2 with probability 0.1, 3 with probability 0.2 and any other number with probability 0.2. Similar logic could be applied to other entities

At first glance this looks a bit complicated regarding implementation/integration, but okay, I got the point, will try to implement something like that...

NarekDW:
@bodiam  @snuyanzin I've added a draft version of such a configuration of probabilities, please review the approach.
Also, I'm not sure regarding the API, is it designed convenient or not...
Currently it's possible to configure boolean values distribution only, but it could be enhanced easily for different distribution types for numbers...

snuyanzin:
I didn't get hoe with current implementation it is possible to configure possibilities for 3 numbers like in the example above
```
ProbabilityConfig.with(1, 0.5).with(2, 0.1).with(3, 0.2);
...
```

NarekDW:
@snuyanzin take a look on the last commit (it's a draft example)

snuyanzin:
why should it be separate for bool and int? otherwise next question would be about long, string and etc...
Config should be same for any type. Implementation of generation could be different

NarekDW:
> why should it be separate for bool and int? otherwise next question would be about long, string and etc... Config should be same for any type. Implementation of generation could be different

I'm not sure if I understood the question regarding separation. Do you mean methods `withInt` and `withBool` should be named the same (like `with(...)`) ?

snuyanzin:
mostly about information storage 
```java
 private final double boolProbability;
 private final Map<Integer, Double> intProbability = new HashMap<>();
```

NarekDW:
It's just convenient as for me. 
What could be the alternative?

snuyanzin:
however each new type requires a new map and it will grow endlessly

NarekDW:
By the way, I think probabilities should be configured only in case if we generate some primitive types (Bool, Number, etc).
I can't imagine how it could be used with providers, like Name, Address, Book, Coin and so on...

NarekDW:
> 

Yep, I understand but it is a bit strange to configure probability for objects, like Address, Name, and so on, isn't it?

snuyanzin:
i would consider it from technical point of view.
Everything which is taken from yml configuration or `Option.option` is present as an array. 
So technically speaking we are talking not about name,address or whatever. We are talking only about index which will be random;y defined

NarekDW:
Ok, let me see

NarekDW:
Closing the PR as there are no agreement regarding probability configuration approach.

PR #567, opened 2022-12-07
Add Hexadecimal in providers/base/Number

dlwnsrb0829:


bodiam:
That doesn't look right :)

PR #570, opened 2022-12-07
Add Number bases provider

dlwnsrb0829:


src/main/java/net/datafaker/providers/base/BaseProviders.java
```diff
@@ -54,6 +54,7 @@ default Azure azure() {
     default Barcode barcode() {
         return getProvider(Barcode.class, Barcode::new);
     }
+
```
snuyanzin, line 57: Do we need this empty line?

dlwnsrb0829: We don't need it

snuyanzin:
1. it looks like it uses a new `Random` that means that it will break reproduction.
2. There is already available implementation of `hex` at https://github.com/datafaker-net/datafaker/blob/f1b90d44a8d45684cecc01e33a6ea7d7e2cea43b/src/main/java/net/datafaker/service/RandomService.java#L88-L104
which is much faster (several times faster then implementation in this PR). 
I would suggest to add implementation of octal and binary in a similar way

dlwnsrb0829:
Okay, I'll try.
Thank you :)

PR #571, opened 2022-12-13
Add avatar style sets

lyrjie:
[RoboHash](https://robohash.org/), used to generate avatars, currently supports 4 different sets of generated images
Added `Faker.avatar().image(Set)` that allows to use different sets

### DIFF
--- a/src/main/java/net/datafaker/providers/movie/Avatar.java
+++ b/src/main/java/net/datafaker/providers/movie/Avatar.java
@@ -6,6 +6,8 @@ import net.datafaker.providers.base.AbstractProvider;
  * @since 0.8.0
  */
 public class Avatar extends AbstractProvider<MovieProviders> {
+
+    private static final Set DEFAULT_SET = Set.ROBOTS;
     private final String baseUrl;
 
     protected Avatar(MovieProviders faker) {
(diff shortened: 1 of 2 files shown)


src/main/java/net/datafaker/providers/movie/Avatar.java
```diff
@@ -14,6 +16,23 @@ protected Avatar(MovieProviders faker) {
...
     public String image() {
-        return baseUrl + faker.regexify("[a-z]{8}") + ".png";
+        return image(DEFAULT_SET);
+    }
+
+    public String image(Set set) {
+        return baseUrl + faker.regexify("[a-z]{8}") + "?set=set" + set.key + ".png";
+    }
+
+    public enum Set {
```
bodiam, line 26: Not sure about this.

1) I'm not a fan of the Set classname, it's confusing with the collections Set.
2) It's now directly tied to robohash, which is maybe not ideal. We had to change the avatar service before, I prefer not to introduce something which may break in the future.

Also, it's not really clear to me: why should we need this functionality?

bodiam:
Hi @lyrjie , thanks for your PR. I'm however not really sure about it, and I left some code reviews. What's the use case you're trying to address?

lyrjie:
@bodiam 
The use case was to have some flexibility in the resulting images, since the underlying API allows for it.
But uou rise a valid point; I failed considered the implications of tying library API to concrete provider.
I'll close the PR now, all things considered it's not worth it.

bodiam:
No worries! I do appreciate the contributions, and I'm sorry it didn't work out this time!

PR #585, opened 2022-12-30
Fix wrong names in ko.yml

benelog:
I found wrong names in ko.yml, and fixed it.


bodiam:
Wow, thanks for fixing, much appreciated!

Issue #586, opened 2023-01-01
Documentation generator creates wrong links 


bodiam:
The links in the Documentation generated here are all giving a 404:

https://www.datafaker.net/documentation/providers/

PR #588, opened 2023-01-01
Fix generation of invalid links in ProvidersDocsGenerator

panilya:
Issue: #586

### DIFF
--- a/src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
+++ b/src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
@@ -17,6 +17,7 @@ import java.io.Writer;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashSet;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeSet;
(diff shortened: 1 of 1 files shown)


src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -170,7 +170,7 @@ private static String generateBaseHeaderRow(char padSymbol, Function<Integer, St
 
         public static String generateRow(char padSymbol, String clazzName, String javaDocComment, String groupName) {
             StringBuilder sb = new StringBuilder();
-            sb.append("|").append(generateColumn(getName(clazzName), padSymbol, NAME.length)).append("|");
+            sb.append("|").append(generateColumn(getName(clazzName, groupName.toLowerCase()), padSymbol, NAME.length)).append("|");
```
snuyanzin, line 173:
```suggestion
            sb.append("|").append(generateColumn(getName(clazzName, groupName.toLowerCase(Locale.ROOT)), padSymbol, NAME.length)).append("|");
```
otherwise it could behave different in some locales e.g. Turkey

panilya: but Datafaker's package names are always in the English locale. Or I'm missing something? 

Like there are `base`, `food`, `movie`, `sport`, `videogame` packages, and I don't think they will be in other locales

snuyanzin: it has nothing to do with datafaker's locales. It's running via `net.datafaker.script.ProvidersDocsGenerator#main` so in case of `toUpperCase()`, `toLowerCase()` it uses default system locale.  So if someone sets Turkish locale as default one it will be used

panilya: Got it

snuyanzin (approved):

bodiam:
❤️ for fixing!

Issue #595, opened 2023-01-04
Non-standard language tags for "Belarusian" and "English (Pakistan)" locales


alx-mag:
Faker uses tags `by` and `en-PAK` for these locales. But the standard ones are `be` (or `be-BY`) and `en-PK`, see the [standard locales list](https://www.localeplanet.com/icu/index.html). These non-standard tags are usable with Faker but Java does not recognize them:
```
import java.util.Locale;

public class Test {

    public static void main(String[] args) {
        System.out.println(getLocaleDisplayName("by"));     // by
        System.out.println(getLocaleDisplayName("be-BY"));  // Belarusian (Belarus)
        System.out.println(getLocaleDisplayName("be"));     // Belarusian

        System.out.println(getLocaleDisplayName("en-PAK")); // pak
        System.out.println(getLocaleDisplayName("en-PK"));  // English (Pakistan)
    }

    static String getLocaleDisplayName(String languageTag) {
        return Locale.forLanguageTag(languageTag)
                .getDisplayName(Locale.ENGLISH);
    }
}
```
I think there should be added `be.yml` (or `be-BY.yml`) and `en-PK.yml`. The `by.yml` and `en-PAK.yml` can be left for backward compatibility.

bodiam:
Thanks for reporting! Would you be willing to do a PR for this? We could deprecate the non standard ymls, and remove them next release

bodiam:
I guess this can be closed now that #598 has been merged.

PR #598, opened 2023-01-09
Add .yml files for 'be' and 'en-PK' locales

alx-mag:
Changes for https://github.com/datafaker-net/datafaker/issues/595


bodiam:
Thanks, looks good to me. You can test the snapshot version if you want, it's automatically published.

PR #631, opened 2023-01-17
Introduce singletone FLocale to use identity hash maps

snuyanzin:
This could bring some perf improvements

### DIFF
--- a/src/main/java/net/datafaker/internal/helper/FLocale.java
+++ b/src/main/java/net/datafaker/internal/helper/FLocale.java
@@ -1,0 +1,48 @@
+package net.datafaker.internal.helper;
+
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+
+public class FLocale {
+    private static final Map<Locale, FLocale> LOCALE2FLOCALE = new HashMap<>();
+    private final Locale locale;
+    private final int hashId = (int) System.nanoTime();
+
+    private FLocale(Locale locale) {
+        this.locale = locale;
+    }
+
+    public static FLocale get(Locale locale) {
+        if (locale == null) {
+            return null;
+        }
+        FLocale res = LOCALE2FLOCALE.get(locale);
+        if (res != null) {
+            return res;
+        }
+        synchronized (FLocale.class) {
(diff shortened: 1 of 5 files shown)


bodiam:
But what is an FLocale?

snuyanzin:
The problem with java's `Locale` is that it is not singletone and can not be used in `IdentityHashMap` where no need to calculate hash for keys and only `equals` is enough.

`FLocale` is a way of having such singletones per Locale. It guarantees that within one jvm/classloader there is only one instance of `FLocale` for  java's `Locale`. It allows to use `IdentityHashMap`.
Finally this trick allows to save time because there is no need to calculate hashcode for `Locale` keys

bodiam:
I'm not sure I follow. I don't think IdentityHashMap is using hashcode nor equals at all, it's using object references afaik.

Still don't really see why it's called FLocale though. F stands for Singleton?

bodiam:
I'm not a big fan of abbreviations where I have to guess what it means, so perhaps SingletonLocale might be a name, or  otherwise just a line of Javadoc could clear things up?

snuyanzin:
makes sense
renamed to `SingletonLocale`

bodiam:
I still don't really understand why there's a hashcode and and an equals in the SingletonLocale if you're only going to use it in an IdentityHashmap. Why did you need those?

snuyanzin:
>I still don't really understand why there's a hashcode and and an equals in the SingletonLocale if you're only going to use it in an IdentityHashmap. Why did you need those?

There are 2 cases
1. `IdentityHashMap` and for this one there is no need for `hashCode`, `equals`.
2. `SingletonLocale` is a field inside `net.datafaker.service.FakerContext` and there are hash maps where `FakerContext` is a key, e.g. `net.datafaker.providers.base.BaseFaker#CLASSES`. For this keys we need to calculate hashcodes as part of calculation of hashcode for `net.datafaker.service.FakerContext`

bodiam:
Ah, I missed that one. Thanks for taking the time to explain!

snuyanzin:
no problem
good that now it seems we are on the same page

PR #648, opened 2023-01-24
Describe provider groups and table with new providers per version

panilya:


panilya:
Why here's also a commit from another PR? How to prevent it?

snuyanzin:
```
git log --pretty=oneline
```
will show your commits in current branch
If you see these commits locally then it's expected.
First make sure you do not have them locally

bodiam:
> Why here's also a commit from another PR? How to prevent it?

I guess you made a branch of the other PR, instead of main?

panilya:
> > Why here's also a commit from another PR? How to prevent it?
> 
> I guess you made a branch of the other PR, instead of main?

Yes, I `git checkout -b` from previous branch :(

PR #649, opened 2023-01-24
Provider groups description and new providers per version table in DOCS

panilya:


src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -14,13 +14,7 @@
...
 import java.io.UncheckedIOException;
 import java.io.Writer;
-import java.util.Arrays;
-import java.util.Comparator;
-import java.util.HashSet;
-import java.util.Locale;
-import java.util.Optional;
-import java.util.Set;
-import java.util.TreeSet;
+import java.util.*;
```
snuyanzin, line 17: it's better not to use asterisk for import for at least 2 reasons
1. performance (compile/build time)
2. clean code

panilya: I see, thank you. Intellij IDEA automatically applies an asterisk for imports

panilya: Thank you. 

Btw, do you know, by any chance, how to fix the problem? When I press shortcuts that involve `Shift+Ctrl` (change language), they are not working because it switches languages. My OS is the latest Ubuntu. Can't find an answer on the Internet

snuyanzin: what if change lang switching to another key stroke e.g. `win + space`?
will it start working after that?

panilya: Yes, it just blocks shortcuts that involve lang switching key stroke

PR #653, opened 2023-01-28
:technologist: JsonTransformer enhancement to support import scenarios

mesketh:
- JsonTransformer didn't have 4 indentation hence, the whitespace noise.
- Added JSONAssert to assist with json assertions. Did not update other test cases to use this.

### DIFF
--- a/src/main/java/net/datafaker/transformations/JsonTransformer.java
+++ b/src/main/java/net/datafaker/transformations/JsonTransformer.java
@@ -10,7 +10,11 @@ public class JsonTransformer<IN> implements Transformer<IN, Object> {
 
     private static final Map<Character, String> ESCAPING_MAP = createEscapeMap();
 
-    private boolean formatAsArray;
+    private static final char[] ARRAY_DELIM = {'[', ']'};
+
+    private static final char[] OBJ_DELIM = {'{', '}'};
+
+    private final boolean formatAsArray;
 
     private JsonTransformer() {
         this(false);
@@ -169,7 +173,7 @@ public class JsonTransformer<IN> implements Transformer<IN, Object> {
         return Collections.unmodifiableMap(map);
     }
     private static char[] delimitedBy(boolean formatAsArray) {
-        return formatAsArray ? new char[]{'[',']'} : new char[] {'{','}'};
+        return formatAsArray ? ARRAY_DELIM : OBJ_DELIM;
     }
 
     public static class JsonTransformerBuilder<IN> {
--- a/src/main/java/net/datafaker/transformations/JsonTransformer.java
+++ b/src/main/java/net/datafaker/transformations/JsonTransformer.java
@@ -56,8 +56,8 @@ public class JsonTransformer<IN> implements Transformer<IN, Object> {
             data.add(apply(in, schema));
         }
 
-        char[] delimitedBy = delimitedBy(formatAsArray);
-        return data.length() > 1 ? delimitedBy[0] + LINE_SEPARATOR + data + LINE_SEPARATOR + delimitedBy[1]
+        char[] wrappers = wrappers(formatAsArray);
+        return data.length() > 1 ? wrappers[0] + LINE_SEPARATOR + data + LINE_SEPARATOR + wrappers[1]
             : data.toString();
     }
 
@@ -71,8 +71,8 @@ public class JsonTransformer<IN> implements Transformer<IN, Object> {
                 sb.append(",").append(LINE_SEPARATOR);
             }
         }
-        char[] delimitedBy = delimitedBy(this.formatAsArray);
-        return limit > 1 ? delimitedBy[0] + LINE_SEPARATOR + sb + LINE_SEPARATOR + delimitedBy[1] : sb.toString();
+        char[] wrappers = wrappers(this.formatAsArray);
+        return limit > 1 ? wrappers[0] + LINE_SEPARATOR + sb + LINE_SEPARATOR + wrappers[1] : sb.toString();
     }
 
     private static void value2String(Object value, StringBuilder sb) {
@@ -172,8 +172,8 @@ public class JsonTransformer<IN> implements Transformer<IN, Object> {
         map.put('\u001F', "\\u001F");
         return Collections.unmodifiableMap(map);
     }
-    private static char[] delimitedBy(boolean formatAsArray) {
-        return formatAsArray ? ARRAY_DELIM : OBJ_DELIM;
+    private static char[] wrappers(boolean wrapAsArray) {
+        return wrapAsArray ? ARRAY_DELIM : OBJ_DELIM;
     }
 
     public static class JsonTransformerBuilder<IN> {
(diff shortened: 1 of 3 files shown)


src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -7,149 +7,185 @@
...
+        sb.append("]");
+        return sb.toString();
+    }
+
+
+    private static Map<Character, String> createEscapeMap() {
+        final Map<Character, String> map = new HashMap<>();
+        map.put('\\', "\\\\");
+        map.put('\"', "\\\"");
+        map.put('\b', "\\b");
```
snuyanzin, line 126: Why do we need to touch these lines within this PR?

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -7,149 +7,185 @@
...
+        map.put('\u0018', "\\u0018");
+        map.put('\u0019', "\\u0019");
+        map.put('\u001A', "\\u001A");
+        map.put('\u001B', "\\u001B");
+        map.put('\u001C', "\\u001C");
+        map.put('\u001D', "\\u001D");
+        map.put('\u001E', "\\u001E");
+        map.put('\u001F', "\\u001F");
+        return Collections.unmodifiableMap(map);
+    }
```
snuyanzin, line 170: Why do we need to touch these lines within this PR?

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -7,149 +7,185 @@
...
+    private static char[] delimitedBy(boolean formatAsArray) {
+        return formatAsArray ? new char[]{'[',']'} : new char[] {'{','}'};
```
snuyanzin, line 172: It seems it creates a new array each call which is not good for performance.
There are cases when we generate millions of rows and this will slow this down

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -7,149 +7,185 @@
...
+            sb.append("null");
+        } else if (value instanceof Integer
+            || value instanceof Long
+            || value instanceof Short
+            || value instanceof BigInteger
+            || value instanceof Boolean
+            || (value instanceof Double
             && BigDecimal.valueOf((Double) value).remainder(BigDecimal.ONE).doubleValue() == 0)
-        || (value instanceof BigDecimal
+            || (value instanceof BigDecimal
```
snuyanzin, line 84: Why do we need to touch these lines within this PR?

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -7,149 +7,185 @@
...
-      } else {
-        value2String(((SimpleField) fields[i]).transform(input), sb);
-      }
-      if (i < fields.length - 1) {
-        sb.append(", ");
-      }
+
+    private static final Map<Character, String> ESCAPING_MAP = createEscapeMap();
+
+    private boolean formatAsArray;
```
snuyanzin, line 13: it should be `final`

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -7,149 +7,185 @@
...
+            } else {
+                value2String(((SimpleField) fields[i]).transform(input), sb);
+            }
+            if (i < fields.length - 1) {
+                sb.append(", ");
+            }
+        }
+        sb.append("}");
+        return sb.toString();
     }
```
snuyanzin, line 42: Why do we need to touch these lines within this PR?

src/test/java/net/datafaker/formats/JsonTest.java
```diff
@@ -30,9 +41,9 @@ void testGenerateFromSchemaWithLimit() {
             field("Bool", () -> faker.bool().bool())
         );
 
-        JsonTransformer<Object> transformer = new JsonTransformer<>();
+        JsonTransformerBuilder<Object> jsonTransformerBuilder = new JsonTransformerBuilder<>();
+        JsonTransformer<Object> transformer = jsonTransformerBuilder.build();
         String json = transformer.generate(schema, 2);
-
```
snuyanzin, line 35: do we need this change?

src/main/java/net/datafaker/transformations/JsonTransformer.java
mesketh: editorconfig says 4 indent for java - this class was 2.

src/main/java/net/datafaker/transformations/JsonTransformer.java
mesketh: ditto - editorconfig violation/drift - this realigns (pardon the pun).

src/main/java/net/datafaker/transformations/JsonTransformer.java
mesketh: editorconfig violation? formatting only. Not sure why this file seemed to have drifted from others - i did check the history and it's always had 2 indent. 🤷

snuyanzin: it's better not to mix checkstyle changes with PR related in same commit

src/main/java/net/datafaker/transformations/JsonTransformer.java
snuyanzin: it should be in a separate commit

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -7,149 +7,189 @@
...
+        map.put('\u0019', "\\u0019");
+        map.put('\u001A', "\\u001A");
+        map.put('\u001B', "\\u001B");
+        map.put('\u001C', "\\u001C");
+        map.put('\u001D', "\\u001D");
+        map.put('\u001E', "\\u001E");
+        map.put('\u001F', "\\u001F");
+        return Collections.unmodifiableMap(map);
+    }
+    private static char[] delimitedBy(boolean formatAsArray) {
```
snuyanzin, line 175: I tend to think that the name is confusing since it is not clear what delimiter is implied...
E.g. `,` is also a delimiter between array/object elements however it does not depend on this method result.

May be something like `isArray` or something related to this

src/main/java/net/datafaker/transformations/JsonTransformer.java
mesketh: I use intellij. It bundles and EditorConfig plugin and applies formatting changes automatically.

src/main/java/net/datafaker/transformations/JsonTransformer.java
mesketh: Well it's all based on the context of use so you don't get confused if you're using it correctly and follow the examples set. Delimiting something is just setting the boundaries so it does apply here IMO but, I want to get this change in so I'm not arguing! I could add a comment but, I'm detecting a precedent of not commenting privates.

PR #657, opened 2023-01-30
Garment Size Test Tweak

kingthorin:
Check the the value returned is one of the expected values, vs just not empty.

### DIFF
--- a/src/test/java/net/datafaker/providers/base/GarmentSizeTest.java
+++ b/src/test/java/net/datafaker/providers/base/GarmentSizeTest.java
@@ -1,6 +1,7 @@
 package net.datafaker.providers.base;
 
 
+import java.util.List;
 import org.junit.jupiter.api.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


src/test/java/net/datafaker/providers/base/GarmentSizeTest.java
```diff
@@ -1,14 +1,25 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class GarmentSizeTest extends BaseFakerTest<BaseFaker> {
 
     @Test
-    void size() {
-        assertThat(faker.garmentSize().size()).isNotEmpty();
+    void testSize() {
+        // Given
+        List<String> actualSizes = getActualSizes();
```
kingthorin, line 14: I had considered doing this as:
```java
List<String> actualSizes = List.of("XS", "S", "M", "L", "XL", "XXL", "XXXL");
```

Since it's likely always a small set of values, but I kept waffling back and forth. So let me know which form you like better.

kingthorin:
I'm sure there's lots of testing that could be more specific like this. This one just caught my eye for some reason. If this is merged I'm happy to tackle more.

src/test/java/net/datafaker/providers/base/GarmentSizeTest.java
kingthorin: Oh I just realized that List.of wouldn’t have been good for your Java 8 support. Still coulda declared it more inline’ishly.

bodiam:
Hi @kingthorin , there's quite a few tests which hardly test anything, so this is a great contribution, thank you!

PR #659, opened 2023-01-30
Enforce EOL=LF

kingthorin:
Set .gitattributes to enforce the EOL as specified in editor config.

I'm developing on Windows, prior to tweaking this I was getting spotless check issues.


bodiam:
I'm on Mac, so I never ran into this. Thanks again for fixing!

PR #660, opened 2023-01-30
Improve unit testing

kingthorin:
- AbstractBasicProviderTest > Added so that applicable tests could extend it in order to ensure that the values the providers return are actually within the expected list.
- various > Updated Unit Test classes to leverage the new abstract.
- Overwatch > moved to the videogame package.

This is in follow-up to: atafaker-net/datafaker#657.

### DIFF
--- a/src/test/java/net/datafaker/providers/base/AbstractBasicProviderTest.java
+++ b/src/test/java/net/datafaker/providers/base/AbstractBasicProviderTest.java
@@ -1,0 +1,54 @@
+package net.datafaker.providers.base;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Supplier;
+
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public abstract class AbstractBasicProviderTest<T extends BaseFaker> extends BaseFakerTest<BaseFaker> {
+    @SuppressWarnings("unchecked")
+    protected List<String> getBaseList(String key) {
+        return (List<String>) faker.fakeValuesService().fetchObject(key, faker.getContext());
+    }
+
+    @ParameterizedTest(name = "{0}")
+    @MethodSource("providerListTest")
+    protected void testProviderList(TestSpec testSpec) {
+        // Given
+        List<String> actual = getBaseList(testSpec.key);
+        // When
+        String item = (String) testSpec.supplier.get();
+        // Then
+        assertThat(item).as("Check item isn't empty").isNotEmpty();
+        assertThat(actual).as("Check actual list isn't empty and contains the item").isNotEmpty()
+                .anyMatch(item::equals);
+    }
+
+    protected abstract Collection<TestSpec> providerListTest();
+
+    protected static class TestSpec {
+        private final Supplier<?> supplier;
+        private final String key;
+
+        private TestSpec(Supplier<?> supplier, String key) {
+            this.supplier = supplier;
+            this.key = key;
+        }
+
+        public static TestSpec of(Supplier<?> supplier, String key) {
+            return new TestSpec(supplier, key);
+        }
+
+        @Override
+        public String toString() {
+            // The result of this toString will be used by IDE in test report
+            return "Key: " + key;
+        }
+    }
+}
--- a/src/test/java/net/datafaker/providers/base/ZodiacTest.java
+++ b/src/test/java/net/datafaker/providers/base/ZodiacTest.java
@@ -1,14 +1,12 @@
 package net.datafaker.providers.base;
 
+import java.util.Arrays;
+import java.util.Collection;
 
-import org.junit.jupiter.api.Test;
+class ZodiacTest extends AbstractBasicProviderTest<BaseFaker> {
 
-import static org.assertj.core.api.Assertions.assertThat;
-
-public class ZodiacTest extends BaseFakerTest<BaseFaker> {
-
-    @Test
-    void testSign() {
-        assertThat(faker.zodiac().sign()).isNotEmpty();
+    @Override
+    protected Collection<TestSpec> providerListTest() {
+        return Arrays.asList(TestSpec.of((() -> faker.zodiac().sign()), "zodiac.signs"));
     }
 }
--- a/src/test/java/net/datafaker/providers/base/ApplianceTest.java
+++ b/src/test/java/net/datafaker/providers/base/ApplianceTest.java
@@ -4,15 +4,18 @@ import org.junit.jupiter.api.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-class ApplianceTest extends BaseFakerTest<BaseFaker> {
+import java.util.Arrays;
+import java.util.Collection;
+
+class ApplianceTest extends AbstractBasicProviderTest<BaseFaker> {
 
     @Test
     void brand() {
         assertThat(faker.appliance().brand()).matches("[A-Za-z .-]+");
     }
 
-    @Test
-    void equipment() {
-        assertThat(faker.appliance().equipment()).isNotEmpty();
+    @Override
+    protected Collection<TestSpec> providerListTest() {
+        return Arrays.asList(TestSpec.of(() -> faker.appliance().equipment(), "appliance.equipment"));
     }
 }
--- a/src/test/java/net/datafaker/providers/base/AustraliaTest.java
+++ b/src/test/java/net/datafaker/providers/base/AustraliaTest.java
@@ -1,23 +1,15 @@
 package net.datafaker.providers.base;
 
-import org.junit.jupiter.api.Test;
+import java.util.Arrays;
+import java.util.Collection;
 
-import static org.assertj.core.api.Assertions.assertThat;
+class AustraliaTest extends AbstractBasicProviderTest<BaseFaker> {
 
-class AustraliaTest extends BaseFakerTest<BaseFaker> {
-
-    @Test
-    void locations() {
-        assertThat(faker.australia().locations()).isNotEmpty();
+    @Override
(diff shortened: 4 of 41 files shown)


kingthorin:
This is in follow-up to: https://github.com/datafaker-net/datafaker/pull/657. Please don't merge this yet, I'm just submitting to get initial feedback, then I'll extended to further providers/properties.

@bodiam are you happy with how this is looking? Is there anything you'd want done differently?

snuyanzin:
what do you think about parameterized test on `BaseFakerTest` level which could be annotated with `@MethodSource` pointing to abstract method providing key and supplier.
All other tests classes should just implement this method

kingthorin:
Can you provide params to a `@MethodSource` call? I thought the annotation was simply `@MethodSource("myParamGenerator")`

Or maybe I'm misunderstanding your vision.

kingthorin:
It seems `@MethodSource` methods are expected to be static, which an abstract method can't be. (Or I'm still not following along :wink:)

snuyanzin:
it could be non-static however in this case `BaseFakerTest` should be annotated as
```java
@TestInstance(TestInstance.Lifecycle.PER_CLASS)
public abstract class BaseFakerTest<T extends BaseFaker> {
...

   @ParameterizedTest(name = "{0}")
    @MethodSource("simpleTestArgs")
    protected void simpleListTest(TestSpec testSpec) {
        // Given
        List<String> actual = getBaseList(testSpec.key);
        // When
        String item = (String) testSpec.supplier.get();
        // Then
        assertThat(actual).anyMatch(item::equals);
    }

    protected abstract Collection<TestSpec> simpleTestArgs();

    protected static class TestSpec {
        private final Supplier<?> supplier;
        private final String key;
        private TestSpec(Supplier<?> supplier, String key) {
            this.supplier = supplier;
            this.key = key;
        }

        public static TestSpec of(Supplier<?> supplier, String key) {
            return new TestSpec(supplier, key);
        }

        @Override
        public String toString() {
            return "Key: " + key;
        }
    }
...
```
It's better to use pojo since there are some java compiliation issues for the case like `Arguments.of(() -> faker.appliance().equipment(), "appliance.equipment")`
.

On child tests it could look like that e.g. just one method instead of 3 in `AustraliaTest`
```
class AustraliaTest extends BaseFakerTest<BaseFaker> {

    @Override
    protected Collection<TestSpec> simpleTestArgs() {
        return Arrays.asList(TestSpec.of(() -> faker.australia().locations(), "australia.locations"),
            TestSpec.of(() -> faker.australia().animals(), "australia.animals"),
            TestSpec.of(() -> faker.australia().states(), "australia.states"));
    }
}

```

kingthorin:
Cool thanks!!

I’ll tweak the names too, I know all the simple* weren’t great. I was just trying to get a functioning model setup.

kingthorin:
I had to separate it into another class so that it could be extended by only those classes where it was applicable. But it's working nicely so I'm moving ahead. I'm about a third done the base provider list. Will get this updated later this evening.

kingthorin:
The LocalityTest failure seems to be related to having added `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` which I guess means that there's some state that was shared previously that is no longer.

kingthorin:
Relocated it to the new Abastract resolved the issue.

src/test/java/net/datafaker/providers/base/AbstractBasicProviderTest.java
```diff
@@ -0,0 +1,58 @@
...
+    protected List<String> getBaseList(String key) {
+        return (List<String>) faker.fakeValuesService().fetchObject(key, faker.getContext());
+    }
+
+    /**
+     * Ensures that the supplied method actually returns a value in the expected list.
+     * 
+     * @param method the method from which the test value will be obtained.
+     * @param key the YAML key from which the list of actual items should be retrieved.
+     */
```
snuyanzin, line 25: I suspect the javadoc comment is obsolete now

src/test/java/net/datafaker/providers/base/ZodiacTest.java
```diff
@@ -1,14 +1,12 @@
 package net.datafaker.providers.base;
 
+import java.util.Arrays;
+import java.util.Collection;
 
-import org.junit.jupiter.api.Test;
+public class ZodiacTest extends AbstractBasicProviderTest<BaseFaker> {
```
snuyanzin, line 6:
```suggestion
class ZodiacTest extends AbstractBasicProviderTest<BaseFaker> {
```
since it is junit5 it could be package private

snuyanzin:
Thanks, it looks great.
In general lgtm.
I added a couple of tiny comments

src/test/java/net/datafaker/providers/base/AbstractBasicProviderTest.java
kingthorin: I re-jigged this. Maybe it's unnecessary? Lemme know if you'd rather I just remove it. (I guess in reality there are now plenty of examples for people to refer to.)

src/test/java/net/datafaker/providers/base/AbstractBasicProviderTest.java
```diff
@@ -0,0 +1,64 @@
...
+    @ParameterizedTest(name = "{0}")
+    @MethodSource("providerListTest")
+    protected void testProviderList(TestSpec testSpec) {
+        // Given
+        List<String> actual = getBaseList(testSpec.key);
+        // When
+        String item = (String) testSpec.supplier.get();
+        // Then
+        assertThat(item).isNotEmpty();
+        assertThat(actual).anyMatch(item::equals);
```
kingthorin, line 29: SonarList suggests tweaking this similar to:
```java
        assertThat(item).as("Check item isn't empty").isNotEmpty();
        assertThat(actual).as("Check actual list contains the item").anyMatch(item::equals);
```

Do you think that helps?

https://github.com/joansmith/sonar-java/blob/master/java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S2698.html

snuyanzin: I'm not sure about having 
```java
        assertThat(item).isNotEmpty();
```
Will not the `anyMatch` condition anyway fail in case `actual` is empty?

snuyanzin: in fact assertj api is mostly self explanatory, so i don't think an extra description is required here...
i would say it's optional... so up to you

src/test/java/net/datafaker/providers/base/AbstractBasicProviderTest.java
snuyanzin: yes.. i think it could be simply dropped

src/test/java/net/datafaker/providers/base/AbstractBasicProviderTest.java
kingthorin: Yes. I did a quick test setting actual to Collections.emptyList() and it does fail in that case.

kingthorin: Well okay the output could be more specific.
```text
java.lang.AssertionError: [Check actual list contains the item] 
Expecting any elements of:
  []
to match given predicate but none did.
```

I could add another assertion that actual isn't empty.

snuyanzin: ok, makes sense

kingthorin: Okay, this is addressed now too.

src/test/java/net/datafaker/providers/base/AbstractBasicProviderTest.java
kingthorin: Done

src/test/java/net/datafaker/providers/base/AbstractBasicProviderTest.java
bodiam: I'm not a fan of using those `as` things, I only use them in very specific cases, I think the default error messages are pretty self-explanatory, but on the other hand, if you think it's nice to have them (I thing that issue, RSPEC-2698, is really not useful, and adds more maintenance burden than it's worth. But I'm not a fan of SonarQube/Linters/etc in general :-) )

bodiam:
This PR looks @kingthorin . If you think it's okay, I'm happy if you want to merge it yourself.

bodiam:
PS: I'm not sure if it's too much too ask, but there's a `ProviderGenerator` class, which creates Provider classes and tests to match them. Would it make sense to update this class too?

kingthorin:
I’ll have a look at that. If it’s quick I’ll add it here, otherwise I’ll tackle it in another PR.

src/test/java/net/datafaker/providers/base/AbstractBasicProviderTest.java
bodiam: No need to remove them, I'm fine with it either way, it just feels a bit like duplication to me, that's all.

bodiam:
> I’ll have a look at that. If it’s quick I’ll add it here, otherwise I’ll tackle it in another PR.

No need to tackle it, I can do it, and I'm not even sure we still need the class since we now have Providers for almost all yml files, except for the gaming ones, the are slightly more complex to auto generate.

kingthorin:
They’re one pending reviewer. Reading back through the log you self requested a review after you suggested merging? 😉 I’m going to guess stray click but want to be sure before I go ahead with it. (Thanks for the collab invite BTW)

snuyanzin (approved):

bodiam:
> @bodiam There's one pending review. Reading back through the log: you self requested a review after you suggested merging? 😉 I’m going to guess stray click but want to be sure before I go ahead with it. (Thanks for the collab invite BTW)

Ah, I was in two minds here, I wasn't sure if we required a review of not, and then I was reviewing, and got distracted :) So let's call it a stray click indeed :)

bodiam (approved):

kingthorin:
Yay 🎉

bodiam:
Woohoo! Nice! 🥇

PR #661, opened 2023-01-30
Correct custom provider example

AlexJFerreira:
It's only possible to create a custom provider with BaseProviders instead of a Faker in it's constructor.

With Faker, it gets the following error: "reason: no instance(s) of type variable(s) exist so that BaseProviders conforms to Faker"


snuyanzin:
thanks
i would suggest to fix the same issue with `InsectFromFile` in same doc

AlexJFerreira:
> thanks i would suggest to fix the same issue with `InsectFromFile` in same doc

Yeah, well noted. It's done.

bodiam:
Thanks! Merged!

PR #663, opened 2023-02-01
Remove usage Format Json from BaseFaker

RVRhub:


src/main/java/net/datafaker/service/FakeValuesService.java
```diff
@@ -458,39 +462,46 @@ public String csv(String delimiter, char quote, boolean withHeader, int limit, S
...
      * Generates json based on input.
      */
-    public Json json(String... fieldExpressions) {
+    public String json(String... fieldExpressions) {
         if (fieldExpressions.length % 2 != 0) {
             throw new IllegalArgumentException("Total number of field names and field values should be even");
         }
-        Json.JsonBuilder jsonBuilder = new Json.JsonBuilder();
+        JsonTransformer.JsonTransformerBuilder<Object> jsonTransformerBuilder = new JsonTransformer.JsonTransformerBuilder<>();
+        JsonTransformer<Object> transformer = jsonTransformerBuilder.build();
```
snuyanzin, line 470: I guess it could be moved to static final

src/main/java/net/datafaker/service/FakeValuesService.java
```diff
@@ -458,39 +462,46 @@ public String csv(String delimiter, char quote, boolean withHeader, int limit, S
...
      * Generates json based on input.
      */
-    public Json jsona(String... fieldExpressions) {
+    public String jsona(String... fieldExpressions) {
         if (fieldExpressions.length % 3 != 0) {
             throw new IllegalArgumentException("Total number of field names and field values should be dividable by 3");
         }
-        Json.JsonBuilder jsonBuilder = new Json.JsonBuilder();
+        JsonTransformer.JsonTransformerBuilder<Object> jsonTransformerBuilder = new JsonTransformer.JsonTransformerBuilder<>();
+        JsonTransformer<Object> transformer = jsonTransformerBuilder.build();
```
snuyanzin, line 490: I guess it could be moved to static final

snuyanzin:
Hi @RVRhub 
thanks for your contribution
i would suggest to remove deprecated class in this PR as well
WDYT?

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -51,8 +52,9 @@ public String generate(FakeSequence<IN> input, Schema<IN, ?> schema) {
         }
 
         StringJoiner data = new StringJoiner(LINE_SEPARATOR);
-        for (IN in : input) {
-            data.add(apply(in, schema));
+        Iterator<IN> iterator = input.iterator();
+        while (iterator.hasNext()){
+            data.add(apply(iterator.next(), schema) + (commaBetweenObjects & iterator.hasNext() ? "," : ""));
```
snuyanzin, line 57: why not `&&`?

src/test/java/net/datafaker/sequence/FakeStreamTest.java
```diff
@@ -247,18 +250,23 @@ void toCsvFromInfiniteSequenceWithLimit() {
...
+        JsonTransformer<Data> transformer = jsonTransformerBuilder.build();
+
+        String json = transformer.generate(stream, Schema.of(
+            field("name", Data::name),
+            field("value", Data::value),
+            field("range", Data::range),
+            field("unit", Data::unit)
+        ));
+
+        System.out.println(json);
```
snuyanzin, line 269: I think we could remove this output

src/main/java/net/datafaker/providers/base/BaseFaker.java
```diff
@@ -307,11 +306,11 @@ public String csv(String separator, char quote, boolean withHeader, int limit, S
         return fakeValuesService().csv(separator, quote, withHeader, limit, columnExpressions);
     }
 
-    public Json json(String... fieldExpressions) {
+    public String json(String... fieldExpressions) {
         return fakeValuesService().json(fieldExpressions);
     }
 
-    public Json jsona(String... fieldExpressions) {
+    public String jsona(String... fieldExpressions) {
```
kingthorin, line 313: Are these considered breaking changes? Since they’re public methods shouldn’t they be deprecated and alternatives added/encouraged?

src/main/java/net/datafaker/service/FakeValuesService.java
```diff
@@ -458,39 +465,42 @@ public String csv(String delimiter, char quote, boolean withHeader, int limit, S
...
             final int index = i;
-            jsonBuilder.set(fieldExpressions[index], () -> fieldExpressions[index + 1]);
+            fields.add(field(fieldExpressions[index], () -> fieldExpressions[index + 1]));
         }
-        return jsonBuilder.build();
+        Schema<Object, ?> schema = Schema.of(fields.toArray(new SimpleField[0]));
+        return JSON_TRANSFORMER.generate(schema, 1);
     }
 
+
```
kingthorin, line 482: Random blank insertion, when there’s already a blank line?

src/main/java/net/datafaker/providers/base/BaseFaker.java
kingthorin: I guess this question applies for FakerValueService too

bodiam: Normally I'd say yes, but I think the usage is so limited that the impact would be minimal. I would prefer depreciating a method, but it might be tricky in this case since only the return type changed. But I thought Json was already deprecated?

kingthorin: I didn’t check in an IDE. This is just me reading things and trying to get used to how the project does things 😉

snuyanzin: yes, `Json` and all other classes from `formats` package were deprecated with corresponding comments.
However it is a fair point that probably in future we should deprecate methods and probably some other usages as well ....

snuyanzin: that's not clear how to embed it in datafaker's pipeline...
Because there could be deprecations from previous releases which should be removed and there could be deprecation done only within current development cycle => should be removed not earlier than after one release

bodiam: The idea about how this project does things is to not have too many rules, respect everyone's contribution and the time they spent, and not nitpick on small things like a space vs a tab kinda issues. Nobody is getting paid, let's keep it light and simple. Happy to change my mind though!

kingthorin: @snuyanzin I'll see what I can get setup and how things look. Let's not have it hold up this PR.

@bodiam that's fair, I'll try to fit that model. Feel free to tell me if I'm pushing it on anything. (Like https://github.com/datafaker-net/datafaker/pull/663#discussion_r1093902880 ? 😉)

src/main/java/net/datafaker/service/FakeValuesService.java
bodiam: Haha, I saw this, don't mind this comment at all, but I wouldn't stop a PR for it either!

kingthorin: That’s fair, I agree. Nice to have reverted but not a blocker

src/main/java/net/datafaker/transformations/JsonTransformer.java
RVRhub: Fixed

src/test/java/net/datafaker/sequence/FakeStreamTest.java
RVRhub: fixed

src/main/java/net/datafaker/service/FakeValuesService.java
RVRhub: Accident, line deleted.

kingthorin (approved):
Seems reasonable to me

snuyanzin:
@RVRhub thanks for addressing comments
lgtm

src/main/java/net/datafaker/providers/base/BaseFaker.java
kingthorin: To get warning from xlint we'd do this:

```diff
warning: in the working copy of 'pom.xml', CRLF will be replaced by LF the next time Git touches it
diff --git a/pom.xml b/pom.xml
index 0c0b607b..d8e4e945 100644
--- a/pom.xml
+++ b/pom.xml
@@ -176,6 +176,9 @@
                     <configuration>
                         <source>${maven.compiler.source}</source>
                         <target>${maven.compiler.target}</target>
+                        <compilerArgs>
+                            <arg>-Xlint:all</arg>
+                        </compilerArgs>
                     </configuration>
                 </plugin>
                 <plugin>
```

To have those warnings error (and fail builds) we'd add another arg entry `<arg>-Werror</arg>`.

There are quite a large number of warnings. Shall I open an issue adding the output? So that we can discuss further. I'm not sure the best way to tackle it. I definitely don't know enough about the code base yet to tackle it all.

snuyanzin: That's the main issue.
Yes we can automate finding all of deprecations and fail in case there is at least one.
However we can not say like fail if there is only from a specific package or class...

kingthorin: From my review most of the identified issues are actually in datafakers code. For the few existing or future occurrences where it's a library we depend on suppressions can be used.

snuyanzin: The idea was to make it failing for deprecated code which was deprecated in previous releases. However it should not fail for the code which was deprecated only in current development cycle.
Only pick each warning/failure and compare against git logs manually

kingthorin: Okay if that's what you're after then yes it isn't something we can do.

We can limit to deprecation concerns `-Xlint:deprecation`, but the catch is that whenever we deprecate something we'd also have to stop using it internally. Which really should be true anyway IMHO but isn't necessarily quick or straight forward.

On that limited front here's what we'd have to address:
```console
[WARNING] /C:/datafaker/src/main/java/net/datafaker/service/FakeValuesService.java:[447,20] net.datafaker.formats.Csv in net.datafaker.formats has been deprecated
[WARNING] /C:/datafaker/src/main/java/net/datafaker/service/FakeValuesService.java:[447,43] net.datafaker.formats.Csv in net.datafaker.formats has been deprecated
[WARNING] /C:/datafaker/src/main/java/net/datafaker/service/FakeValuesService.java:[457,9] net.datafaker.formats.Csv in net.datafaker.formats has been deprecated
[WARNING] /C:/datafaker/src/main/java/net/datafaker/service/FakeValuesService.java:[457,36] net.datafaker.formats.Csv in net.datafaker.formats has been deprecated
[WARNING] /C:/datafaker/src/main/java/net/datafaker/service/FakeValuesService.java:[460,30] net.datafaker.formats.Csv in net.datafaker.formats has been deprecated
[WARNING] /C:/datafaker/src/main/java/net/datafaker/service/FakeValuesService.java:[462,16] net.datafaker.formats.Format in net.datafaker.formats has been deprecated
[WARNING] /C:/datafaker/src/main/java/net/datafaker/transformations/JavaObjectTransformer.java:[20,31] newInstance() in java.lang.Class has been deprecated

```

snuyanzin: yes, probably we can go that way, agree.

kingthorin: Okay. I’ll open an issue. I started to look at some of these. But my lack of familiarity with the code base quickly became a challenge :wink:

snuyanzin: From one side this would be great.
From another side just keep in mind that @RVRhub is working on removal of `Format.Json`

PR #664, opened 2023-02-04
Include XML files in .gitattributes as text

kingthorin:
So that EOL don't accidentally get switched to CRLF (or mixed).


kingthorin:
I'm not sure what's up with codecov, there should be zero change related to this PR 🤷

PR #654, opened 2023-01-28
Computer brands feature request

HattoriHenzo:
Update the Computer faker with brands

Issue #652, opened 2023-01-26
Computer brands feature request


HattoriHenzo:
**Is your feature request related to a problem? Please describe.**
Very nice project by the way. I would like to update to Computer provider with computer brands. And after provide some more Providers.

**Describe the solution you'd like**
I would like to provide a code to update the Computer provider with available brands on the market such as: such as Apple, HP, Asus etc...

**Additional context**
We hope my contribution will help.

bodiam:
Hi @HattoriHenzo , we would more than welcome such contribution! Feel free to make the changes, we can probably include this in the 1.8 release!

kingthorin:
@HattoriHenzo are you tackling this?  If not I might take a shot at it.

HattoriHenzo:
@kingthorin Hi, I already send a PR for this.

kingthorin:
Oh sorry, I didn't see it linked here. I'll go find it.

kingthorin:
I guess datafaker-net/datafaker#654 closes this :grinning:

For future reference :wink: https://docs.github.com/en/enterprise-server@3.4/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

PR #665, opened 2023-02-04
Add computer brands

kingthorin:
- Framework is gaining momentum (https://frame.work/)
- Gateway is a long time brand (how could we forget/miss that cow pattern logo 😉 https://gatewayusa.com/)
- Gigabyte not just motherboards and components, they do have laptops and desktops https://www.gigabyte.com/

Follow-up to: datafaker-net/datafaker#654 & datafaker-net/datafaker#652

### DIFF
--- a/src/main/resources/en/computer.yml
+++ b/src/main/resources/en/computer.yml
@@ -49,7 +49,10 @@ en:
         - Asus
         - Corsair
         - Dell
+        - Framework
         - Fujitsu
+        - Gateway
+        - Gigabyte
         - HP
         - Huawei
         - Lenovo


src/main/resources/en/computer.yml
```diff
@@ -58,6 +61,7 @@ en:
         - Packard Bell
         - Razer
         - Samsung
+        - Sony
```
kingthorin, line 64: Ummm "what-the-diff" has lost it's mind, this is definitely an addition.

Though arguably maybe it should be VAIO?
- https://www.sony.ca/en/electronics/computers/vaio-laptops
- https://en.wikipedia.org/wiki/Vaio
- https://us.vaio.com/

bodiam: What the diff isn't perfect, I'm just trialing it. It's 80% or so correct, and it's more aan FYI than anything.

Having said that: I was surprised too, but then again, I don't think they still sell computers: https://www.sony.com.au/electronics/computers/vaio-laptops

kingthorin: Okay. It looks like Sony spun Vaio off and it’s now an independent company.

I’ll just skip it for now. If someone feels stronger about it in the future hopefully they’ve dig into it more 😉

bodiam: So, you're saying the AI was right all along! Scary! :-)

kingthorin:
Tweaked

bodiam:
Also LGTM! Thanks!

Issue #666, opened 2023-02-05
Address deprecation


kingthorin:
In order move toward implementing `Xlint` and ensuring we don't have any compiler warnings. A first somewhat easy step is to address deprecation.

We can limit to deprecation concerns `-Xlint:deprecation`, but the catch is that whenever we deprecate something we'd also have to stop using it internally. Which really should be true anyway IMHO but isn't necessarily quick or straight forward.

On that limited front here's what we'd have to address:
```console
[WARNING] /C:/datafaker/src/main/java/net/datafaker/service/FakeValuesService.java:[447,20] net.datafaker.formats.Csv in net.datafaker.formats has been deprecated
[WARNING] /C:/datafaker/src/main/java/net/datafaker/service/FakeValuesService.java:[447,43] net.datafaker.formats.Csv in net.datafaker.formats has been deprecated
[WARNING] /C:/datafaker/src/main/java/net/datafaker/service/FakeValuesService.java:[457,9] net.datafaker.formats.Csv in net.datafaker.formats has been deprecated
[WARNING] /C:/datafaker/src/main/java/net/datafaker/service/FakeValuesService.java:[457,36] net.datafaker.formats.Csv in net.datafaker.formats has been deprecated
[WARNING] /C:/datafaker/src/main/java/net/datafaker/service/FakeValuesService.java:[460,30] net.datafaker.formats.Csv in net.datafaker.formats has been deprecated
[WARNING] /C:/datafaker/src/main/java/net/datafaker/service/FakeValuesService.java:[462,16] net.datafaker.formats.Format in net.datafaker.formats has been deprecated
[WARNING] /C:/datafaker/src/main/java/net/datafaker/transformations/JavaObjectTransformer.java:[20,31] newInstance() in java.lang.Class has been deprecated

```

_Originally posted by @kingthorin in https://github.com/datafaker-net/datafaker/pull/663#discussion_r1096534504_

---

To get warnings from xlint we'd do this:

```diff
warning: in the working copy of 'pom.xml', CRLF will be replaced by LF the next time Git touches it
diff --git a/pom.xml b/pom.xml
index 0c0b607b..d8e4e945 100644
--- a/pom.xml
+++ b/pom.xml
@@ -176,6 +176,9 @@
                     <configuration>
                         <source>${maven.compiler.source}</source>
                         <target>${maven.compiler.target}</target>
+                        <compilerArgs>
+                            <arg>-Xlint:all</arg>
+                        </compilerArgs>
[...]

PR #667, opened 2023-02-05
implement currently passing Xlint checks

kingthorin:
Related to: datafaker-net/datafaker#666

It seems the only outstanding items in the project are `-Xlint:deprecation` and `-Xlint:unchecked`.

Ultimately we'll be able to replace the individual options with `Xlint:all`. But while we're building toward that we may as well prevent these from creeping in :wink:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -176,6 +176,25 @@
                     <configuration>
                         <source>${maven.compiler.source}</source>
                         <target>${maven.compiler.target}</target>
+                        <compilerArgs>
+                            <arg>-Xlint:cast</arg>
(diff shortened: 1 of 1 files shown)


bodiam:
Lgtm!

bodiam:
I said: LGTM, but I am having second thoughts now. I can't run the code anymore due to a `java: warnings found and -Werror specified` error/warning. Also running any individual test is currently failing. I'm going to revert this change.

kingthorin:
Could you let me know how you’re running individual test? I’ll dig into this more.

PR #676, opened 2023-02-09
Add Chess faker

panilya:


bodiam (approved):

src/test/java/net/datafaker/providers/sport/ChessTest.java
```diff
@@ -0,0 +1,29 @@
+package net.datafaker.providers.sport;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ChessTest extends SportFakerTest {
```
kingthorin, line 7: These should follow the newer testing mechanism. Similar to: https://github.com/datafaker-net/datafaker/blob/c0ef9b95f691c8e514cc5f238ca336b894d79489/src/test/java/net/datafaker/providers/base/AviationTest.java#L32-L38

kingthorin (changes requested):

src/test/java/net/datafaker/providers/sport/ChessTest.java
kingthorin: Okay, sorry I missed that.

panilya: It's okay

kingthorin (approved):

PR #677, opened 2023-02-09
Apply comparison against yaml file content for tests within entertainment package

snuyanzin:
The PR applies comparison against yaml file content for tests in entertainment package.
P.S. Probably later will do for others

### DIFF
--- a/src/test/java/net/datafaker/providers/entertainment/GratefulDeadTest.java
+++ b/src/test/java/net/datafaker/providers/entertainment/GratefulDeadTest.java
@@ -1,19 +1,19 @@
 package net.datafaker.providers.entertainment;
 
-import org.junit.jupiter.api.Test;
+import net.datafaker.providers.base.GratefulDead;
 
-import static org.assertj.core.api.Assertions.assertThat;
+import java.util.Arrays;
+import java.util.Collection;
 
 class GratefulDeadTest extends EntertainmentFakerTest {
 
-    @Test
-    void players() {
-        assertThat(faker.gratefulDead().players()).isNotEmpty();
-    }
+    private final GratefulDead gratefulDead = getFaker().gratefulDead();
 
-    @Test
-    void songs() {
-        assertThat(faker.gratefulDead().songs()).isNotEmpty();
+    @Override
+    protected Collection<TestSpec> providerListTest() {
+        return Arrays.asList(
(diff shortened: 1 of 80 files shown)


src/test/java/net/datafaker/providers/entertainment/StarWarsTest.java
```diff
@@ -2,48 +2,35 @@
...
-
-    @Test
-    void droid() {
-        assertThat(faker.starWars().droids()).isNotEmpty();
-    }
-
-    @Test
-    void wookieWords() {
-        assertThat(faker.starWars().wookieWords()).isNotEmpty();
+    void alternativeSpelling() {
```
kingthorin, line 19: This can’t be done the new way? (I’m on my
Phone and couldn’t check the code or data file

kingthorin:
Thanks for tackling this. FYI any tests in the base package that used regex I just skipped over. I’ll have a better look at them and decide if others should be moved to the new infrastructure.

src/test/java/net/datafaker/providers/entertainment/StarWarsTest.java
snuyanzin: There are some methods in different classes including this one which are a bit trickier under the hood...
e.g. this method looks like 
```java
    public String alternateCharacterSpelling() {
        return resolve("star_wars.alternate_character_spellings." + CHARACTERS[faker.random().nextInt(CHARACTERS.length)]);
    }
```
so it requires some more sophisticated handling... 
I would rather migrate everything that's possible to new way and then try to handle the rest

src/test/java/net/datafaker/providers/entertainment/GratefulDeadTest.java
```diff
@@ -1,19 +1,19 @@
 package net.datafaker.providers.entertainment;
 
-import org.junit.jupiter.api.Test;
+import net.datafaker.providers.base.GratefulDead;
```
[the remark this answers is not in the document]
snuyanzin: good catch
done, thanks

snuyanzin:
>FYI any tests in the https://github.com/datafaker-net/datafaker/pull/660 package that used regex I just skipped over.

yes, some  of them are tricky...
either in java implementation or also in yaml config there could be macroses...

PR #680, opened 2023-02-10
Fix Dermatology typo

kingthorin:
- Deprecate misspelt method.
- Update tests.
- Fix data file.


kingthorin:
How are release/change notes generated for this project? Should I be doing something specific to ensure this is included?

snuyanzin:
based on git history, usually done by github
+ there is `net.datafaker.script.ProvidersDocsGenerator` which generates provider docs with version from `@since` javadoc annotation

PR #682, opened 2023-02-10
More unit test improvements

kingthorin:
- BaseFakerTest.java > Allow callers to optionally specify a regex to match against.
- UnitTest classes > Use newer unit test infrastructure where practical.


kingthorin:
I plan to continue through this, just wanted to submit it in case: 1) Anyone hated how I'd changed things. 2) You'd know it was being worked on 😉

src/test/java/net/datafaker/providers/base/BaseFakerTest.java
```diff
@@ -54,7 +56,7 @@ protected List<String> getBaseList(String key) {
 
     @ParameterizedTest(name = "{0}")
     @MethodSource("providerListTest")
-    protected void testProviderList(TestSpec testSpec, TestInfo testInfo) {
```
snuyanzin, line 57: not sure about that change...
when it is protected it allows to use this in other projects to test custom fakers...

kingthorin: Okay, will restore

src/test/java/net/datafaker/providers/base/AbstractBasicProviderTest.java
```diff
@@ -1,7 +1,5 @@
 package net.datafaker.providers.base;
 
-
-
 import org.junit.jupiter.api.TestInstance;
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
snuyanzin, line 5: I would suggest to remove this class completely

I moved logic to `BaseFakerTest` so it should work

kingthorin: 👍

kingthorin:
No longer WIP, ready for review.

src/test/java/net/datafaker/providers/base/BaseFakerTest.java
```diff
@@ -72,22 +74,29 @@ protected void testProviderList(TestSpec testSpec, TestInfo testInfo) {
...
+            this.regex = regex;
         }
 
         public static TestSpec of(Supplier<?> supplier, String key) {
-            return new TestSpec(supplier, key);
+            return new TestSpec(supplier, key, "");
+        }
+
+        public static TestSpec of(Supplier<?> supplier, String key, String regex) {
+            return new TestSpec(supplier, key, "");
```
snuyanzin, line 99: shouldn't it be like that?
```suggestion
            return new TestSpec(supplier, key, regex);
```
or even more i would suggest `Pattern.compile` to compile it during test input init

snuyanzin:
I didn't get where regex is checked in new way?

src/test/java/net/datafaker/providers/base/BaseFakerTest.java
kingthorin: Yup I think that was a copy/paste mistake 🤦‍♂️

kingthorin:
> I didn't get where regex is checked in new way?

Ugh, thanks. I lost it along the way in some step of the implementation/re-implementation. Fixed now.

kingthorin:
@snuyanzin I think I've addressed both your concerns. I put the Pattern.compile inside the regex assertion, let me know if you think it should be in the `of` method. Checking regex string isEmpty seemed easiest. But, I guess I could try/catch on PatternSyntaxException or check empty pattern `^$` and set it null, then null check it prior to the assertion if that's preferable.

kingthorin:
Fixed the build/test failure.

snuyanzin (approved):

PR #690, opened 2023-02-11
Migrated Java 8 code to Java 17 syntax improvements.

bodiam:


src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -159,7 +159,7 @@ private String providersPerVersionTable() {
     }
 
     /**
-     * Searching through all providers {@link subTypes}, getting since tag, and building
+     * Searching through all providers subTypes, getting since tag, and building
```
kingthorin, line 162: Was this intentional?

kingthorin:
Did you do this with some sort of automation, or is this just Search/Replace and memory?

src/main/java/net/datafaker/providers/base/Locality.java
```diff
@@ -44,11 +43,11 @@ public Locality(BaseProviders baseProviders) {
      * @return a List of Strings with the name of the locale (eg. "es", "es-MX")
      */
     public List<String> allSupportedLocales() {
-        return allSupportedLocales(new HashSet<>(Collections.singletonList("datafaker")));
+        return allSupportedLocales(new HashSet<>(List.of("datafaker")));
```
snuyanzin, line 46:
```suggestion
        return allSupportedLocales(Set.of("datafaker"));
```

src/main/java/net/datafaker/providers/base/Locality.java
```diff
@@ -44,11 +43,11 @@ public Locality(BaseProviders baseProviders) {
...
     }
 
     public List<String> allSupportedLocales(Set<String> fileMasks) {
-        Set<String> langs = new HashSet<>(Arrays.asList(Locale.getISOLanguages()));
+        Set<String> langs = new HashSet<>(List.of(Locale.getISOLanguages()));
```
snuyanzin, line 50:
```suggestion
        Set<String> langs = Set.of(Locale.getISOLanguages());
```

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -41,7 +41,7 @@ public class ProvidersDocsGenerator {
     private final Set<Class<?>> subTypes = new TreeSet<>(providersComparatorBySimpleName);
 
     // Exclude non-providers from generation
-    private final Set<String> providersToExcludeFromGeneration = new HashSet<>(Arrays.asList("CustomFakerTest", "InsectFromFile", "Insect"));
+    private final Set<String> providersToExcludeFromGeneration = new HashSet<>(List.of("CustomFakerTest", "InsectFromFile", "Insect"));
```
snuyanzin, line 44:
```suggestion
    private final Set<String> providersToExcludeFromGeneration = Set.of("CustomFakerTest", "InsectFromFile", "Insect");
```

kingthorin:
This is conflicting now that the unit test PR was merged. There’s probably a bunch more places that could use `List.of` now.

timtebeek:
Big pull requests like these are likely to give conflicts when there's other changes at the same time. Might I suggest that instead of one big PR, this is broken up into smaller PRs, with automation to handle individual changes? Here's a couple of such automations that can be applied quickly and easily, and when they're isolated don't require much review:

1. [Prefer List.of, Map.of, Set.of](https://docs.openrewrite.org/reference/recipes/java/migrate/util/javautilapis)
2. [Prefer "%s %s".formatted(a, b)](https://docs.openrewrite.org/reference/recipes/java/migrate/lang/stringformatted)
3. [Instance of pattern matching](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/cleanup/InstanceOfPatternMatch.java)
4. [Other modernization recipes](https://docs.openrewrite.org/reference/recipes/java/migrate)

Typically these can be applied as one off Maven commands already, although some require a snapshot version of OpenRewrite. Alternatively these can be applied [through our platform](https://public.moderne.io/) as well.

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
bodiam: Yes, IntelliJ gives an error on it

bodiam:
> This is conflicting now that the unit test PR was merged. There’s probably a bunch more places that could use `List.of` now.

That's unfortunate, but I'll have a go at it, otherwise I'll redo this PR, it's mostly automated anyway.

bodiam:
> Did you do this with some sort of automation, or is this just Search/Replace and memory?

This was done using OpenRewrite for some of the automation. Unfortunately, it missed quite a few enhancements such as improved switch statements, but IntelliJ also has Java 11, 12, 13, 14 and 16 inspections, so I ran those too. And after that, I fixed all the failing tests, since `List.of` creates immutable lists, and they are hard to sort in-place :-)

bodiam:
@timtebeek Would it be an option to also add these migrations to OpenRewrite:

image

Especially the `enhanced switch` results in quite a bit of code cleanup, and maybe the text block one is reasonably easy t implement? Just an idea.

timtebeek:
@bodiam Good to hear you already used OpenRewrite for part of this migration. :) We can (and should) cover more of the cases you've mentioned. Some, like text blocks, are already [on our project board](https://github.com/orgs/openrewrite/projects/4/views/10?filterQuery=text%20block). Others like the enhanced switch are indeed good suggestions to put on the board. Feel free to open issues for those!

I see you've already merged #696, so glad to see most of that work is done. I'll try to see if there's any more changes to be made with the existing recipes. I'd have expected [more than one case of instanceof](https://github.com/datafaker-net/datafaker/pull/696/files#diff-5c9f8f3f19760ff01f1b55eaf77ba580015927085fa7082997f40aec5028297fR99) as an example. It'll take a couple hours for public.moderne.io to pick up the most recent changes, but we can explore after that.

bodiam:
> I see you've already merged #696, so glad to see most of that work is done. I'll try to see if there's any more changes to be made with the existing recipes. I'd have expected [more than one case of instanceof](https://github.com/datafaker-net/datafaker/pull/696/files#diff-5c9f8f3f19760ff01f1b55eaf77ba580015927085fa7082997f40aec5028297fR99)

The instanceof improvement was picked up by IntelliJ, not by OpenRewrite. We do have several other places with `instanceof`, but none of them can be autocast like this, it seems.

PR #695, opened 2023-02-12
Fix maven central version badge

kingthorin:
The heroku app wasn’t updating. (Likely due to their free tier changes back in Nov). Use shields.io instead.


bodiam:
I think it was a cache issue, since it's updated now.

PR #693, opened 2023-02-11
Support list of objects for JSON

RVRhub:
PR adds object list support for JSON.
An example of usage

 ```
JsonTransformer.<Name>builder().build().generate(
            Schema.of(
                field("text", () -> "Mrs. Brian Braun"),
                field("objectCollection", () -> List.of(
                        compositeField(null, new Field[]{
                                field("country", () -> "Denmark"),
                                field("city", () -> "Port Angel")
                            }
                        )
                    )
                )
            ), 1);
```

Result:
```
{"text": "Mrs. Brian Braun", "objectCollection": [{"country": "Denmark", "city": "Port Angel"}]}
```

PR #698, opened 2023-02-12
Removal of deprecated code and renamed the old formats to transformers.

bodiam:


bodiam:
@snuyanzin I've commented out some code related to testing nested JSON, since it was using the old `Format` class. What would be the best way to proceed here?

snuyanzin:
currently there are some cases supported by deprecated code and not supported by transformers...
So first i would support them and then it could be removed.

e.g. for Json @RVRhub submitted #693 
so then need also support for xml

bodiam:
> currently there are some cases supported by deprecated code and not supported by transformers... So first i would support them and then it could be removed.
> 
> e.g. for Json @RVRhub submitted #693 so then need also support for xml
> 
> now with #699 json could be removed

oki, in that case, I'll redo this PR, but won't touch the transformation classes.

PR #703, opened 2023-02-12
Added Qodana support (experiment)

bodiam:
Qodana is a code quality platform similar to Snyk. This is an experiment to try it out. See for more info https://www.youtube.com/watch?v=WrhnUnzMUCg&t=203s or https://www.jetbrains.com/qodana/


kingthorin:
Looks like docker had to be installed in the runners to use this.

bodiam:
Yes, seems like it. I'm probably using the wrong build file, need to look into this. If you have an easy fix, please do share :)

PR #707, opened 2023-02-12
Fix non-capturing groups

kingthorin:
I'd previously used the wrong syntax, luckily it didn't break things.

I'm sure there are a ton of places this should be changed further. I'll look at that for another PR.


kingthorin:
Did "we" give up on what-the-diff? Or has it stopped working?

bodiam:
> Did "we" give up on what-the-diff? Or has it stopped working?

No available credits at the moment :(

bodiam:
I just applied for an Open Source plan! I hope we could get accepted, I do think these PR summaries are quite helpful!

kingthorin:
Okay

PR #708, opened 2023-02-12
refactor: Secure the use of SnakeYAML's constructor

timtebeek:
Based on [Secure the use of SnakeYAML's constructor](https://docs.openrewrite.org/reference/recipes/java/security/marshalling/securesnakeyamlconstructor).

### DIFF
--- a/src/main/java/net/datafaker/service/FakeValues.java
+++ b/src/main/java/net/datafaker/service/FakeValues.java
@@ -2,6 +2,7 @@ package net.datafaker.service;
 
 import net.datafaker.internal.helper.SingletonLocale;
 import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.SafeConstructor;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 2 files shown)


snuyanzin:
well, it should keep working, otherwise does not make sense

bodiam:
What problem does this solve, when you manage to get it to execute ?

bodiam:
> Note: if you want to limit objects to standard Java objects like List or Long you need [to use SafeConstructor](http://bitbucket.org/snakeyaml/snakeyaml/src/tip/src/test/java/examples/SafeConstructorExampleTest.java).

Really not sure why we want to do this.

timtebeek:
Yes apologies; was a little too quick to open a pull request before running & exploring locally. Had expected this to "just work".
The issue here is that calling load on SnakeYaml is not safe by default; with malicious input you can potentially expose yourself to security risks, as any class can be instantiated with inputs. A way to protect yourself from that is through the `SafeConstructor`, but seeing how we load a `Map` here, I'm afraid some more changes are in order, such as using defined types only.

Feel free to close the PR until I find a better way to work around this; did not mean to add to the noise here.

bodiam:
@timtebeek please don't apply random OpenRewrite PRs to our repo. In the future, please create a ticket first to discuss these things. Thank you.

timtebeek:
Noted, yes, apologies once again & will do.

bodiam:
> The issue here is that calling load on SnakeYaml is not safe by default; with malicious input you can potentially expose yourself to security risks, as any class can be instantiated with inputs.

This is not applicable to Datafaker at all. We don't allow the loading of malicious input, all YAML code is bundled, and only when creating custom fakers we are potentially not in control of the YAML loaded, but that's beyond the responsibility of Datafaker.

timtebeek:
Understandable that you'd consider this out of scope for the bundled resources. I'd have thought to be able to also close this for custom providers by adding `TypeDescription`s for the known types, but I won't explore that further as the feedback is clear.

PR #717, opened 2023-02-21
Add first time contributor guide

jamievasales:
- Add a first time contributor guide and add it to the README.MD and the mkdocs

I myself am a first time contributor (both to the project and open source). I have made some assumptions so that _something_ can be written but please do correct me on any assumptions that may be false. This is all quite meta 😄


bodiam:
I like it! Not sure if we need two documents, but we can deal with the later! Thanks!

PR #719, opened 2023-02-25
Do not cache methods of expressions with args

### LINKED ISSUE #716: `java.lang.IllegalArgumentException` is thrown on expressions with optional args
**To Reproduce**
Run the following code:
```java
package test;

import net.datafaker.Faker;

public class Test
{
    {
        Faker faker = new Faker();

        // Warm up start
        String flight1 = faker.expression("#{Aviation.flight}");
        System.out.println(flight1);
        // Warm up end

        String flight2 = faker.expression("#{Aviation.flight 'ICAO'}");
        System.out.println(flight2);
    }
}
```

**Actual behavior**
The error is thrown:
```
Exception in thread "main" java.lang.RuntimeException: java.lang.IllegalArgumentException: wrong number of arguments [ICAO]
	at net.datafaker.service.FakeValuesService.resolveExpression(FakeValuesService.java:668)
	at net.datafaker.service.FakeValuesService.resolveExpression(FakeValuesService.java:548)
	at net.datafaker.service.FakeValuesService.expression(FakeValuesService.java:426)
	at net.datafaker.providers.base.BaseFaker.expression(BaseFaker.java:442)
	at test.Test.main(Test.java:16)
```

**Expected behavior**
No errors

**Versions:**
This is a regression bug introduced in datafaker `1.8.0`, it's not reproduced in `1.7.0` and below.

**Additional context**
The root cause is this change: https://github.com/datafaker-net/datafaker/commit/01bc1fc5c66814d790e63aa5d895eb9d0ae50010#diff-c58f806813b4de1ea7cc4eb107ffb1eda306d72b65dee5a62871ef1c2b841912R35.
Methods are cached by their name only, but there could be overloaded methods (same name, but different args), so key for caching should be composite: method name + array of input argument types (or drop caching of methods, not sure how much benefit will so complicated caching bring).

valfirst:
Fixes #716

### DIFF
--- a/src/main/java/net/datafaker/service/FakeValuesService.java
+++ b/src/main/java/net/datafaker/service/FakeValuesService.java
@@ -639,33 +639,35 @@ public class FakeValuesService {
         final int dotIndex = getDotIndex(directive);
 
         Object resolved;
-        // resolve method references on CURRENT object like #{number_between '1','10'} on Number or
-        // #{ssn_valid} on IdNumber
-        if (dotIndex == -1) {
-            if (current instanceof AbstractProvider) {
-                final Method method = BaseFaker.getMethod((AbstractProvider<?>) current, directive);
+        if (args.length == 0) {
+            // resolve method references on CURRENT object like #{number_between '1','10'} on Number or
+            // #{ssn_valid} on IdNumber
+            if (dotIndex == -1) {
+                if (current instanceof AbstractProvider) {
+                    final Method method = BaseFaker.getMethod((AbstractProvider<?>) current, directive);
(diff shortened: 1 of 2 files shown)


bodiam:
Thanks! Do you think this is worth an 1.8.1 release?

valfirst:
@bodiam yes, it makes sense from my point of view

bodiam:
Yeah. I'll give it a shot this weekend!

bodiam:
1.8.1 with this fix should be available now.

valfirst:
@bodiam thank you!

PR #724, opened 2023-02-28
Add allergens in Food provider

panilya:


src/test/java/net/datafaker/providers/foods/FoodTest.java
```diff
@@ -21,6 +21,7 @@ void measurement() {
     protected Collection<TestSpec> providerListTest() {
         return List.of(
             TestSpec.of(food::dish, "food.dish"),
+            TestSpec.of(food::allergen, "food.allergens"),
```
snuyanzin, line 24: I would prefer abc order

snuyanzin:
in general lgtm 
there is a nit comment

src/test/java/net/datafaker/providers/foods/FoodTest.java
panilya: Fixed

kingthorin (approved):
Thanks

Issue #711, opened 2023-02-14
Kotlin documentation for transformers not working


bodiam:
The Documentation for the kotlin json examples, and potentially others, doesn't compile.

panilya:
Fixed in #725

PR #725, opened 2023-02-28
Fix Kotlin examples in Transformers documentation

panilya:
Issue: #711

### DIFF
--- a/docs/documentation/schemas.md
+++ b/docs/documentation/schemas.md
@@ -33,6 +33,8 @@ Example of schema definition:
 === "Kotlin"
 
     ``` kotlin
+        val faker = BaseFaker()
+
         val schema = Schema.of(
             field("first_name",
                 Supplier { faker.name().firstName() }),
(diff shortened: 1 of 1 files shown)


docs/documentation/schemas.md
```diff
@@ -72,7 +74,7 @@ CSV transformer could be build with help of `CsvTransformer.CsvTransformerBuilde
 === "Kotlin"
 
     ``` kotlin
-        val transformer = CsvTransformer.<String>builder().header(true).separator(separator).build()
+        val transformer = CsvTransformer.builder<String>().header(true).separator(separator).build()
```
kingthorin, line 77: I'm not a kotlin person, but in java on the right of an assignment you can usually just use a diamond operator `<>` without specifying the type because it's inferred by the JVM. (Though in that case it is needed on the left.)

panilya: I thought the same, but it doesn't compile. But my change fixes the problem.

You can try by yourself:
``` kotlin
    val faker = BaseFaker()

    val schema = Schema.of(
        field("first_name",
            Supplier { faker.name().firstName() }),
        field("last_name",
            Supplier { faker.name().lastName() }),
        field<String, String>("address",
            Supplier { faker.address().streetAddress() })
    )

    val transformer = CsvTransformer.builder<String>().header(true).separator(" : ").build()

    val csv = transformer.generate(schema, 2)
    print(csv)
```

kingthorin: Okay, thanks

kingthorin (approved):

docs/documentation/schemas.md
panilya: you're welcome!

PR #726, opened 2023-03-08
Improve aviation provider

panilya:


src/main/resources/en/aviation.yml
```diff
@@ -1,7 +1,7 @@
...
+      aircraft: [Airbus A220-100, Airbus A220-300, Airbus A318, Airbus A319, Airbus A320, Airbus A320neo, Airbus A321, Airbus A321neo, Airbus A330-200, Airbus A330-200F, Airbus A330-300, Airbus A330-800neo, Airbus A330-900neo, Airbus A340, Airbus A340-300, Airbus A340-500, Airbus A340-600, Airbus A350, Airbus A350-900, Airbus A350-1000, Airbus A380, Airbus A380-800, Boeing 707, Boeing 717, Boeing 727, Boeing 737, Boeing 747, Boeing 757, Boeing 767, Boeing 777, Boeing 787, Boeing 747-8, Boeing 737 MAX, Antonov An-12, Antonov An-24, Antonov An-26, Antonov An-28, Antonov An-30, Antonov An-32, Antonov An-72, Antonov An-124 Ruslan, Antonov An-140, Antonov An-148, Antonov An-158, Antonov An-225 Mriya, McDonnell Douglas MD11, McDonnell Douglas MD80, McDonnell Douglas MD81, McDonnell Douglas MD82, McDonnell Douglas MD83, McDonnell Douglas MD87, McDonnell Douglas MD88, McDonnell Douglas MD90]
```
kingthorin, line 4: Is there a reason to reduce this instead of just adding new ones?

panilya: 1) there are helicopters and airplanes, maybe it makes sense to have helicopters and planes lists
2) it contains mainly warplanes from WWII, which is, for me, less appropriate for most cases than civil planes used by airlines

kingthorin: For my two cents maybe create a new category/provider for ww2 (or military) planes and helicopters?

If it makes sense (I'm no aeronautics expert) you could separate helicopters in to military and civilian or commercial as well.

panilya: But I'm not sure what to do with the `aircraft()` method in Aviation, we need to preserve backward compatibility and at the same we're removing support of aircrafts and introduce airplanes and helicopters instead

kingthorin: I'd have to dig around, I've only been working with the project for a month or two. I wonder if there's a way to resolve from within a hierarchy?

```yaml
aircraft:
   airplanes: plane1, plane2
   helicopters: helo1, helo2
```

So `aircraft()` would resolve a random value from either list.

kingthorin (approved):
Thanks! Nice work

Issue #714, opened 2023-02-16
Unable to create custom provider using yaml file when bundled as a jar.


jwanggg:
Following the instructions [here](https://www.datafaker.net/documentation/custom-providers/#custom-provider-using-yaml-file), I am able to create a custom provider that reads the yaml file from the file path. 

However when I bundle my app together and try to run it the same it can't find the path/uri correctly to create the custom faker.

I am using 
```
faker.addPath(Paths.get(this.getClass().getResource("testFaker.yml").toURI))
```
to get the yml file locally and create the custom provider.

When bundled as a jar the above fails due to path being invalid from a jar. The path ends up being something like ```jar:file:/mnt/kubernetes/sandbox/dist/lib/main.jar!/testFaker.yml```

I am forced to do something like 

```
            final InputStream is = getClass().getResourceAsStream(testFaker.json);
            return mapper.readValue(is, new TypeReference<Map<String, List<String>>>() {});
```

and making somewhat of a hardcoded custom provider.

Is it possible to add a faker path through the resource stream or any way to add support for it?

kingthorin:
Is testfaker.yml really expected to be in the root of your jar?

jwanggg:
> Is testfaker.yml really expected to be in the root of your jar?

I've put the testfaker.yml under the resources folder. The whole path would be /src/main/resources

bodiam:
Try to load the file with a / in front of it, or move the file to the same directory as the source file is in.

bodiam:
I think if you have the file under the root folder, you need to add an extra `/`, like this: 
```
faker.addPath(Paths.get(this.getClass().getResource("/testFaker.yml").toURI))
```

Otherwise, you might have to put the YML file in the same location as where the source file is.

snuyanzin:
@jwanggg in case the issue still exists may I ask you to share the code to reproduce?
I can potentially have a look during the weekend

jwanggg:
@bodiam the above works when I am running my code locally. The reason I had to use ```ResourceAsStream`` is because  in deployed environments, it loads out of a jar, not off the filesystem directly. When the resource is loaded out of a Jar, it isn’t a “normal” URL so the Path.get fails in the above example

bodiam:
@jwanggg ahh, sorry, I see, it's doing the addPaths, that's not ideal indeed. We should maybe add an "addSource(InputStream)" method or so, would that be better?

jwanggg:
[...]

PR #727, opened 2023-03-10
Added 'addUrl' method

ghost:
fixes https://github.com/datafaker-net/datafaker/issues/714

### DIFF
--- a/src/main/java/net/datafaker/service/FakeValues.java
+++ b/src/main/java/net/datafaker/service/FakeValues.java
@@ -6,8 +6,6 @@ import org.yaml.snakeyaml.Yaml;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
-import java.nio.file.Files;
-import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


snuyanzin (changes requested):
@tcttk thanks for the contribution
1 .Can you please explain how this change relates to custom providers?

2. And as I mentioned in #714 before fixing it please provide a reproducible test case first

bodiam:
Thanks for this PR. Interesting to use a URL for this, how does this work? I expected something like an InputStream or so.

snuyanzin:
@tcttk thanks for the explanation
well I put a code snippet working for the case you've mentioned
> 
```
my-app
|
|----> depends my-datafaker-library (addPath(locale, Paths.get("src/main/resources/de/sample.yml")) does not work ❌)
                           |
                           |----> /resources/sample.yml  
                           |----> depends datafaker-library
```
https://github.com/datafaker-net/datafaker/issues/714#issuecomment-1451568692

it should work.


Regarding changes I'm still a bit skeptic about: there is no any test confirming that it works.
And it means there is no any guarantee that it will keep working while merging other changes

snuyanzin:
ok i see... loading files from other jars looks strange for me...
anyway
>The error I get when I test the case above is
java.lang.IllegalArgumentException: Path should be an existing readable file.

to fix it the condition could be tuned like 
```java
if (path == null || Files.isDirectory(path)) {
            throw new IllegalArgumentException("Path should be an existing file");
        }
```

and no need for a new methods and so on.
the only missing thing still absence of a specific test for  that in tests

snuyanzin:
Can you share the code giving an error?


I tested the code from https://github.com/datafaker-net/datafaker/issues/714#issuecomment-1451568692

and it works ok

snuyanzin:
and what it is the trace in case it is failing?
to be honest it is not reproduced on my machine.... 
that's another point to have a dedicated test for this in repo

snuyanzin:
Thanks for your explanation @tcttk 
Finally I was able to reproduce it...
From one side it's not trivial to create an integration test for this, from another side I made `addPath` working via `addUrl` so most of the time the code is reused and at covered by tests for old `addPath`. 
Please double check if it keeps working for your use case and let us know if you are ok with such changes.

Issue #675, opened 2023-02-08
Datafaker annotations


snuyanzin:
Currently this is just an idea out loud. So feel free to criticize/propose improvements or something completely new.

The idea is to have 
1. class level annotation with locale and see info (optional)
2. field level annotations with method name
3. after all validations passed it should be able to generate objects with fields filled based on datafaker specific set annotations.

bodiam:
Something like:

```
@FakeLocale("nl", "nl")
class Person {
   @FakeValue("name.firstName")
    private String firstName;
}
```

And then:

```
val persons: List<Person> = faker.generate(Person::class, 100)
```

?

kingthorin:
I like this idea, but I'll be honest I don't know much about custom annotations.

snuyanzin:
kind of
```java
@Fake(country="nl", language="nl", seed = 31415)
class Person {
   @Provider(expression="#{...faker expression here ...}")
    private String firstname;
    @MethodProvider("methodToUse")
   private Address address;
}
```
`Fake` annotation takes locale specific and seed info
`Provider` works only for string fields since expressions could result in string only
`MethodProvider` works for any object however there should be validation done that types match.
Here could be 2 possibilities: 
1. `MethodProvider` has same type and then it just sets the return value
2. `MethodProvider` returns a collection or an array and then only one value will be picked to set

ssimiao:
would these notes stay in the domain class? I believe it would be polluted if that same class has other annotations, but I liked the idea

snuyanzin:
in the class which will be faked.
Another way (without annotation pollution) is https://github.com/datafaker-net/datafaker/pull/513
or custom builders however it requires more code

bodiam:
I don't really see the need for the top level annotation but yes, I think this could be helpful for classes which are used for generating test data only.

bodiam:
> I like this idea, but I'll be honest I don't know much about custom annotations.

The idea is that you create a new annotation class with some fields, just like a normal class, and using reflection you can check if the class has the annotation, get access to it and its fields, and handle them how you see fit. It's less magic than it seems.

bodiam:
> would these annotations stay in the domain class? I believe it would be polluted if that same class has other annotations, but I liked the idea

They would, but I wouldn't recommend mixing production classes, like hibernate or Jackson annotated classes with classes typically used for testing, like Datafaker classes. But that depends mostly on the use case I guess.

snuyanzin:
[...]

PR #735, opened 2023-03-19
Add annotation support for data classes and records

RVRhub:
PR adds annotation support for data class based on issue: #675 
An example of usage

Data class definition:  
```java
@Fake(languageTag = "en-en", seed = 10)
public static class Person {

  @FieldFake(method = "net.datafaker.providers.base.Name#fullName", seed = 1)
  private String seed;

  @FieldFake(method = "net.datafaker.providers.base.Name#fullName", languageTag = "de-de")
  private String germanName;

  @FieldFake(method = "net.datafaker.providers.base.Name#fullName", languageTag = "de-de", seed = 2)
        private String germanNameAndSeedTwo;

  @FieldFake(method = "net.datafaker.providers.base.Name#fullName")
  private String defaultName;
}
```

Object generation example:
```java
  var person = Faker.generate(Person.class);
```


src/main/java/net/datafaker/annotaions/FakeProvider.java
```diff
@@ -0,0 +1,82 @@
...
+                setChildObject(object, field, childObject);
+            }
+        }
+
+        return object;
+    }
+
+    private <K> void setField(K object, Field field, Object fieldValue) {
+        try {
+            if (fieldValue instanceof String fieldValueString) {
```
bodiam, line 53: I don't fully understand what's happening here. Why are strings special here?

snuyanzin:
Thanks for your contribution @RVRhub 
I will try to have a look closer to the end of the week

src/main/java/net/datafaker/annotations/FakeResolver.java
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        if (!clazz.isAnnotationPresent(Fake.class)) {
+            throw new RuntimeException("The class %s is not annotated with Fake".formatted(clazz.getSimpleName()));
+        }
+    }
+
+    private static BaseFaker createFaker(Fake fakeAnnotation, String classLanguageTag, long[] classSeed) {
+        if (!classLanguageTag.isEmpty() && classSeed.length != 0) {
+            return new BaseFaker(new Locale(fakeAnnotation.languageTag()), new Random(classSeed[0]));
```
snuyanzin, line 45: It's much better to use `Locale#forLanguageTag` since current constructor is deprecated since jdk19

src/main/java/net/datafaker/annotations/FakeResolver.java
```diff
@@ -0,0 +1,54 @@
...
+        } else if (!classLanguageTag.isEmpty()) {
+            return new BaseFaker(new Locale(fakeAnnotation.languageTag()));
```
snuyanzin, line 47: It's much better to use `Locale#forLanguageTag` since current constructor is deprecated since jdk19

src/test/java/net/datafaker/annotations/EmbeddedFakeProviderTest.java
```diff
@@ -0,0 +1,36 @@
...
+        assertThat(person.childPerson.firstName).isNotNull();
+    }
+
+    @Fake(languageTag = "fr-en", seed = 1)
+    public static class ParentPerson {
+
+        @FieldFake(expression = "#{Name.first_name}")
+        private String firstName;
+
+        @EmbeddedFake
```
snuyanzin, line 26: What does it mean and how it works?

src/test/java/net/datafaker/annotations/EmbeddedFakeProviderTest.java
```diff
@@ -0,0 +1,36 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class EmbeddedFakeProviderTest {
+
+    @Test
+    void shouldGenerateEntityWithInheritance() {
+        var person = Faker.generate(ParentPerson.class);
+
+        assertThat(person).isNotNull();
+        assertThat(person.firstName).isNotNull();
```
snuyanzin, line 15: Instead of checking for `notNull` i guess we could check against real values.
Since seed is same shouldn't it be constant value each time?

src/test/java/net/datafaker/annotations/FakeRecordProviderTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class FakeRecordProviderTest {
+
+    @Test
+    void shouldGenerateRecordWithAnnotationFake() {
+        var person = Faker.generate(RecordPerson.class);
+
+        assertThat(person).isNotNull();
+        assertThat(person.birthTimestamp).isNotNull();
+        assertThat(person.firstName).isNotNull();
```
snuyanzin, line 18: same question about constant seed and tests against real values

src/main/java/net/datafaker/annotations/EmbeddedFake.java
```diff
@@ -0,0 +1,10 @@
+package net.datafaker.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface EmbeddedFake { }
```
bodiam, line 10: Would it be possible to add a small bit of documentation on this? It's not really clear to me what an EmbeddedFake is.

RVRhub:
In favor of a more flexible solution what is implemented in PR: https://github.com/datafaker-net/datafaker/pull/754

PR #739, opened 2023-03-28
Refactoring - Implementation refactoring

almasfiza:
This PR has three different implementation refactoring performed.
1. Extract method refactoring in datafaker/src/main/java/net/datafaker/providers/base/Barcode.java (used enhanced switch as suggested in issue)
2.  Rename variable for better understanding in datafaker/src/main/java/net/datafaker/service/FakerIDN.java (renamed Stringbuilder sb to resultASCII since it stores the ASCII equivalent).
3.  Decompose conditional refactoring in src/main/java/net/datafaker/providers/base/Number.java ( creating a method isValidRange to return the result of the condition instead of using the long condition in the if statement directly).

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Barcode.java
+++ b/src/main/java/net/datafaker/providers/base/Barcode.java
@@ -67,13 +67,7 @@ public class Barcode extends AbstractProvider<BaseProviders> {
             even = odd;
             odd = tmp;
         }
-        int var = 0;
-
-        if (length == 13) {
-            var = odd + even + (even << 1);
-        } else if (length == 8 || length == 14 || length == 12) {
-            var = odd + even + (odd << 1);
-        }
+        int var = calculateVar(length, odd, even);
 
         int rounded = roundToHighestMultiplyOfTen((var));
         int checkDigit = rounded - var;
@@ -82,6 +76,19 @@ public class Barcode extends AbstractProvider<BaseProviders> {
             product *= 10;
         }
         return firstPart * product + checkDigit;
+    }
+
+    // Implementation Refactoring: Extract Method- to calculate the var and return it to ean()
+    private int calculateVar(int length, int odd, int even) {
+        int var = 0;
+
+        if (length == 13) {
+            var = odd + even + (even << 1);
+        } else if (length == 8 || length == 14 || length == 12) {
+            var = odd + even + (odd << 1);
+        }
+
+        return var;
     }
 
     public String type() {
--- a/src/main/java/net/datafaker/providers/base/Barcode.java
+++ b/src/main/java/net/datafaker/providers/base/Barcode.java
@@ -67,10 +67,9 @@ public class Barcode extends AbstractProvider<BaseProviders> {
             even = odd;
             odd = tmp;
         }
-        int var = calculateVar(length, odd, even);
-
-        int rounded = roundToHighestMultiplyOfTen((var));
-        int checkDigit = rounded - var;
+        int output = calculateOutput(length, odd, even);
+        int rounded = roundToHighestMultiplyOfTen((output));
+        int checkDigit = rounded - output;
         int product = 10;
         while (product <= checkDigit) {
             product *= 10;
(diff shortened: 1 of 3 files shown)


src/main/java/net/datafaker/providers/base/Barcode.java
```diff
@@ -67,23 +67,24 @@ private long ean(int length) {
...
         while (product <= checkDigit) {
             product *= 10;
         }
         return firstPart * product + checkDigit;
     }
 
+    private int calculateOutput(int length, int odd, int even) {
+        return switch(length) {
+            case 13 -> odd + even + (even << 1);
+            case 8, 14, 12 -> odd + even + (odd << 1);
```
snuyanzin, line 83:
```suggestion
            case 8, 12, 14 -> odd + even + (odd << 1);
```
if we are talking about improvement of readability then it's better to keep ordering

src/main/java/net/datafaker/providers/base/Number.java
```diff
@@ -52,12 +52,18 @@ public int numberBetween(int min, int max) {
...
-        if (realMax - realMin >= realMin && (realMin >= 0 || realMax - realMin >= realMax)) {
+        if (isValidRange(realMin, realMax)) {
             return faker.random().nextInt(realMax - realMin) + realMin;
         }
         return decimalBetween(realMin, realMax).intValue();
     }
 
+   
+    private boolean isValidRange(int realMin, int realMax) {
+        return realMax - realMin >= realMin && (realMin >= 0 || realMax - realMin >= realMax);
```
snuyanzin, line 63: Probably `realMax - realMin` could be extracted into a var called e.g. `amplitude`

src/main/java/net/datafaker/providers/base/Barcode.java
```diff
@@ -67,23 +67,24 @@ private long ean(int length) {
...
-            var = odd + even + (even << 1);
-        } else if (length == 8 || length == 14 || length == 12) {
-            var = odd + even + (odd << 1);
-        }
-
-        int rounded = roundToHighestMultiplyOfTen((var));
-        int checkDigit = rounded - var;
+        int output = calculateOutput(length, odd, even);
+        int rounded = roundToHighestMultiplyOfTen((output));
+        int checkDigit = rounded - output;
```
snuyanzin, line 72:
```suggestion
        final int output = calculateOutput(length, odd, even);
        final int rounded = roundToHighestMultiplyOfTen((output));
        final int checkDigit = rounded - output;
```
why not final?

src/main/java/net/datafaker/providers/base/Barcode.java
```diff
@@ -67,23 +67,24 @@ private long ean(int length) {
...
-
-        if (length == 13) {
-            var = odd + even + (even << 1);
-        } else if (length == 8 || length == 14 || length == 12) {
-            var = odd + even + (odd << 1);
-        }
-
-        int rounded = roundToHighestMultiplyOfTen((var));
-        int checkDigit = rounded - var;
+        int output = calculateOutput(length, odd, even);
```
snuyanzin, line 70: frankly speaking i'm not sure that `output` is a good name here... Did you check how this values called in normal `ean` generation process?

src/main/java/net/datafaker/providers/base/Number.java
```diff
@@ -68,12 +74,16 @@ public long numberBetween(long min, long max) {
         if (min == max) return min;
         final long realMin = Math.min(min, max);
         final long realMax = Math.max(min, max);
-        if (realMax - realMin >= realMin && (realMin >= 0 || realMax - realMin >= realMax)) {
+        if (isValidRangeLong(realMin, realMax)) {
```
snuyanzin, line 77: why do we need `long` in name if we already have `long` in types?

bodiam:
Strange PR. Is this a test of an AI refactoring or so?

src/main/java/net/datafaker/providers/base/Barcode.java
almasfiza: > frankly speaking i'm not sure that `output` is a good name here... Did you check how this values called in normal `ean` generation process?

It was initially named as var, since it was suggested in the issue to rename var since there is a similar keyword in Java now.

src/main/java/net/datafaker/providers/base/Barcode.java
almasfiza: > why not final?

I did not plan to change the initialization of variables, hence.

src/main/java/net/datafaker/providers/base/Number.java
almasfiza: > Probably `realMax - realMin` could be extracted into a var called e.g. `amplitude`

Changed.

src/main/java/net/datafaker/providers/base/Barcode.java
almasfiza: Changed

almasfiza:
> Strange PR. Is this a test of an AI refactoring or so?

No, I am working on an assignment, which is on refactoring and encourages us to contribute to open source repositories.

bodiam:
> No, I am working on an assignment, which is on refactoring and encourages us to contribute to open source repositories.

Ah, that explains! I was surprised by the overly structured changeset, normally it's something like "changed a thing", and not used to such extended descriptions.

Thank you for picking this project for your assignment. Happy to integrate your contributions into this project!

bodiam:
(Though it seems the build is failing)

almasfiza:
> (Though it seems the build is failing)

I will check on that and open a fresh PR. Any suggestions on how I should proceed?

PR #742, opened 2023-03-29
OscarMovie should store rules to access data but not the actual field names

### LINKED ISSUE #741: Data provider `Oscar Movie` doesn't generate new values on each iteration
**Describe the bug**
When using the data provider `Oscar Movie`, all generated values are the same as previous results. 

**To Reproduce**
See https://github.com/MartijnVisser/datafaker-oscarmovie/blob/main/src/Main.java. Every run will generate the same 3 lines, for example:

```
Zero Dark Thirty
Zero Dark Thirty
Zero Dark Thirty
```

```
Sound of Metal
Sound of Metal
Sound of Metal
```

**Expected behavior**
A new value should be generate during each iteration, e.g.:
```
Zero Dark Thirty
Room
Sound of Metal
```

**Versions:**
 - OS: Macbook Pro Ventura 13.2.1 (22D68)
 - JDK: Temurin 17.0.5-tem
 - Faker Version [e.g. 22]. Tried both 1.8.1 and 2.0.0-SNAPSHOT

**Additional context**

snuyanzin:
The issue is that `OscarMovie` stores field names in `year`, `choice`, `str` which are used to find out the next result. 
Providers should not do that, so just replace it with suppliers storing only the rules how to get such data but not the actual field names

fixes #741

### DIFF
--- a/src/test/java/net/datafaker/providers/entertainment/OscarMovieTest.java
+++ b/src/test/java/net/datafaker/providers/entertainment/OscarMovieTest.java
@@ -1,6 +1,14 @@
 package net.datafaker.providers.entertainment;
 
 import org.junit.jupiter.api.RepeatedTest;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.util.Strings.isNullOrEmpty;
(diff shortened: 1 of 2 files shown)


src/test/java/net/datafaker/providers/entertainment/OscarMovieTest.java
```diff
@@ -33,4 +41,29 @@ void character() {
...
+     */
+    @ParameterizedTest
+    @MethodSource("argsProvider")
+    void issue741(Function<OscarMovie, String> f) {
+        final OscarMovie oscarMovie = faker.oscarMovie();
+        assertThat(
+            faker.stream(() -> f.apply(oscarMovie)).len(10).build()
+                .<Stream<?>>get().collect(Collectors.toSet()))
+            .hasSizeGreaterThan(1);
+    }
```
kingthorin, line 56: This doesn't really test that the values aren't consistent, just that something is generated. (Or I'm missing something?)

snuyanzin: the problem was that it generated same values for a series of invocations like
 ```java
     Faker faker = new Faker();

    String movie1 = faker.oscarMovie().movieName();
    String movie2 = faker.oscarMovie().movieName();
    String movie3 = faker.oscarMovie().movieName();
    ```
before that fix it generates same values for all 3 vars.

So the test checks that the generated value are not same after 10 times of invocations

kingthorin: Sorry missed it was collecting to set which would mean unique entries. Tks

kingthorin (approved):

PR #743, opened 2023-03-29
Add test of deterministic and non deterministic provider's methods

snuyanzin:
The PR is aiming to solve #741 for all the providers (not only `OscarMovie`)
E.g. it was detected that `Mbti` has same issue.

It retrieves all the declared public methods from provider with zero arg and invokes them several times.
The result is expected to be not the same at least for some of the invocations. 

As an exception there `Deterministic` annotation added to mark methods which are expected to return same value for any number of invocations.


kingthorin (approved):

kingthorin:
DOH, needs a spotlessapply

PR #745, opened 2023-03-29
Increase numberOfTestsPerMethod to 100 for testDeterministicAndNonDeterministicProvidersReturnValues

snuyanzin:
It looks like 10 is not enough for some cases e.g. https://github.com/datafaker-net/datafaker/actions/runs/4554807351/jobs/8033243612

Also there is no need to continue invocations in case set's size is >= 2

### DIFF
--- a/src/test/java/net/datafaker/FakerTest.java
+++ b/src/test/java/net/datafaker/FakerTest.java
@@ -364,21 +364,22 @@ class FakerTest extends AbstractFakerTest {
 
     @Test
     void testDeterministicAndNonDeterministicProvidersReturnValues() {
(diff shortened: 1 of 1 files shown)


kingthorin (approved):
Logically this should be fine, but technically we could end-up with situations where low choice sets (sex, boolean, etc) do generate the same value 100 times. (Improbable but not impossible.) In such cases the failing GitHub action/workflow could just be re-run from the WebUI.

snuyanzin:
yes, should be very rare case.
Locally I ran this test 1000 times in a row and no failures

PR #746, opened 2023-03-30
Move method refactoring.

almasfiza:
normalizeLocale() logic seems better to be present in the SingletonLocale.java than FakerContext.java. Applied Move method refactoring. I think it increases the cohesiveness of SingletonLocale.java class.

### DIFF
--- a/src/main/java/net/datafaker/internal/helper/SingletonLocale.java
+++ b/src/main/java/net/datafaker/internal/helper/SingletonLocale.java
@@ -1,8 +1,12 @@
 package net.datafaker.internal.helper;
 
+import net.datafaker.service.FakerContext;
+
 import java.util.HashMap;
+import java.util.IdentityHashMap;
 import java.util.Locale;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 /**
  * This class allows to use {@link java.util.IdentityHashMap} for locales.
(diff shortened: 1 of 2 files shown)


kingthorin:
I believe it was in FakerContext because locales are used across different fakers to provide localized values.

almasfiza:
> I believe it was in FakerContext because locales are used across different fakers to provide localized values.

Oh, I see. However, does this change make the code better in any way, in your opinion?

snuyanzin:
I do not see any good reason to make private method public especially method having some synchronized blocks insides.
Can you justify this decision?

almasfiza:
> I do not see any good reason to make private method public especially method having some synchronized blocks insides. Can you justify this decision?

I am just looking into the repository to find ways to refactor it on architecture level- move method, extract class and so on. I thought that the method would better belong to the SingletonLocale, and hence it would be more cohesive. I see your point on making the private method public :).  I will keep looking at the repo for other refactoring! Thank you!

bodiam:
I'm not sure if refactoring for the sake of refactoring is the best use of time. I personally have more success by implementing a feature, and refactoring some code as a result of this change. 

You can always move stuff around, or extract things, but that doesn't mean things are necessarily better. 

Maybe have a look at the PR which has the new annotation support? Maybe something can be improved to make that one even better?

almasfiza:
> I'm not sure if refactoring for the sake of refactoring is the best use of time. I personally have more success by implementing a feature, and refactoring some code as a result of this change.
> 
> You can always move stuff around, or extract things, but that doesn't mean things are necessarily better.
> 
> Maybe have a look at the PR which has the new annotation support? Maybe something can be improved to make that one even better?

I get your point, yes it would be better tor refactor as we code. Thank you!

PR #749, opened 2023-03-31
Update spotless and refactor WordUtils

snuyanzin:


src/main/java/net/datafaker/internal/helper/WordUtils.java
```diff
@@ -1,13 +1,13 @@
...
-import java.util.Locale;
 
 public class WordUtils {
 
     public static String capitalize(String input) {
         if (input == null) return null;
-        if (input.equals("")) return "";
-        if (Character.isUpperCase(input.charAt(0))) return input;
-        return input.substring(0, 1).toUpperCase(Locale.ROOT) + input.substring(1);
+        if (input.isEmpty()) return "";
```
kingthorin, line 8: Why not just return input, instead of instantiating a litteral?

snuyanzin: good point, sorry didn't notice and merged

PR #754, opened 2023-04-02
Add annotation support for data classes based on JavaObjectTransformer

RVRhub:
PR adds annotation support for data class based on issue: https://github.com/datafaker-net/datafaker/issues/675
After some discussion, the idea came up to:
- implement only one annotation: @Fake("defaultSchema") to specify the default schema 
- add two methods to BaseFaker for populating objects that use the JavaObjectTransformer.
An example of usage

Data class definition:
```java
@FakeForSchema("net.datafaker.annotations.FakeAnnotationTest#defaultSchema")
public class Person {
    private String fullName;

    public String getFullName() {
        return fullName;
    }

    public void setFullName(String fullName) {
        this.fullName = fullName;
    }
}
```

> If a schema is defined in the same class as the population call, the full path with the class name is not required. Example: @FakeForSchema("defaultSchema")

Schema definition:
```java
 public static Schema<Object, ?> defaultSchema() {
        var faker = new Faker(Locale.forLanguageTag("fr-en"), new RandomService(new Random(1)));
        return Schema.of(field("fullName", () -> faker.name().fullName()));
    }
```

Object generation with default schema:
```java
var person = Faker.populate(Person.class);
```

Object generation with custom schema:
```java
var person = Faker.populate(Person.class, customSchema());
```


src/main/java/net/datafaker/annotations/FakeResolver.java
```diff
@@ -0,0 +1,67 @@
...
+            return defaultSchema[0].split("#");
+        } else {
+            String[] classAndMethod = new String[2];
+            classAndMethod[0] = callingClassName;
+            classAndMethod[1] = defaultSchema[0];
+            return classAndMethod;
+        }
+    }
+
+    private Schema<Object, ?> getSchema(Schema<Object, ?> schema, String callingClassName, String[] pathToSchema) {
```
snuyanzin, line 39:
```suggestion
    private Schema<Object, ?> getSchema(Schema<Object, ?> schema, Class<?> callingClass, String[] pathToSchema) {
```

src/main/java/net/datafaker/annotations/FakeResolver.java
```diff
@@ -0,0 +1,67 @@
...
+        if (schema != null) {
+            return schema;
+        } else if (pathToSchema.length != 0) {
+            try {
+                String[] classNameAndMethod = getClassNameAndMethod(callingClassName, pathToSchema);
+                Class<?> myClass = Class.forName(classNameAndMethod[0]);
+                Method myStaticMethod = myClass.getMethod(classNameAndMethod[1]);
+                myStaticMethod.setAccessible(true);
+                return (Schema<Object, ?>) myStaticMethod.invoke(null);
```
snuyanzin, line 48:
```suggestion
                String[] classNameAndMethod = getClassNameAndMethod(callingClassName, pathToSchema);
                Class<?> myClass = Class.forName(classNameAndMethod[0]);
                Method myStaticMethod = myClass.getMethod(classNameAndMethod[1]);
                myStaticMethod.setAccessible(true);
                return (Schema<Object, ?>) myStaticMethod.invoke(null);
```
final int sharpIndex = pathToSchema[0].indexOf('#');
                final Class<?> classToCall;
                final String methodName;
                if (sharpIndex >= 0) {
                    classToCall = Class.forName(pathToSchema[0].substring(0, sharpIndex));
                    methodName = pathToSchema[0].substring(sharpIndex + 1);
                } else {
                    classToCall = callingClass;
                    methodName = pathToSchema[0];
                }
                Method myStaticMethod = classToCall.getMethod(methodName);

src/main/java/net/datafaker/annotations/FakeResolver.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class FakeResolver<T> {
+
+    private final Class<T> clazz;
+
+    public FakeResolver(Class<T> clazz) {
+        this.clazz = clazz;
+    }
+
+    public T get(String callingClassName, Schema<Object, ?> schema) {
```
snuyanzin, line 18:
```suggestion
    public T get(Class<?> callingClass, Schema<Object, ?> schema) {
```

src/main/java/net/datafaker/annotations/FakeResolver.java
```diff
@@ -0,0 +1,67 @@
...
+        checkFakeAnnotation(clazz);
+
+        Fake fakeAnnotation = clazz.getAnnotation(Fake.class);
+        Schema<Object, ?> useSchema = getSchema(schema, callingClassName, fakeAnnotation.value());
```
snuyanzin, line 22:
```suggestion
       Schema<Object, ?> useSchema = getSchema(schema, callingClass, fakeAnnotation.value());
```

src/main/java/net/datafaker/annotations/FakeResolver.java
```diff
@@ -0,0 +1,67 @@
...
+    private static String[] getClassNameAndMethod(String callingClassName, String[] defaultSchema) {
+        if (defaultSchema.length != 0 && defaultSchema[0].contains("#")) {
+            return defaultSchema[0].split("#");
+        } else {
+            String[] classAndMethod = new String[2];
+            classAndMethod[0] = callingClassName;
+            classAndMethod[1] = defaultSchema[0];
+            return classAndMethod;
+        }
+    }
```
snuyanzin, line 37: This method could be completely removed once other changes are applied

src/main/java/net/datafaker/providers/base/BaseFaker.java
```diff
@@ -298,6 +300,18 @@ public void addUrl(Locale locale, URL url) {
         fakeValuesService().addUrl(locale, url);
     }
 
+    public static <T> T populate(Class<T> clazz) {
+        String callingClassName = Thread.currentThread().getStackTrace()[2].getClassName();
```
snuyanzin, line 304:
```suggestion
```

src/main/java/net/datafaker/providers/base/BaseFaker.java
```diff
@@ -298,6 +300,18 @@ public void addUrl(Locale locale, URL url) {
...
+        var fakeFactory = new FakeResolver<>(clazz);
+        return fakeFactory.get(callingClassName, null);
```
snuyanzin, line 306:
```suggestion
        return fakeFactory.get(clazz.getEnclosingClass(), null);
```

src/main/java/net/datafaker/providers/base/BaseFaker.java
```diff
@@ -298,6 +300,18 @@ public void addUrl(Locale locale, URL url) {
...
+    }
+
+    public static <T> T populate(Class<T> clazz, Schema<Object, ?> schema) {
+        String callingClassName = Thread.currentThread().getStackTrace()[2].getClassName();
+        var fakeFactory = new FakeResolver<>(clazz);
+        return fakeFactory.get(callingClassName, schema);
```
snuyanzin, line 312:
```suggestion
        return fakeFactory.get(clazz.getEnclosingClass(), schema);
```

src/main/java/net/datafaker/providers/base/BaseFaker.java
```diff
@@ -298,6 +300,18 @@ public void addUrl(Locale locale, URL url) {
...
     }
 
+    public static <T> T populate(Class<T> clazz) {
+        String callingClassName = Thread.currentThread().getStackTrace()[2].getClassName();
+        var fakeFactory = new FakeResolver<>(clazz);
+        return fakeFactory.get(callingClassName, null);
+    }
+
+    public static <T> T populate(Class<T> clazz, Schema<Object, ?> schema) {
+        String callingClassName = Thread.currentThread().getStackTrace()[2].getClassName();
```
snuyanzin, line 310:
```suggestion
```

src/main/java/net/datafaker/annotations/FakeResolver.java
RVRhub: Calling class was removed.

src/main/java/net/datafaker/Faker.java
```diff
@@ -36,4 +36,5 @@ public Faker(Locale locale, RandomService randomService) {
     public Faker(FakeValuesService fakeValuesService, FakerContext context) {
         super(fakeValuesService, context);
     }
+
```
snuyanzin, line 39: this line is not required and could be removed

src/main/java/net/datafaker/annotations/FakeResolver.java
```diff
@@ -0,0 +1,64 @@
...
+            }
+        } else {
+            return null;
+        }
+    }
+
+    private void checkFakeAnnotation(Class<T> clazz) {
+        if (Objects.isNull(clazz)) {
+            throw new RuntimeException("The class is null.");
+        }
```
snuyanzin, line 57:
```suggestion
        Objects.requireNonNull(clazz, "The class is null.");
```
could be simplified

src/test/java/net/datafaker/annotations/FakeAnnotationTest.java
```diff
@@ -0,0 +1,58 @@
...
+        assertThat(person.getFullName()).isEqualTo("Dr Alexis Noël");
+    }
+
+    @Test
+    void shouldGenerateEntityWithCustomSchema() {
+        var person = Faker.populate(Person.class, customSchema());
+
+        assertThat(person).isNotNull();
+        assertThat(person.getFullName()).isNotNull();
+        assertThat(person.getFullName()).isEqualTo("Dr Alexis Noël");
```
snuyanzin, line 32: no need to check for `isNotNull` since `isEqualTo` does it under the hood

src/test/java/net/datafaker/annotations/FakeAnnotationTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class FakeAnnotationTest {
+
+    @Test
+    void shouldGenerateEntityWithDefaultSchema() {
+        var person = Faker.populate(Person.class);
+
+        assertThat(person).isNotNull();
+        assertThat(person.getFullName()).isNotNull();
+        assertThat(person.getFullName()).isEqualTo("Dr Alexis Noël");
```
snuyanzin, line 23: no need to check for `isNotNull` since `isEqualTo` does it under the hood

src/main/java/net/datafaker/providers/base/BaseFaker.java
```diff
@@ -298,6 +300,16 @@ public void addUrl(Locale locale, URL url) {
         fakeValuesService().addUrl(locale, url);
     }
 
+    public static <T> T populate(Class<T> clazz) {
+        var fakeFactory = new FakeResolver<>(clazz);
+        return fakeFactory.get(null);
```
snuyanzin, line 305: Not sure that `get` is a right name here.... Probably something like `generate` would be better since it really generates

src/main/java/net/datafaker/providers/base/BaseFaker.java
```diff
@@ -298,6 +300,16 @@ public void addUrl(Locale locale, URL url) {
...
+    }
+
+    public static <T> T populate(Class<T> clazz, Schema<Object, ?> schema) {
+        var fakeFactory = new FakeResolver<>(clazz);
+        return fakeFactory.get(schema);
```
snuyanzin, line 310: Not sure that `get` is a right name here.... Probably something like `generate` would be better since it really generates

src/test/java/net/datafaker/annotations/FakeAnnotationTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public static Schema<Object, ?> defaultSchema() {
+        var faker = new Faker(Locale.forLanguageTag("fr-en"), new RandomService(new Random(1)));
+        return Schema.of(field("fullName", () -> faker.name().fullName()));
+    }
+
+    public static Schema<Object, ?> customSchema() {
+        var faker = new Faker(Locale.forLanguageTag("fr-en"), new RandomService(new Random(1)));
+        return Schema.of(field("fullName", () -> faker.name().name()));
+    }
```
snuyanzin, line 49: It would be better to have custom and default schemas different since if you look at implementation of `Name#name`
```java
    public String fullName() {
        return name();
    }
```
they are the same =)

snuyanzin:
thanks for your contribution @RVRhub 
lgtm from one side
however it would be great to have some documentation for that

//cc @bodiam since you mentioning something about that
with this PR it will be possible to make calls like
```java
var person = Faker.populate(Person.class);
// or
var person2 = Faker.populate(Person.class, customSchema());
```
schema could be defined separately

bodiam:
@snuyanzin's sorry, I missed some of the conversions on this. Is the annotation support now dropped in favor of some schema method?

snuyanzin:
There are 2 reasons

After having several chats and POCs with @RVRhub we faced 2 issues:
1. it looks like the code is becoming polluted with annotations especially for classes with multiple fields
2. it's impossible to have several objects generated in different ways.

To address these issues we decided to have only one annotation `@FakeForSchema` which is a reference to a method providing a default schema to generate objects.

The idea is to build the solution based on java transformers approach https://github.com/datafaker-net/datafaker/pull/513 which already can generate objects.

So the objects could be generated in 2 ways:
1. 
```java
MyClass object = Faker.populate(MyClass.class);
```
this will generate an object with usage of default schema mentioned by  `@FakeForSchema` 
2. 
```java
MyClass object = Faker.populate(MyClass.class, myschema);
```

this will allow to generate objects based on others schemas.

In this way it's possible to have a number of different schemas and generate objects based on them.
At the same time schemas and objects generations are decoupled

bodiam:
Okay, makes sense. Thanks for the explanation!

docs/documentation/schemas.md
```diff
@@ -346,3 +346,133 @@ address:
...
+        val schema: Schema<Any, Any> = Schema.of(
+            field("firstName", Supplier { faker.name().firstName() }),
+            field("lastName", Supplier { faker.name().lastName() }),
+            field("birthDate", Supplier { faker.date().birthday() }),
+            field("id", Supplier { faker.number().positive() }))
+
+        println(jTransformer.apply(Person::class.java, schema))
+    ```
+
+will generate object with fields populated with random values.
```
snuyanzin, line 408:
```suggestion
will generate object with fields populated with random values based on specified suppliers.
```

docs/documentation/schemas.md
```diff
@@ -346,3 +346,133 @@ address:
...
+
+    ```kotlin
+        @FakeForSchema("net.datafaker.annotations.FakeAnnotationTest#defaultSchema")
+        data class Person(
+            var fullName: String
+        )
+    ```
+
+Then you can use `net.datafaker.providers.base.BaseFaker.populate(java.lang.Class<T>)` to populate object with default predefined schema.
+Or you can use `net.datafaker.providers.base.BaseFaker.populate(java.lang.Class<T>, net.datafaker.schema.Schema<java.lang.Object, ?>)` to populate object with custom schema.
```
snuyanzin, line 464: Probably need code snippets for both cases

RVRhub: Added one more snippet of object template population with custom schema

snuyanzin (approved):

Issue #759, opened 2023-04-10
FakeValuesService is not thread-safe


rwong-gw:
**Describe the bug**
Using Gatling to simulate hundreds of users making requests to a server, data will occasionally become corrupted, causing all threads to eventually hang or deadlock. A thread dump taken after a deadlock indicates a problem with `java.util.WeakHashMap`, which is known to be [not thread-safe](http://dowbecki.com/WeakHashMap-is-not-thread-safe/):
```
"Thread-62" prio=5 Id=78 RUNNABLE
	at java.base@11.0.15/java.util.WeakHashMap.transfer(WeakHashMap.java:529)
	at java.base@11.0.15/java.util.WeakHashMap.resize(WeakHashMap.java:506)
	at java.base@11.0.15/java.util.WeakHashMap.put(WeakHashMap.java:467)
	at app//net.datafaker.service.FakeValuesService.split(FakeValuesService.java:246)
	at app//net.datafaker.service.FakeValuesService.fetchObject(FakeValuesService.java:195)
	at app//net.datafaker.service.FakeValuesService.safeFetch(FakeValuesService.java:155)
	at app//net.datafaker.service.FakeValuesService.resolve(FakeValuesService.java:408)
	at app//net.datafaker.providers.base.AbstractProvider.resolve(AbstractProvider.java:21)
	...

"Thread-229" prio=5 Id=245 RUNNABLE
	at java.base@11.0.15/java.util.WeakHashMap.transfer(WeakHashMap.java:529)
	at java.base@11.0.15/java.util.WeakHashMap.resize(WeakHashMap.java:506)
	at java.base@11.0.15/java.util.WeakHashMap.put(WeakHashMap.java:467)
	at app//net.datafaker.service.FakeValuesService.split(FakeValuesService.java:246)
	at app//net.datafaker.service.FakeValuesService.fetchObject(FakeValuesService.java:195)
	at app//net.datafaker.service.FakeValuesService.safeFetch(FakeValuesService.java:155)
	at app//net.datafaker.service.FakeValuesService.resolve(FakeValuesService.java:408)
	at app//net.datafaker.providers.base.AbstractProvider.resolve(AbstractProvider.java:21)
	...

"Thread-285" prio=5 Id=301 RUNNABLE
	at java.base@11.0.15/java.util.WeakHashMap.transfer(WeakHashMap.java:529)
	at java.base@11.0.15/java.util.WeakHashMap.resize(WeakHashMap.java:494)
[...]

PR #760, opened 2023-04-11
Fix concurrency issues in FakeValuesService

rwong-gw:
* Replace  IdentityHashMaps with ConcurrentHashMaps
* Replace WeakHashMaps with Guava weak-key caches
* Add dependency for Guava 31.1-jre


src/main/java/net/datafaker/service/FakeValuesService.java
```diff
@@ -1,5 +1,7 @@
 package net.datafaker.service;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
```
kingthorin, line 4: I think these can go?

src/main/java/net/datafaker/service/FakeValuesService.java
```diff
@@ -22,17 +24,8 @@
...
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.IdentityHashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Objects;
-import java.util.WeakHashMap;
+import java.util.*;
```
kingthorin, line 27: Specific imports should be used vs wildcard

src/main/java/net/datafaker/service/FakeValuesService.java
```diff
@@ -1,19 +1,9 @@
...
-import net.datafaker.providers.base.BaseFaker;
-import net.datafaker.providers.base.Name;
-import net.datafaker.providers.base.ProviderRegistration;
-import net.datafaker.transformations.CsvTransformer;
-import net.datafaker.transformations.Field;
-import net.datafaker.transformations.JsonTransformer;
-import net.datafaker.transformations.Schema;
-import net.datafaker.transformations.SimpleField;
+import net.datafaker.providers.base.*;
+import net.datafaker.transformations.*;
```
kingthorin, line 6: 🤦‍♂️

snuyanzin (changes requested):
I'm a bit skeptic about this change without reproducible unit test.

Moreover replacement of e.g. IdentityHashMaps with ConcurrentHashMaps leads to some slow downs and this should be kept in mind

src/main/java/net/datafaker/service/FakeValuesService.java
```diff
@@ -222,27 +219,7 @@ private String[] split(String string) {
...
-        for (int i = 0; i < length; i++) {
-            if (string.charAt(i) == splitChar) {
-                if (i - start > 0) {
-                    result[j++] = String.valueOf(chars, start, i - start);
-                }
-                start = i + 1;
-            }
-        }
-        result[j] = String.valueOf(chars, start, chars.length - start);
+        result = string.split("\\.");   // String#split is plenty efficient for single-char patterns
```
snuyanzin, line 222: This statement does not look like true

please provide proof for that.

Dummy test with jmh shows that current code is about 10-15% faster than `result = string.split("\\.");`

```
Benchmark             Mode  Cnt      Score      Error   Units
Issue760.testSplit   thrpt    5  24329.658 ±  734.929  ops/ms
Issue760.testSplit2  thrpt    5  21338.631 ± 1043.678  ops/ms
```
code snippet
```java
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
@State(Scope.Benchmark)
@Fork(value = 2, jvmArgs = {"-Xms2G", "-Xmx2G"})
public class Issue760 {
    private static final String STRING_2_SPLIT = "Address.streetAddress";
    public static void main(String[] args) throws RunnerException {

        Options opt = new OptionsBuilder()
                .include(Issue760.class.getSimpleName())
                .forks(1)
                .build();

        new Runner(opt).run();
    }

    @Benchmark
    public void testSplit(Blackhole blackhole) {
        final String string = STRING_2_SPLIT;
        int size = 0;
        final char splitChar = '.';
        final int length = string.length();
        for (int i = 0; i < length; i++) {
            if (string.charAt(i) == splitChar) {
                size++;
            }
        }
        String[] result = new String[size + 1];
        final char[] chars = string.toCharArray();
        int start = 0;
        int j = 0;
        for (int i = 0; i < length; i++) {
            if (string.charAt(i) == splitChar) {
                if (i - start > 0) {
                    result[j++] = String.valueOf(chars, start, i - start);
                }
                start = i + 1;
            }
        }
        result[j] = String.valueOf(chars, start, chars.length - start);
        blackhole.consume(result);
    }

    @Benchmark
    public void testSplit2(Blackhole blackhole) {
        final String string = STRING_2_SPLIT;
        String[] result = string.split("\\.");
        blackhole.consume(result);
    }
}

```

snuyanzin:
Also see my comment https://github.com/datafaker-net/datafaker/issues/759#issuecomment-1505644458

snuyanzin:
Closing in favor of #770

PR #762, opened 2023-04-12
Fix aviation empty entry and formatting.

bodiam:


kingthorin (commented):
Should we be using/mandating quotes more specifically?

Is there something we should be adding to tests here? (Not specific to aviation, but in general)

bodiam:
> Should we be using/mandating quotes more specifically?

With what purpose?

> Is there something we should be adding to tests here? (Not specific to aviation, but in general)

Maybe. I think it would be good to make sure a result can never be an empty string or so? (I assume that would have been the result of this?)

kingthorin:
1. Just for consistency. I have no idea why half of these have quotes and half don’t. As a new contributor it would/could seem messy 🤷‍♂️
2. I’ll see if I can add a not empty condition to the test set we abstracted a few months ago.

bodiam:
Regarding 1, I had the same. I went through a few yml files to check them, but it's very inconsistent, but changing this might be a bit of an effort.

Regarding 2: thanks!

kingthorin:
Looks like we could make it consistent leveraging spotless https://github.com/diffplug/spotless/tree/main/plugin-maven#yaml

We already use spotless so shall I look into doing that?

bodiam:
@kingthorin  just remove some items and create a  list with empty elements in YAML, that should do it

kingthorin:
Do we not have `spotless:check` running as part of our CI? Because I can't get it to pass locally. It seems upset about some of the Java 17 upgrades that have been done.

```console
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.833 s
[INFO] Finished at: 2023-04-15T18:43:10-04:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.35.0:check (default-cli) on project datafaker: Unable to format file C:\pathtoworkspace\datafaker\src\main\java\net\datafaker\idnumbers\NricNumber.java: java.lang.reflect.InvocationTargetException: 19:10: error: illegal start of expression -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException

```

PR #769, opened 2023-04-15
BaseFakerTest - Assert that provider lists should not have blank entries

kingthorin:
As discussed in: https://github.com/datafaker-net/datafaker/pull/762

Tested with entries such as:
`""` (empty)
`" "` (space)
`"	"` (tab)
`"\n"` (new line literal)


bodiam:
Nice, looks good!

PR #770, opened 2023-04-16
Make FakeValuesService thread safe

snuyanzin:
The PR assumes that most of the time there is no writes to the maps and only reads.
~In this case there could be used optimistic locks most of the time and cheap `StampedLock` 's~

~based on measurements there is overhead only about 10% for some of methods, for others there is no noticeable difference~
e.g.
`Datafaker_KotlinFakerBenchmark.kotlinFakerTest`
before PR
```
Benchmark                                       Mode  Cnt    Score   Error  Units
Datafaker_KotlinFakerBenchmark.kotlinFakerTest  avgt   10  390.813 ± 2.512  ms/op
```
after (10% slower)
```
Benchmark                                       Mode  Cnt    Score    Error  Units
Datafaker_KotlinFakerBenchmark.kotlinFakerTest  avgt   10  431.599 ± 13.887  ms/op
```
`DatafakerSimpleMethods.firstname`
before
```
DatafakerSimpleMethods.firstname      thrpt   10   6016.154 ± 247.178  ops/ms
```
after (about 10% slower)
```
DatafakerSimpleMethods.firstname      thrpt   10   5330.902 ±   67.988  ops/ms
```
`DatafakerSimpleMethods.fullname`
before
```
DatafakerSimpleMethods.fullname       thrpt   10   2524.751 ±  26.856  ops/ms
```
after (about 10% slower)
```
DatafakerSimpleMethods.fullname       thrpt   10   2235.404 ±   67.011  ops/ms
```

UPD: a new approach passing added test is with `COWMap`


kingthorin (commented):
Is it possible/practical to add tests?

The changes seem sane to me. I haven’t oiled and used them at all. Or, looked for any places updates might have been missed (I feel like other probably know the code base better than I do).

snuyanzin:
that's still a tricky part to create a 100% reproducible test for failing non thread safe  FakeValuesService...
will see this week if i able to make it or not

bodiam:
Would this be something @rwong-gw could maybe help validate?

rwong-gw:
For the past couple days, I have been using CI to run my Gatling tests every half hour with this version, and everything looks good, with no hanging or deadlocks.

One suggestion: `StampedLockMap` does not implement several methods in Java's `Map` API, and just throws an `UnsupportedOperationException` for them. This causes it to interact poorly with debuggers (see attached image from IntelliJ). You should implement at least the `size` and `entrySet` methods to get it to work nicely with IntelliJ, and it probably wouldn't hurt to implement at least the remaining non-mutating operations.

StampedLockMap_UOE

snuyanzin:
makes sense, added implementation

rwong-gw:
Thanks, I can now drill down into StampedLockMaps in IntelliJ's debugger:
StampedLockMap_ro

BTW, you might want to add `@Override` annotations to the methods `putIfAbsent`, `computeIfAbsent`, and `merge`.

src/main/java/net/datafaker/service/FakeValuesService.java
```diff
@@ -48,26 +48,28 @@ public class FakeValuesService {
     private static final String[] EMPTY_ARRAY = new String[0];
     private static final Logger LOG = Logger.getLogger("faker");
 
-    private final Map<SingletonLocale, FakeValuesInterface> fakeValuesInterfaceMap = new IdentityHashMap<>();
+    private final StampedLockMap<SingletonLocale, FakeValuesInterface> fakeValuesInterfaceMap = new StampedLockMap<>(IdentityHashMap::new);
```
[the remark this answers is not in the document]
snuyanzin: yes, for some of them it makes sense, for some not since there is `updateNestedValue` method specific for `StampedLockMap`

src/main/java/net/datafaker/internal/helper/StampedLockMap.java
```diff
@@ -0,0 +1,156 @@
...
+        long stamp = lock.tryOptimisticRead();
+        V curValue = map.get(key);
+        if (!lock.validate(stamp)) {
+            stamp = lock.readLock();
+            try {
+                curValue = map.get(key);
+            } finally {
+                lock.unlockRead(stamp);
+            }
+        }
```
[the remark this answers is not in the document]
snuyanzin: in fact it was kind of test version...
As measurement shows here it's that rare case when code duplication wins. With `doROOperation` it becomes 10% slower.... The reason is that it requires `Supplier` object which is generated during runtime. Without this method there is no need for that extra object.

snuyanzin:
Funny thing is that I finally adapted the test from #759 attached by @rwong-gw 
And once I decreased sleep time it starts failing with `IllegalMonitorStateException` since it can not unlock locks ...

At the same time another option I had in mind was CopyOnWriteMaps (`COWMap`) which for some reason is not provided by java core. 

So I replaced it with `COWMap` and the problem has gone.
Another good thing that it is a bit faster than with stamplockmaps (almost same speed as with unsafe option).
Probably at the beginning it could consume a bit more memory.

@kingthorin so at least one test is added
@rwong-gw could you please once again double check your case if it is still ok? I'm asking since as mentioned above I replaced it with `COWMap` because of `IllegalMonitorStateException`

src/main/java/net/datafaker/internal/helper/COWMap.java
```diff
@@ -0,0 +1,93 @@
...
+        return map.containsValue(value);
+    }
+
+    @Override
+    public V get(Object key) {
+        return map.get(key);
+    }
+
+    @Override
+    public V put(K key, V value) {
```
NarekDW, line 43: I think this function should be `synchronized` otherwise it might produce incorrect result, don't you think?
For example 2 threads will access this function at the same time, one of the pair of key-value will be lost...

src/main/java/net/datafaker/internal/helper/COWMap.java
```diff
@@ -0,0 +1,93 @@
...
+        Map<K, V> newMap = mapSupplier.get();
+        newMap.putAll(map);
+        final V result = newMap.put(key, value);
+        map = newMap;
+        return result;
+    }
+
+    @Override
+    public V remove(Object key) {
```
NarekDW, line 52: The same, don't you think that all the modification functions should be `synchronized` ?

src/main/java/net/datafaker/internal/helper/COWMap.java
```diff
@@ -0,0 +1,93 @@
...
+        Map<K, V> newMap = mapSupplier.get();
+        newMap.putAll(map);
+        final V result = newMap.remove(key);
+        map = newMap;
+        return result;
+    }
+
+    @Override
+    public void putAll(Map<? extends K, ? extends V> m) {
```
NarekDW, line 61: `synchronized` ?

src/main/java/net/datafaker/internal/helper/COWMap.java
```diff
@@ -0,0 +1,93 @@
...
+        Map<K, V> newMap = mapSupplier.get();
+        newMap.putAll(map);
+        newMap.putAll(m);
+        map = newMap;
+    }
+
+    @Override
+    public void clear() {
```
NarekDW, line 69: `synchronized`

src/main/java/net/datafaker/internal/helper/COWMap.java
```diff
@@ -0,0 +1,93 @@
...
+        map = newMap;
+        return result;
+    }
+
+    @Override
+    public V remove(Object key) {
+        Map<K, V> newMap = mapSupplier.get();
+        newMap.putAll(map);
+        final V result = newMap.remove(key);
+        map = newMap;
```
NarekDW, line 56: Isn't it better to keep the internal map as unmodifiable? Just for safety.
Like:
```java
map = Collections.unmodifiableMap(newMap);
```

NarekDW (approved):

src/main/java/net/datafaker/internal/helper/COWMap.java
snuyanzin: No, that's the main idea of COW approach: it is lock free.
Ok let's consider the case when 2 threads accessing this map.
Yes, in worst case one of the value will be lost. However it is ok. 
We use this as a cache. It means that the lost value will be recalculated and put into map next time. 
So there is no issue.

src/main/java/net/datafaker/internal/helper/COWMap.java
snuyanzin: see the answer above

src/main/java/net/datafaker/internal/helper/COWMap.java
snuyanzin: see the answer above

src/main/java/net/datafaker/internal/helper/COWMap.java
snuyanzin: see the answer above

rwong-gw:
> @rwong-gw could you please once again double check your case if it is still ok? I'm asking since as mentioned above I replaced it with `COWMap` because of `IllegalMonitorStateException`

@snuyanzin I'll give it a try.

rwong-gw:
> @rwong-gw could you please once again double check your case if it is still ok? I'm asking since as mentioned above I replaced it with `COWMap` because of `IllegalMonitorStateException`

@snuyanzin I'll give it a try.

src/main/java/net/datafaker/internal/helper/COWMap.java
```diff
@@ -0,0 +1,93 @@
...
+        return map.values();
+    }
+
+    @Override
+    public Set<Entry<K, V>> entrySet() {
+        return map.entrySet();
+    }
+
+    public <K2, V2> void updateNestedValue(K key, Supplier<V> valueSupplier, K2 key2, V2 value) {
+        map.putIfAbsent(key, valueSupplier.get());
```
rwong-gw, line 89: If `map` is not thread-safe, this `putIfAbsent` call is susceptible to a ConcurrentModificationException.

Both `FakeValueService#key2fetchedObject` and `FakeValueService#key2Expression` use `IdentityHashMap`, which is not thread-safe.

To make this thread-safe, I believe you can use the same copy-on-write behavior as the other mutating operations:

```
    public <K2, V2> void updateNestedValue(K key, Supplier<V> valueSupplier, K2 key2, V2 value) {
        if (!map.containsKey(key)) {
            Map<K, V> newMap = mapSupplier.get();
            newMap.putAll(map);
            newMap.put(key, valueSupplier.get());
            map = newMap;
        }
        // It is assumed that nested could be only Map
        ((Map<K2, V2>)map.get(key)).put(key2, value);
    }

```

snuyanzin: good catch, thanks

done

PR #772, opened 2023-04-16
Remove unused package private method and tests

snuyanzin:
Since now tests able to read data from file to construct expected set we don't need to call tests thousand number of times


kingthorin (approved):
Thanks

PR #780, opened 2023-04-18
Updated 1.9.0 documentation

bodiam:


kingthorin:
Do we need to keep the branch? I don't see any other release branches, but I should probably ask before I just assume and remove it :grinning:

bodiam:
I don't think so :-) It's gone.

PR #783, opened 2023-04-21
Add Final Fantasy XIV faker

panilya:


src/main/java/net/datafaker/providers/videogame/FinalFantasyXIV.java
```diff
@@ -0,0 +1,37 @@
+package net.datafaker.providers.videogame;
```
kingthorin, line 1: Should we move other video games in here?
Should this be a sub-package within entertainment?

@bodiam

bodiam: Whatever we do, it will never be right. Lord of the Rings. Is that a book, movies, a game, tv series?

But I am not really bothered with how it's organized. If there's some consistency, that's nice, and if you want to move don't packages around, I won't stop you :)

kingthorin: Alright,that's fair.

PR #787, opened 2023-04-23
Add support populate Java Object with custom schema without FakeForSc…

RVRhub:
PR adds support to populate Java objects with a custom schema without using the FakeForSchema annotation. The existing implementation of `net.datafaker.providers.base.BaseFaker#populate(java.lang.Class<T>, net.datafaker.transformations.Schema<java.lang.Object,?>)` requires the use FakeForSchema annotation to generate fake data for the object.


src/main/java/net/datafaker/annotations/FakeResolver.java
```diff
@@ -9,26 +9,33 @@
...
 
-        JavaObjectTransformer jTransformer = new JavaObjectTransformer();
         return (T) jTransformer.apply(clazz, useSchema);
     }
 
-    private Schema<Object, ?> getSchema(Schema<Object, ?> schema, String[] pathToSchema) {
-        if (schema != null) {
-            return schema;
-        } else if (pathToSchema.length != 0) {
+    public T generateFromCustomSchema(Schema<Object, ?> schema) {
```
snuyanzin, line 29: why can't we keep same method name `public T generate(Schema<Object, ?> schema)` ?

src/main/java/net/datafaker/annotations/FakeResolver.java
```diff
@@ -9,26 +9,33 @@
 
 public class FakeResolver<T> {
 
+    private static final JavaObjectTransformer jTransformer = new JavaObjectTransformer();
```
snuyanzin, line 12: constant should be uppercase

src/main/java/net/datafaker/annotations/FakeResolver.java
```diff
@@ -9,26 +9,33 @@
...
-        } else if (pathToSchema.length != 0) {
+    public T generateFromCustomSchema(Schema<Object, ?> schema) {
+        if (Objects.isNull(schema)) {
+            return generateFromDefaultSchema();
+        }
+
+        return (T) jTransformer.apply(clazz, schema);
+    }
+
+    private Schema<Object, T> getSchema(String[] pathToSchema) {
```
snuyanzin, line 37: why do we need an array here if we use only the first element?
I guess just `String` should be enough

RVRhub: Changed value type from array to a single value.

src/main/java/net/datafaker/annotations/FakeResolver.java
RVRhub: Renamed constant

src/main/java/net/datafaker/annotations/FakeResolver.java
RVRhub: Made one `public T generate(Schema<Object, ?> schema)` method.

snuyanzin (approved):

snuyanzin:
lgtm

PR #792, opened 2023-04-29
Introduce cache of suppliers and remove usage of Objects.hash to speed up

snuyanzin:
The problem with `Objects.hash` is that it generates an array each time it is called.

Suppliers could be cached to avoid parsing of expressions each time

it gives speed up for a wide range of methods up to 50 %

expressions
```
before
Benchmark                        Mode  Cnt     Score     Error   Units
DatafakerExpressions.bothify    thrpt   10  5089.609 ± 112.084  ops/ms
DatafakerExpressions.letterify  thrpt   10  5706.392 ± 326.134  ops/ms
DatafakerExpressions.numerify   thrpt   10  6027.549 ±  87.064  ops/ms
DatafakerExpressions.regexify   thrpt   10  2826.806 ±  31.933  ops/ms

VS

after
Benchmark                        Mode  Cnt      Score     Error   Units
DatafakerExpressions.bothify    thrpt   10  10936.056 ±  39.143  ops/ms
DatafakerExpressions.letterify  thrpt   10  13036.411 ±  94.658  ops/ms
DatafakerExpressions.numerify   thrpt   10  13084.755 ± 423.533  ops/ms
DatafakerExpressions.regexify   thrpt   10   3541.279 ±   6.988  ops/ms
```

csv generation
```
before
Benchmark                  Mode  Cnt    Score   Error   Units
DatafakerCsvjson.csvJson  thrpt   10  301.604 ± 2.242  ops/ms

VS

after
Benchmark                  Mode  Cnt    Score    Error   Units
DatafakerCsvjson.csvJson  thrpt   10  377.927 ± 10.929  ops/ms
```

kotlin faker benchmark

```
before
Benchmark                                       Mode  Cnt    Score   Error  Units
Datafaker_KotlinFakerBenchmark.kotlinFakerTest  avgt   10  378.447 ± 1.685  ms/op

VS

after
Benchmark                                       Mode  Cnt    Score   Error  Units
Datafaker_KotlinFakerBenchmark.kotlinFakerTest  avgt   10  324.256 ± 2.438  ms/op
```

simple methods calls
```
before
DatafakerSimpleMethods.firstname      thrpt   10   5838.567 ±  508.992  ops/ms
DatafakerSimpleMethods.fullname       thrpt   10   2429.474 ±  211.390  ops/ms
DatafakerSimpleMethods.numberBetween  thrpt   10  66001.570 ± 2144.033  ops/ms
DatafakerSimpleMethods.streetAddress  thrpt   10   1724.282 ±   62.587  ops/ms

VS

after
Benchmark                              Mode  Cnt       Score     Error   Units
DatafakerSimpleMethods.firstname      thrpt   10    8972.172 ±  99.469  ops/ms
DatafakerSimpleMethods.fullname       thrpt   10    3232.930 ± 203.367  ops/ms
DatafakerSimpleMethods.numberBetween  thrpt   10  102087.734 ± 655.514  ops/ms
DatafakerSimpleMethods.streetAddress  thrpt   10    2503.758 ±  19.876  ops/ms
```

### DIFF
--- a/src/main/java/net/datafaker/service/FakeValuesContext.java
+++ b/src/main/java/net/datafaker/service/FakeValuesContext.java
@@ -89,13 +89,23 @@ class FakeValuesContext {
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
-        if (!(o instanceof FakeValuesContext that)) return false;
-        return sLocale == that.sLocale && Objects.equals(filename, that.filename) && Objects.equals(path, that.path) && Objects.equals(url, that.url);
+        if (o == null || getClass() != o.getClass()) return false;
+
+        FakeValuesContext that = (FakeValuesContext) o;
+
+        if (!Objects.equals(sLocale, that.sLocale)) return false;
+        if (!Objects.equals(filename, that.filename)) return false;
+        if (!Objects.equals(path, that.path)) return false;
+        return Objects.equals(url, that.url);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(sLocale, filename, path, url);
+        int result = sLocale != null ? sLocale.hashCode() : 0;
+        result = 31 * result + (filename != null ? filename.hashCode() : 0);
+        result = 31 * result + (path != null ? path.hashCode() : 0);
+        result = 31 * result + (url != null ? url.hashCode() : 0);
+        return result;
     }
 
     @Override
--- a/src/main/java/net/datafaker/service/FakeValuesContext.java
+++ b/src/main/java/net/datafaker/service/FakeValuesContext.java
@@ -2,6 +2,7 @@ package net.datafaker.service;
 
 import net.datafaker.internal.helper.SingletonLocale;
 
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.Locale;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


src/main/java/net/datafaker/service/FakeValuesContext.java
```diff
@@ -89,13 +89,23 @@ public URL getUrl() {
...
+        return Objects.equals(url, that.url);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(sLocale, filename, path, url);
+        int result = sLocale != null ? sLocale.hashCode() : 0;
+        result = 31 * result + (filename != null ? filename.hashCode() : 0);
+        result = 31 * result + (path != null ? path.hashCode() : 0);
+        result = 31 * result + (url != null ? url.hashCode() : 0);
```
kingthorin, line 107: Apparently qudona and other SAST suggest against this with URLs because it actually causes a DNS resolution, which can be slow or fail. 

> In general it is better to use the URI class until access to the resource is actually needed, at which point you can just convert the URI to a URL using URI.toURL().

snuyanzin: hmmm, i wonder why qudona and other SAST keep silence when we invoke `Objects.hash(args)` and then under the hood `Arrays.hashcode(args)` and the same code as above....

kingthorin: 🤷‍♂️ it’s just the URL one in particular because of the impl I guess

snuyanzin: ok, anyway changed it for url
+ added hashes as constants for constant fields since there is no need to calculate them each time
may be similar things could be done for other classes however in a separate PR + need some measurements (for this one it improves a bit about 3-4 %)

kingthorin (approved):
LGTM

PR #797, opened 2023-04-30
Add shading for generex and automaton

snuyanzin:
It seems these 2 were missed

to double check the whole list of libs that are used during runtime and should and shaded this command could be used

```
mvn dependency:tree | grep compile
```
output
```
[INFO] +- org.yaml:snakeyaml:jar:2.0:compile
[INFO] +- com.github.mifmif:generex:jar:1.0.2:compile
[INFO] |  \- dk.brics.automaton:automaton:jar:1.11-8:compile
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -241,6 +241,8 @@
                     <artifactSet>
                         <includes>
                             <include>org.yaml:snakeyaml:*:*</include>
+                            <include>com.github.mifmif:generex:*:*</include>
+                            <include>dk.brics.automaton:automaton:*:*</include>
                         </includes>
                     </artifactSet>
                     <filters>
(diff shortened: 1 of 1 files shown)


snuyanzin:
// cc @bodiam

pom.xml
```diff
@@ -241,6 +241,8 @@
                     <artifactSet>
                         <includes>
                             <include>org.yaml:snakeyaml:*:*</include>
+                            <include>com.github.mifmif:generex:*:*</include>
+                            <include>dk.brics.automaton:automaton:*:*</include>
```
bodiam, line 245: Why not just include all dependencies? It's unlikely we'll add more perhaps, but if we would, we'd need to remember to update this too.

snuyanzin: anyway you need to update relocation part individually , so it will not solve the issue

PR #824, opened 2023-05-07
Add Java Record support in JavaObjectTransformer

panilya:
At the moment, Datafaker doesn't support Java Records. This PR addresses this problem

### DIFF
--- a/src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
+++ b/src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
@@ -3,7 +3,9 @@ package net.datafaker.transformations;
 import net.datafaker.sequence.FakeSequence;
 
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.RecordComponent;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.IdentityHashMap;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
```diff
@@ -23,7 +25,32 @@ public Object apply(Object input, Schema<Object, ?> schema) {
...
+                    .toArray(Class<?>[]::new);
+
+                final Field<Object, ?>[] fields = schema.getFields();
+                final Object[] values = new Object[fields.length];
+                for (int i = 0; i < fields.length; i++) {
+                    values[i] = fields[i].transform(result);
+                }
+
+                try {
+                    result = clazz.getDeclaredConstructor(componentTypes).newInstance(values);
```
snuyanzin, line 46: it's better to cache it as well to avoid constant calculation logic

panilya: I don't it's possible to cache object, but I can cache constructor. WDYT?

snuyanzin: yes, constructor should be enough

panilya: @snuyanzin I have a question for you, imagine an article about Datafaker 2.0, what would you expect to see in that article, I mean topics or something like this

snuyanzin (approved):

src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
snuyanzin: depends on the target audience...

there is a list of topics missed in faker related talks

1. custom providers 
2. there are ways to generate sequences (stream and sequence) 
3. the transformation approach (csv, json, sql, java, custom)
4. object population as a replacement for builders

panilya: Could I send you my draft of the article about Datafaker 2.0? I would like to hear your opinion if you don't mind and you have time, of course

snuyanzin: yep, sure

bodiam:
Nice feature Illia, amazing work!

PR #834, opened 2023-05-15
chore: add support for Azure Spring Apps

bowen5:
Add support for Azure Spring Apps (https://azure.microsoft.com/en-au/products/spring-apps/)

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Azure.java
+++ b/src/main/java/net/datafaker/providers/base/Azure.java
@@ -142,7 +142,7 @@ public class Azure extends AbstractProvider<BaseProviders> {
     }
     
     public String springApps() {
-        return "sa-" = randHex();
+        return "sa-" + randHex();
     }
 
     private String randHex() {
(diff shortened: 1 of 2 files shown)


snuyanzin (changes requested):
Could you please also add a test to
https://github.com/datafaker-net/datafaker/blob/main/src/test/java/net/datafaker/providers/base/AzureTest.java

bowen5:
> Could you please also add a test to https://github.com/datafaker-net/datafaker/blob/main/src/test/java/net/datafaker/providers/base/AzureTest.java

sure thing. added.

snuyanzin (approved):

snuyanzin:
Thanks for th contribution @bowen5

snuyanzin:
there is no precise calendar
however AFAIK @bodiam is working on a new version (please correct me if I'm wrong)

bodiam:
Well, not actively working on it at this moment, but the idea is to "soon" release 2.x

PR #837, opened 2023-05-15
Make FakerRepeatabilityIntegrationTest running in same thread

snuyanzin:


kingthorin:
Is this instead of the other PR or in addition?

snuyanzin:
addition to fix repeatability

bodiam:
I don't see a reason for this. This code worked fine before `d9c2cc262d29deb3552c9f49ab71edf0b3f6dc30`.

snuyanzin:
how do you that exactly d9c2cc262d29deb3552c9f49ab71edf0b3f6dc30 is the reason?
after that there were about 10 more commits to main branch and no such issue

bodiam:
So, that's interesting: if we enabled the `springApps()` method in Azure, things started failing.

bodiam:
> i wonder if it's because of `springApps()` or just because of any 1 more method which means that if we still have enabled `springApps()` however remove any another method will it work?

I don't think it's because of `springApps` per se, it's not doing much, but I think it's related to this:

```
private static final Random SHARED_RANDOM = new Random()
```

I've done a few random tests, and the funny things is:

```
    public String springApps() {
        String hex = faker.random().hex(4, false); // works
        return "sa-" + hex;
    }
```

This consistently works.

```
    public String springApps() {
        String hex = faker.random().hex(8, false); // works
        return "sa-" + hex;
    }
```

but this:

```
    public String springApps() {
        String hex = faker.random().hex(9, false); // 9 fails. Anything above fails. 8 = 1 byte, 9 = 2 bytes?
        return "sa-" + hex;
    }
```

bodiam:
> well I still suggest to make repeatabale tests in same thread.

Okay, that's fine!

PR #838, opened 2023-05-20
Speed up Number#numberBetween for int values

snuyanzin:
The test to check 
```java
    @Test
    void test() {
        for (int i = 0; i < 100_000_000; i++) {
            faker.number().numberBetween(-1000_000_000, 2000_000_000);
        }
    }
```
before the change it takes about 20x times longer to complete ...

the main idea is in case of overloading the limit use fallback to `long` instead of fallback to `BigDecimal`


bodiam:
Do the final parameters make any difference here?

snuyanzin:
right now no difference
the only reason I made it it is  kind of self-protection from possible reassigning values in methods in future.

Issue #840, opened 2023-05-22
faker.internet().url() creates the hostname only without the scheme


BeniaminK:
**Describe the bug**
faker.internet().url() creates the hostname only without the scheme

**To Reproduce**
```
faker.internet().url()
```
return `www.sherita-mcglynn.io`

**Expected behavior**
```
faker.internet().url()
```
return `https://www.sherita-mcglynn.io`

**Versions:**
 - OS: any
 - JDK 19
 - Faker Version: 1.9.0

**Additional context**
No context - I think there should be a method that crates the real URL, not just a hostname

kingthorin:
Agreed, seems reasonable to me.

bodiam:
It's a bit of a breaking change, but I think you're right too. Maybe we can keep the behaviour of the old method by renaming it, and create a new one which takes the http method into account. Should we only create https schemes, or should we take into account other schemes? (we have a method for random schemes too).

Also, we could consider adding a `uri` method next to the `url` method perhaps?

kingthorin:
PR coming.

PR #841, opened 2023-05-23
Internet provider: re-implement url, add webdomain method

kingthorin:
- Internet.java > Re-implement url. Have it leverage the new webdomain method (which contains the original url impl). The new url implementation creates a fully formed URL with random elements.
- InternetTest.java > Updated to test the new url and webdomain methods.

Fixes datafaker-net/datafaker#840

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Internet.java
+++ b/src/main/java/net/datafaker/providers/base/Internet.java
@@ -19,6 +19,7 @@ import java.util.regex.Pattern;
 public class Internet extends AbstractProvider<BaseProviders> {
     private static final Pattern SINGLE_QUOTE = Pattern.compile("'");
     private static final Pattern COLON = Pattern.compile(":");
+    private static final List<String> HTTP_SCHEMES = List.of("http://", "https://");
 
     protected Internet(BaseProviders faker) {
         super(faker);
@@ -66,7 +67,49 @@ public class Internet extends AbstractProvider<BaseProviders> {
         return resolve("internet.domain_suffix");
     }
 
+    /**
+     * Returns a string representing a web URL, randomly including: http/https scheme, port, path
+     * elements (2 or none), file element (1 or none), params (2 or none), anchor (1 or none).
+     * 
+     * @return a web URL
+     * @since 2.0.0
+     */
     public String url() {
+       final byte[] bts = faker.random().nextRandomBytes(6);
+        return url(bts[0] % 2 == 0, bts[1] % 2 == 0,
+                bts[2] % 2 == 0, bts[3] % 2 == 0,
+                bts[4] % 2 == 0, bts[5] % 2 == 0);
+    }
+
+    /**
+     * Returns a string representing a web URL, with various elements controlled by the caller.
+     * 
+     * @param schemeChoice if true will be random http or https, if false will be https
+     * @param portChoice if true a random port will be included, if false no port will be included
+     * @param pathChoice if true two random path elements will be included, if false no path elements will be included
+     * @param fileChoice if true the path will end with a random word element instead of a slash, if false it will end with a slash
+     * @param paramsChoice if true two random name value pairs will be included, if false no params will be included
+     * @param anchorChoice if true a random anchor will be included, if false no anchor will be included
+     * @return a web URL
+     * @since 2.0.0
+     */
+    public String url(boolean schemeChoice, boolean portChoice, boolean pathChoice, boolean fileChoice, boolean paramsChoice, boolean anchorChoice) {
+        String scheme = schemeChoice ? HTTP_SCHEMES.get(faker.random().nextInt(0, 1)) : "https://";
+        String port = portChoice ? (":" + port()) : "";
+        String path = pathChoice ? ("/" + slug(faker.lorem().words(2), "/")) : "/";
+        String file = fileChoice ? faker.lorem().words(1).get(0) : "";
+        String params = paramsChoice ? ("?" + slug(faker.lorem().words(2), "=") + "&" + slug(faker.lorem().words(2), "=")) : "";
+        String anchor = anchorChoice ? ("#" + faker.lorem().words(1).get(0)) : "";
+        return scheme + webdomain() + port + path + file + params + anchor;
+    }
+
+    /**
+     * Returns a web domain.
+     * 
+     * @return a web domain in the form "www.example.com"
+     * @since 2.0.0
+     */
+    public String webdomain() {
         return String.join("",
             "www",
             ".",
--- a/src/test/java/net/datafaker/providers/base/InternetTest.java
+++ b/src/test/java/net/datafaker/providers/base/InternetTest.java
@@ -8,7 +8,8 @@ import org.junit.jupiter.api.Test;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.List;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.Locale;
 import java.util.Random;
 import java.util.stream.Collectors;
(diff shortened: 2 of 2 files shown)


src/main/java/net/datafaker/providers/base/Internet.java
```diff
@@ -66,7 +66,36 @@ public String domainSuffix() {
...
+    }
+
+    /**
+     * Returns a string representing a URL, in the form "https://www.example.com/lorem/ipsun".
+     * 
+     * @return a URL, in the form "https://www.example.com/lorem/ipsun"
+     * @since 2.0.0
+     */
+    public String uri() {
+        return "https://" + hostname() + "/" + slug(faker.lorem().words(2), "/");
```
snuyanzin, line 89: since it's only `hhtps` shouldn't it be called httpsurl?
there are could be others e.g. `http`, `https`, `ftp`, `mailto`, `file`, `data` and `irc`

based on https://en.wikipedia.org/wiki/URL.wikipedia.org/wiki/Internet_Relay_Chat#URI_scheme)

kingthorin: And `tel` and `ws` and `wss` nearly un-ending….

I guess the question is do we want to maximize technical correctness or usability? Arguably we could also include port randomly. Username and password for http/https/ftp, etc.

I went with what would seem obvious to users (from my perspective). I figured assuming https in 2023 was fine, but I guess I could randomly use with or without the s 🤷‍♂️

bodiam: I wouldn't remove URL. An URL is different from a URI, would be good to keep them both.

src/main/java/net/datafaker/providers/base/Internet.java
```diff
@@ -66,7 +66,36 @@ public String domainSuffix() {
...
+    }
+
+    /**
+     * Returns a web domain.
+     * 
+     * @return a web domain in the form "www.example.com"
+     * @since 2.0.0
+     */
+    public String hostname() {
```
bodiam, line 98: A hostname can be anything, AFAIK. Eg "mars" could be a hostname, it's not something which has to start with www

src/main/java/net/datafaker/providers/base/Internet.java
kingthorin: It wasn’t removed. I agree they’re different things. I’ll drop the scheme to make the uri method more correct, and randomly include creds, port, params, and anchor as well as renaming it to weburi I guess. Or maybe I do include scheme randomly as well and call it weburl 🤷‍♂️

snuyanzin: i have an idea how it could be done (may be in a separate PR)
1. there could be an enum with all supported urls e.g.
```java
enum UrlType {
  HTTP('http', new HttpUrlGenerator()),
  HTTPS('https', new HttpsUrlGenerator()),,
  FTP('ftp', new FtpUrlGenerator()),,
  ...
}
```
for which there could be some rules specified.
then there could be methods
`url(UrlType type)` which generates based on requested type
and `url()` which is same as `url(Option.option(UrlTyps.class))` which means that first it picks a random type and then runs url generation for that type

kingthorin: I do like that idea, but I think it's more effort than anticipated for this PR :wink:

snuyanzin: that's why i mentioned that (may be in a separate PR)

kingthorin: Yup, I don't seem to be getting myself across clearly today. I just wanted you to know I was in agreement, but that I wouldn't be tackling it here. (I appreciate that you suggested or acknowledged it up front.)

bodiam: :) @kingthorin you were both clear, and you were both in agreement: simple change now, maybe more extended change later.

bodiam: Thanks btw for looking into this @kingthorin , nice work!

src/main/java/net/datafaker/providers/base/Internet.java
bodiam: Yeah, sorry, I should have suggested something I guess. Maybe it's good to have a look at the RFCs, and see what is the terminology for url/uri/urn/hostname/etc

I don't mind having a hostname method, but I'm not sure what it should return:

localhost 
hello.localhost
example.com
www.example.com
demo.example.com
http://example.com
https://www.example.com

??

kingthorin: Okay so for the purposes of this PR I'm going to go with web domain and have it return www.domain.tld :grinning:

Other stuff can be handled/added later if needed.

bodiam: Sounds fair!

src/main/java/net/datafaker/providers/base/Internet.java
```diff
@@ -66,7 +67,64 @@ public String domainSuffix() {
...
+     * elements (2 or none), file element (1 or none), params (2 or none), anchor (1 or none).
+     * 
+     * @return a web URL
+     * @since 2.0.0
+     */
+    public String weburl() {
+        return weburl(faker.random().nextBoolean(), faker.random().nextBoolean(),
+                faker.random().nextBoolean(), faker.random().nextBoolean(),
+                faker.random().nextBoolean(), faker.random().nextBoolean(),
+                faker.random().nextBoolean());
```
snuyanzin, line 94: what about
```suggestion
       final byte[] bts = faker.random().nextRandomBytes(7);
        return weburl(bts[0] % 2 == 0, bts[1] % 2 == 0,
                bts[2] % 2 == 0, bts[3] % 2 == 0,
                bts[4] % 2 == 0, bts[5] % 2 == 0,
                bts[6] % 2 == 0);
```
it will be faster since there is only one call to `random`

kingthorin: Sure, I'm good with that. I hadn't put any thought into the performance.

kingthorin:
@bodiam happy with the current implementation?

bodiam:
Thanks for your work here. It's not entirely what I expected / had in mind. I don't want to deprecate the url() method, I rather have it return sensible urls, and the same for urn().

The current methods seem not like great replacements, I don't think I've ever seen a username and password in a web url (in ftp urls perhaps), and I don't think they would provide useful generators like this.

kingthorin:
I don't mind dropping the credential part, no big deal.

As for completely re-working URL I think that's a mistake. I mean I guess we are going to 2.0.0 so a breaking change is fine (semver wise), but we need to be sure this is a "headline" in the release notes if we go that way.

bodiam:
I don't think returning a url from a url method is that unexpected. But launching 2.0 with already deprecated methods in it, that's not ideal.

kingthorin:
Okay. I'll trim the cred part and rename `weburl()` -> `url()`

kingthorin:
@bodiam Done? :grinning:

src/main/java/net/datafaker/providers/base/Internet.java
```diff
@@ -66,7 +67,49 @@ public String domainSuffix() {
...
+    /**
+     * Returns a string representing a web URL, randomly including: http/https scheme, port, path
+     * elements (2 or none), file element (1 or none), params (2 or none), anchor (1 or none).
+     * 
+     * @return a web URL
+     * @since 2.0.0
+     */
     public String url() {
+       final byte[] bts = faker.random().nextRandomBytes(6);
+        return url(bts[0] % 2 == 0, bts[1] % 2 == 0,
```
bodiam, line 79: tiny formatting thing here, but all good.

bodiam:
LGTM!

PR #856, opened 2023-06-11
Updated 2.0.0 documentation.

bodiam:


kingthorin:
Add a “Breaking Change” heading in 2.0.0.md and mention that internet.url() now actually returns a URL not just a domain name?

bodiam:
Good point. Mind doing it?

kingthorin:
No problem, it’ll be a few hours though.

PR #871, opened 2023-06-20
Chore: Remove duplicate 'import java util.List'

dbmalkovsky:
The result of migrating code from Java 8 to Java 17 syntax improvements.


kingthorin:
I'm surprised spotless didn't catch this.

PR #874, opened 2023-06-20
Chore: various code cleanups

dbmalkovsky:
- Remove type specification from new array creation; not needed
- Fix missorted modifiers
- Use `isEmpty()` instead of comparing to the length; `isEmpty()` is the form most used in the code currently

Have run `mvn clean test` without error after the changes

### DIFF
--- a/src/main/java/net/datafaker/sequence/FakeSequence.java
+++ b/src/main/java/net/datafaker/sequence/FakeSequence.java
@@ -34,7 +34,7 @@ public abstract class FakeSequence<T> implements Iterable<T> {
         return null;
     }
 
-    public static abstract class Builder<T> {
+    public abstract static class Builder<T> {
         protected final List<Supplier<T>> suppliers;
         protected int minLength = -1;
         protected int maxLength = -1;
--- a/src/main/java/net/datafaker/idnumbers/EsMXIdNumber.java
+++ b/src/main/java/net/datafaker/idnumbers/EsMXIdNumber.java
@@ -11,7 +11,7 @@ import java.util.List;
  */
 public class EsMXIdNumber implements IdNumbers {
 
-    private static final String[] CHA = new String[]{
+    private static final String[] CHA = {
         "HEFA560427MVZRRL04",
         "DKEM193827HDQWEF05",
         "KAKS142444HNSSFAW6",
--- a/src/main/java/net/datafaker/idnumbers/NricNumber.java
+++ b/src/main/java/net/datafaker/idnumbers/NricNumber.java
@@ -29,7 +29,7 @@ public class NricNumber implements IdNumbers {
             }
         }
 
-    private static final int[] CODE = new int[]{0, 2, 7, 6, 5, 4, 3, 2};
+    private static final int[] CODE = {0, 2, 7, 6, 5, 4, 3, 2};
     private static final String FIN_LETTERS = "XWUTRQPNMLK";
     private static final String UIN_LETTERS = "JZIHGFEDCBA";
     private static final Map<Type, NricType> INITIALIZER = new EnumMap<>(Type.class);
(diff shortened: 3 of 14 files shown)


src/main/java/net/datafaker/sequence/FakeSequence.java
```diff
@@ -34,7 +34,7 @@ public T singleton() {
         return null;
     }
 
-    public static abstract class Builder<T> {
+    public abstract static class Builder<T> {
```
snuyanzin, line 37: why?

dbmalkovsky: To match the ordering of all other modifiers in the project.

kingthorin (approved):

src/main/java/net/datafaker/sequence/FakeSequence.java
snuyanzin: I didn't get it
here you proposed the opposite ordering
https://github.com/datafaker-net/datafaker/pull/874/files#diff-87ea11ea16a9ecf7545c61f2ea1f7ed0d98278075881a2359529da3db54683b6R46

dbmalkovsky: I can't find the original Sun/Oracle document but this document is based on it; http://yohanan.org/steve/projects/java-code-conventions#SECTION00074000000000000000

which says:

6.4 ORDERING OF MODIFIERS
Though Java allows modifiers in any order, a consistent ordering improves readability of code. The following is the proper order of modifiers for declarations.

First any access modifiers (public, protected, or private) followed by (**in order, if present**) abstract, static, transient, volatile, synchronized, final, and native.

snuyanzin: Ok, thanks for clarification

then there is another question: wouldn't it be better to set up a ruleset with automation of checks and fix all violations rather than constantly having someone coming with (hope same) set of rules and fixing it?

kingthorin: Yes, but we could also say something is better than nothing 😉

bodiam:
Just curious : how did you find these improvements? Is that coming from a tool like Checkstyle or PMD, or did you use a different approach here?

dbmalkovsky:
I am using IntelliJ Analyze function for many of them but also PMD and SonarLint.

kingthorin:
I use SonarLint on a few projects, I'm sure I've used for stuff on this one. Some of its findings can be noisy, but it's tunable.

src/main/java/net/datafaker/sequence/FakeSequence.java
snuyanzin: i'm about +0 to this change meaning:
1 . currently i'm not going to block it, so you can merge it in case you think it is helpful
2. however from my point of view it doesn't solve the main problem: it doesn't prevent same violations in future

bodiam: I don't mind it

snuyanzin: then at some point the whole repo will be full of commits changing nothing but the order of modifiers...

bodiam: Wouldn't that be a bit of a False Dilemma? 

This if the first time this happens, if someone in the future wants to rearrange things alphabetically, we can always say no.

Having said that, I'm also happy to reject this PR, I see zero value in reordering keywords.

Okay: @dbmalkovsky , please revert the reordering of keywords, there's no significant value in it, and I don't want to enforce the sequence of these keywords in the future.

dbmalkovsky: Done

dbmalkovsky:
@kingthorin Sorry to have touched off a hornets nest of comments.

Issue #883, opened 2023-07-14
Memory leak in PROVIDERS_MAP


lukelast:
**Describe the bug**
Each instance of `Faker()` created will permanently consume memory. If you create enough instances you'll OOM.

**To Reproduce**
Code to reproduce is here:
https://github.com/lukelast/datafaker-bug-repro/blob/main/src/main/kotlin/Main.kt

The summary of the code is to create a lot of instances, let them be GC'd, then check heap usage.

**Expected behavior**
I expect heap usage to go back to the original size as everything is GC'd away.

**Versions:**
 - OS: Windows, Linux
 - JDK: 17, 20
 - Faker Version: 2.0.1

**Additional context**
![image](https://github.com/datafaker-net/datafaker/assets/1649361/7545fbbb-451a-453b-804c-f1a456ed3458)
![MicrosoftTeams-image](https://github.com/datafaker-net/datafaker/assets/1649361/5cfb55f1-7dcb-413f-80e5-2751033e4d61)

snuyanzin:
thanks for reporting
the problem waws in absence of hashcode/equals
there is a fix available for that

PR #885, opened 2023-07-14
Add hashcode and equals to RandomService

snuyanzin:
The absence of hashcode/equals was the reason of map growing at #883

### DIFF
--- a/src/main/java/net/datafaker/service/RandomService.java
+++ b/src/main/java/net/datafaker/service/RandomService.java
@@ -1,5 +1,6 @@
 package net.datafaker.service;
 
+import java.util.Objects;
 import java.util.Random;
 
 public class RandomService {
@@ -104,5 +105,19 @@ public class RandomService {
 
     public Random getRandomInternal() {
         return random;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof RandomService that)) return false;
+
+        return Objects.equals(random, that.random);
+    }
+
+    @Override
+    public int hashCode() {
+        if (random == SHARED_RANDOM) return 1;
+        return random != null ? random.hashCode() : 0;
     }
 }
--- a/src/test/java/net/datafaker/FakerTest.java
+++ b/src/test/java/net/datafaker/FakerTest.java
@@ -318,7 +318,7 @@ class FakerTest extends AbstractFakerTest {
         for (int i = 0; i < 10_000_000; i++) {
             Faker f = new Faker();
             String s = f.ancient().god();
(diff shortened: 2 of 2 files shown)


src/test/java/net/datafaker/FakerTest.java
```diff
@@ -310,6 +312,18 @@ void differentLocalesTest() {
         }
     }
 
+    @Test
+    @Timeout(value = 100_000, unit = TimeUnit.MILLISECONDS)
```
snuyanzin, line 316: before the fix this test was able to reproduce OOM on my laptop
it takes about 3s on my laptop in case it is executed alone and about 10 s in multithreaded junit5 test run
100 s just for safety

src/main/java/net/datafaker/service/RandomService.java
```diff
@@ -105,4 +106,18 @@ public String hex(int length, boolean upper) {
...
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof RandomService that)) return false;
+
+        return Objects.equals(random, that.random);
+    }
+
+    @Override
+    public int hashCode() {
+        if (random == SHARED_RANDOM) return 1;
```
snuyanzin, line 120: no need for extra computation in this case

src/test/java/net/datafaker/FakerTest.java
kingthorin: Why not actually use 100 seconds?

src/test/java/net/datafaker/FakerTest.java
```diff
@@ -310,6 +312,18 @@ void differentLocalesTest() {
...
+    void issue883Test() throws InterruptedException {
+        for (int i = 0; i < 10_000_000; i++) {
+            Faker f = new Faker();
+            String s = f.ancient().god();
+            if (i % 1000_000 == 0) {
```
kingthorin, line 321:
```suggestion
            if (i % 1_000_000 == 0) {
```

src/test/java/net/datafaker/FakerTest.java
snuyanzin: seconds is default timeunit in junit5 and as a result IntellijIdea tends to remove it when seconds are specified as a unit...
At the same time no unit means seconds in junit5 and millis in junit4... 

i think we can put just 1 min here to simplify it and to not allow Intellijidea introducing confusion

kingthorin: Sure that should still be lots.

kingthorin (approved):

kingthorin:
Thanks

Issue #888, opened 2023-07-15
Random generator from java Enums


vadeg:
**Is your feature request related to a problem? Please describe.**

I am often need to create test data using random values from a specific `enum`.

**Describe the solution you'd like**
Extend `RandomService` with a function which can return random `enum` value from a provided enum

**Describe alternatives you've considered**
Write generator for every project

**Additional context**
-

snuyanzin:
Did you have a look at `Options.option`?

bodiam:
We already have supprt for this in the Option class, didn't that work for you?

vadeg:
Yes. It is hard to find it. Closing the ticket. Thank for your response.

snuyanzin:
yes, sometimes documentation is weird
you can contribute to documentation by the way to make it more clear

bodiam:
An overview of the basic data generators (strings, ints, lists, enums, etc) night be very handy if this is missing from the documentation, and contributions would be very welcome.

kingthorin:
I’m re-opening this one to review/tackle docs after vacation too 😉

kingthorin:
I’m thinking this https://github.com/datafaker-net/datafaker/blob/main/docs/documentation/schemas.md#java-object-transformation is a good place to add some content on Enums and Options.option. Sound good?

bodiam:
Hmm, I don't think Enums have much overlap with the schema part, I think it would be more easily findable at the basic usage section, since all you have to do is do faker.options().option(Enum.class)?

kingthorin:
Okay. Thanks.

PR #889, opened 2023-07-15
Add a random generator for `enum` values (#888)

vadeg:
Add a random generator from a provided `enum` class

Fixes: #888


snuyanzin:
There is already existing `Options.option` which can do that.
Did you have a look at it?

snuyanzin (changes requested):
Why do we need  another implementation of this?

vadeg:
I haven't found it in documentation and it is not there. I will close PR

PR #895, opened 2023-07-19
Extract FakerRepeatabilityIntegrationTest to integration test phase

snuyanzin:
This should allow to avoid impact of other tests on it and should fix 
such flaky issue https://github.com/datafaker-net/datafaker/actions/runs/5596562572/jobs/10234110845#step:4:1039

### DIFF
--- a/src/test/java/net/datafaker/FakerTest.java
+++ b/src/test/java/net/datafaker/FakerTest.java
@@ -313,7 +313,7 @@ class FakerTest extends AbstractFakerTest {
     }
 
     @Test
-    @Timeout(value = 1, unit = TimeUnit.MINUTES)
+    @Timeout(value = 3, unit = TimeUnit.MINUTES)
     void issue883Test() throws InterruptedException {
         for (int i = 0; i < 10_000_000; i++) {
             Faker f = new Faker();
--- a/src/test/java/net/datafaker/integration/FakerRepeatabilityIntegrationTest.java
+++ b/src/test/java/net/datafaker/integration/FakerRepeatabilityIntegrationTest.java
@@ -54,7 +54,7 @@ public class FakerRepeatabilityIntegrationTest {
         assertThat(difference.entriesDiffering()).hasSizeGreaterThan(difference.entriesInCommon().size());
     }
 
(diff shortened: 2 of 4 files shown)


src/test/java/net/datafaker/FakerTest.java
```diff
@@ -313,7 +313,7 @@ void differentLocalesTest() {
     }
 
     @Test
-    @Timeout(value = 1, unit = TimeUnit.MINUTES)
+    @Timeout(value = 10, unit = TimeUnit.MINUTES)
```
kingthorin, line 316: I’d suggest 2, the failures I’ve seen have been decimals between 60 and 61 sec.

snuyanzin: there was one failure with 160 sec (and it exceeds 2 mins)
https://github.com/datafaker-net/datafaker/actions/runs/5597303576/jobs/10235552889#step:4:939

that's why i increased it

kingthorin: Of course just to prove me wrong one just hit 65.25sec…..but 2m should still be fine.

snuyanzin: it seems for macos it takes longer (no idea why), as i mentioned there was a case where it took almost 3min (so far just once)....
we can set 3...
anyway it is just a timeout which means the max duration of the task which usually ends faster

kingthorin: Yeah, I was just concerned that with 10 someone might misinterpret in the future. I’m probably overthinking it.

bodiam: I don't think you are overthinking, I was confused by this too, I thought the 1M generations took 9-10 minutes or so.

snuyanzin: i changed it to 3 min, let's see if macos will be happy...

btw the test is NOT about 1M, it's about 10M =)

kingthorin (approved):
LGTM, thanks!

src/test/java/net/datafaker/FakerTest.java
bodiam: 10M!! That's ambitious :)

PR #919, opened 2023-08-11
Speed up Number#randomNumber

snuyanzin:
Use integer `pow` instead of `Math.pow`

This brings some speed since no need to deal with fractions

before 
```
DatafakerSimpleMethods.numberBetween  thrpt   10  32624.208 ± 873.033  ops/ms
```
after
```
DatafakerSimpleMethods.numberBetween  thrpt   10  94411.143 ± 6192.659  ops/ms
```


kingthorin (approved):

kingthorin:
Thanks!

Issue #918, opened 2023-08-11
Kotlin example for JavaObjectTransformer does not work


zoobert:
**Describe the bug**
Using the example code provided in the [Java Object transformation](https://www.datafaker.net/documentation/schemas/#yaml-transformation) for Kotlin produces a `NoSuchMethodException`.

**To Reproduce**
1. Copy the Kotlin from the section for both the `Person` data class and the transformation
2. Run the test 
3. Boom!

**Expected behavior**
A successful run using the code copied from the documentation.

**Versions:**
 - OS: OSX
 - JDK 19
 - Faker Version 2.0.1

**Additional context**
This appears to be an incompatibility between how Java objects are constructed and what is available for a Kotlin data class. The call to `getDeclaredConstructor()` does not succeed, so a new instance is not created. I don't know how this would work for a Kotlin data class where I believe it has a single constructor where the values are required. Converting this to a non-data class or one of the methods suggested in [this](https://www.baeldung.com/kotlin/instantiate-data-class-empty-constructor) article seems to work. However, the code in the documentation as is, does not work. 

```
java.lang.RuntimeException: java.lang.NoSuchMethodException: com.modios.taxsolution.calc.kafka.Person.<init>()
	at net.datafaker.transformations.JavaObjectTransformer.apply(JavaObjectTransformer.java:69)
...
Caused by: java.lang.NoSuchMethodException: com.modios.taxsolution.calc.kafka.Person.<init>()
	at java.base/java.lang.Class.getConstructor0(Class.java:3641)
	at java.base/java.lang.Class.getDeclaredConstructor(Class.java:2810)
	at net.datafaker.transformations.JavaObjectTransformer.apply(JavaObjectTransformer.java:65)
```

snuyanzin:
Could you please share the code snippet to reproduce the issue with aformentioned `com.modios.taxsolution.calc.kafka.Person`?

zoobert:
Sure. It looks exactly like the one from the tutorial:

```
    data class Person(
        var firstName: String,
        var lastName: String,
        var birthDate: Date,
        var id: Int
    )
```

I did not extract the problem into its own repository since it was basically what was in the tutorial. I'm using Kotlin 1.8 and targeting Java 1.19.

bodiam:
Hi @zoobert , thanks for reporting this. Could you give the latest milestone a try (2.0.2-SNAPSHOT)? I've just implemented a fix for this.

bodiam:
[...]

PR #920, opened 2023-08-12
Added support for Kotlin Data classes.

bodiam:
Should fix https://github.com/datafaker-net/datafaker/issues/918

### DIFF
--- a/src/test/kotlin/net/datafaker/example/SchemaExampleTest.kt
+++ b/src/test/kotlin/net/datafaker/example/SchemaExampleTest.kt
@@ -1,0 +1,42 @@
+package net.datafaker.example
+
+import net.datafaker.Faker
+import net.datafaker.transformations.Field
+import net.datafaker.transformations.JavaObjectTransformer
+import net.datafaker.transformations.Schema
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import java.util.*
+import java.util.function.Supplier
+
+class SchemaExampleTest {
+
+    val faker = Faker()
+
+    @Test
+    fun `should support kotlin data classes`() {
+
+        val jTransformer = JavaObjectTransformer()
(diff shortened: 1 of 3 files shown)


src/test/kotlin/net/datafaker/example/SchemaExampleTest.kt
```diff
@@ -0,0 +1,42 @@
+package net.datafaker.example
+
+import net.datafaker.Faker
+import net.datafaker.transformations.Field
+import net.datafaker.transformations.JavaObjectTransformer
+import net.datafaker.transformations.Schema
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import java.util.*
```
kingthorin, line 9: Use specific imports?

kingthorin (commented):
That's the only thing I notice but I'm also not up on maven build or kotlin code.

src/test/kotlin/net/datafaker/example/SchemaExampleTest.kt
bodiam: Why?

kingthorin: I'm not overly picky about it. Might be a spotless thing, I'd have to double check. I just figured that'd be more consistent, but honestly I didn't check the existing code base to see.

bodiam: Well, it's not me who brought it up :-) I have really no idea, I just use IntelliJ defaults, and that's it.

kingthorin: That's fair, if that's how IntelliJ defaults then it likely is like that elsewhere in the code base. We can close this.

bodiam:
I am not sure if this is a complete fix, it might still break if you have multiple constructors etc, but it does pass the test now.

PR #921, opened 2023-08-12
Start of reusable AssertJ assertions.

bodiam:
This is just a POC, but sometimes I just want to know if a value is a zipcode, a firstname, etc. It would be nice to have a bunch of assertions which can do this. I'm not sure if it should be like: `isFullname()` or `hasPrefix()`, but I think option 2 is more composable, like:

```
assertThat(name).hasPrefix().hasMiddleName().has....
```

Ideas opinions welcome.

### DIFF
--- a/src/test/java/net/datafaker/assertj/base/NameAssert.java
+++ b/src/test/java/net/datafaker/assertj/base/NameAssert.java
@@ -1,0 +1,39 @@
+package net.datafaker.assertj.base;
+
+import org.assertj.core.api.StringAssert;
+
+public class NameAssert extends StringAssert {
+
+    public NameAssert(String s) {
+        super(s);
+    }
+
+    public static NameAssert asserThat(String actual) {
+        return new NameAssert(actual);
+    }
+
+    public NameAssert isName() {
+        isNotNull();
+
+        // check condition
+        if (!actual.matches("([\\w']+\\.?( )?){2,4}")) {
+            failWithMessage("Expected string to be a firstname, but it wasn't");
+        }
+
+        // return the current assertion for method chaining
+        return this;
+    }
+
+    public NameAssert isNameWithMiddleName() {
+        isNotNull();
+
+        // check condition
+        String regex = "^[A-Za-z.']+(?:\\s[A-Za-z.']+){2,}$";
+        if (!actual.matches(regex)) {
+            failWithMessage("Expected name with middle name to match <%s>, but was <%s>", regex, actual);
+        }
+
+        // return the current assertion for method chaining
+        return this;
+    }
+}
--- a/src/test/java/net/datafaker/assertj/DatafakerAssertions.java
+++ b/src/test/java/net/datafaker/assertj/DatafakerAssertions.java
@@ -1,0 +1,11 @@
+package net.datafaker.assertj;
+
+import net.datafaker.assertj.base.NameAssert;
+import org.assertj.core.api.Assertions;
+
+public class DatafakerAssertions extends Assertions {
+
+    public static NameAssert assertThat(String actual) {
+        return new NameAssert(actual);
+    }
+}
--- a/src/test/java/net/datafaker/providers/base/NameTest.java
+++ b/src/test/java/net/datafaker/providers/base/NameTest.java
@@ -1,10 +1,11 @@
 package net.datafaker.providers.base;
 
+import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
 
 import java.util.Locale;
 
-import static org.assertj.core.api.Assertions.assertThat;
+import static net.datafaker.assertj.DatafakerAssertions.assertThat;
 
 class NameTest extends BaseFakerTest<BaseFaker> {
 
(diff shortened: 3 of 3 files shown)


src/test/java/net/datafaker/assertj/base/NameAssert.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public static NameAssert asserThat(String actual) {
+        return new NameAssert(actual);
+    }
+
+    public NameAssert isName() {
+        isNotNull();
+
+        // check condition
+        if (!actual.matches("([\\w']+\\.?( )?){2,4}")) {
```
kingthorin, line 19: If the groups aren't reused then may as well make them non-capturing.

bodiam: You're probably right, but that's not really the purpose of this PR.

snuyanzin:
There are 3 ways how data could be generated
1. The list of available data is in file, based on random number on element is taken from the list
2. The list is in file, however sometimes the taken element is not a final value rather an expression which should be evaluated, so the result is the result of this expression
3. Generation is done without file usage

For the first approach there should be used `TestSpec`, the main advantage of `TestSpec` is that it reads the corresponding file and checks if a generated value is contained in the file or not, thus it is more pedantic approach. 
It could be updated to support the second point

Why couldn't we upgrade `TestSpec` to support other cases or build something based on it? It will minimize maintainers efforts

bodiam:
The purpose of this PR is to create a set of reusable assertions based on the data in Datafaker, but beyond the Datafaker project. As such, how the data is generated, by a regex or by a yml file, isn't of interest to me, but what is, is that I can use the assertions from Datafaker in other projects, and things like: 

`assertThat(input).isZipcode()` or `assertThat(quote).isShakespeare()`, or maybe better:

```
assertThat(phoneNumber)
   .hasInternationalDialingCode()
   .isAustralian()
```

snuyanzin:
Ok, thanks for explanation

then it's not clear how accurate should/could it be?
e.g. if there is a check for middle name is any regex e.g. like from this PR `^[A-Za-z.']+(?:\\s[A-Za-z.']+){2,}$`
then it assumes it works only for english alphabet... Even if a language uses latin symbols it will not work because of some additions like German, Polish, Czech and so on...

Also with regex it there will be plenty of false positives e.g. for something like `null null null` ... So either we have and maintain a kind database with real names and whatever we are going to provide or we should explicitely mentioned a number of limitations

bodiam:
It will never be 100% accurate I guess. It will be more like "is this value a city: yes.". Could the same value also be an author or the name of a movie: maybe also yes. Is that value a zipcode? Probably not. 

So yes, there's a lot of values for which we probably can't exactly say what it is. Some things can be solved by passing in a locale, like:

AssertThat(value).isZipcode(Locale.US)

For something like null null null, you probably shouldn't use this, but I don't encounter  results like that often.

snuyanzin:
I didn't get it
>Some things can be solved by passing in a locale, like:
>AssertThat(value).isZipcode(Locale.US)

a couple of comments before I suggested to go with `TestSpec` approach. You might be not care about how the data is generated which you are going to check however it's possible to use exact match (based on what we have in files for locales) instead of regexp.

Otherwise I don't understand how you are going to support `.isZipcode(Locale.US)`, `.isZipcode(Locale.ES)`, `.isZipcode(Locale.JP)` and others... Will there be a regexp per every locale, reuse of datafaker's functionality (the it's `TestSpec`) or something else

>For something like null null null, you probably shouldn't use this, but I don't encounter results like that often.

If we are talking about some external usage then we never know agaist what input it will applied.

btw to get string `null null null` is quite easy
```java
String somMethod() {
String s1 = null;
String s2 = null;
String s3 = null;
// some logic
return s1 + " " + s2 + " " + s3;
}
```
And now if there is a bug with assignment of values to `s1`, `s2`, `s3` then it will return `null null null`

bodiam:
How can TestSpec be used with AssertJ?

snuyanzin:
Not `TestSpec` itself however it's approach.
There is `net.datafaker.providers.base.BaseFakerTest#testProviderList` which retrieves all the values for the current key and locale. And then the only thing which should be done is checking whether a candidate string is in that collection or not. It could be incapsulated into under the hood of custom assert

>btw to get string null null null is quite easy
>I don't see your point.

it was an example of false positive when wrong assignment of variables or even absent of assignments (e.g. because of bug) could lead to wrong string e.g. `null null null` which then will be treated as a valid name with middle name

bodiam:
> Not `TestSpec` itself however it's approach. There is `net.datafaker.providers.base.BaseFakerTest#testProviderList` which retrieves all the values for the current key and locale. And then the only thing which should be done is checking whether a candidate string is in that collection or not. It could be incapsulated into under the hood of custom assert

That could work. I'm not so concerned about the actual implementation, I'm more interested in the caller's api and the AssertJ integration, and using the same functionality within Datafaker.

> 
> > btw to get string null null null is quite easy
> > I don't see your point.
> 
> it was an example of false positive when wrong assignment of variables or even absent of assignments (e.g. because of bug) could lead to wrong string e.g. `null null null` which then will be treated as a valid name with middle name

There will be many false positives. Something which can be a first name can maybe also be a last name or a city. And I'm not sure if null null null would be a valid full name, I think not, since it doesn't start with an upper case, so it's unlikely it's a full name, but that depends on the implementation and how accurate it should be.

I think it's more a "can value be an X", as in : 

Given the string "David":

can this string be a full name?  True 
can this string be a movie title? Probably also true 
But can string X be a US zipcode? False

bodiam:
> the accuracy is one of the most important things from my point of view to get more users of that feature. And I fear that we can not provide it on a cutting edge level with just regexps or resource files. As an example there is `libphonenumber` lib which we use in tests to answer a question whether it is a valid phone or not... Besides phone there are multiple other things...

All good, I'm still going to build it. I just typed in:

```
assertThat(mailbox.getEmailAddress()).isEmailAddress()
```

This doesn't compile, but this is something Datafaker could provide.

bodiam:
I'm going to close this for now. It's tricky to have multiple assertion groups, since they all work on strings, and this doesn't make the code nice. Would still be nice if we can reuse, but I don't see a good way forward for now.

PR #923, opened 2023-08-13
Speed up sql transformer

snuyanzin:
In some cases it's possible to replace `StringBuilder` with `StringJoiner`
The problem of `StringBuilder#append` is that each time it checks for arrays size even if it makes adding of just one symbol... it consumes lots of time

before
```
Benchmark                  Mode  Cnt    Score   Error   Units
DatafakerCsvSql.sql       thrpt   10  203.876 ± 2.975  ops/ms
```

after
```
Benchmark                  Mode  Cnt    Score   Error   Units
DatafakerCsvSql.sql       thrpt   10  232.296 ± 3.526  ops/ms
```

### DIFF
--- a/src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
+++ b/src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
@@ -92,20 +92,16 @@ public class SqlTransformer<IN> implements Transformer<IN, CharSequence> {
                 return SqlDialect.getFirstRow(
                         dialect, () -> appendTableInfo(fields), () -> addValues(input, fields), keywordCase);
             } else {
-                return "," + LINE_SEPARATOR
-                        + SqlDialect.getOtherRow(
-                        dialect, () -> appendTableInfo(fields), () -> addValues(input, fields), keywordCase);
+                return String.join(LINE_SEPARATOR, ",",
+                        SqlDialect.getOtherRow(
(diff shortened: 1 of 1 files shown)


bodiam:
Amazing. How do you find these things?

src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
```diff
@@ -87,69 +87,59 @@ public CharSequence apply(IN input, Schema<IN, ?> schema, int rowId) {
...
+                    + " "
+                    + appendTableInfo(fields)
+                    + " "
+                    + VALUES.getValue(keywordCase)
+                    + " "
+                    + addValues(input, fields);
         }
 
-        return sb.toString();
     }
```
kingthorin, line 109: I think in this particular part using the String builder might actually be better vs concatenation.

Maybe the original instantiation should specify a larger initial capacity (IIRC default is 16 char).

snuyanzin: yes, you're right
there is another way: use of `String.join` which does not have the issue of rebuilding an array...
however there is an issue of vararg... 
anyway sith `String.join` it is still faster than the initial version

snuyanzin:
>How do you find these things?
with generation of several millions lines lots of invisible things become visible...

kingthorin (approved):
LGTM

PR #932, opened 2023-08-21
Speed up cnpj

snuyanzin:
The idea is pretty straightforward: minimize amount of `substring` and regexps


src/main/java/net/datafaker/idnumbers/pt/br/DocumentFormatterUtil.java
```diff
@@ -2,6 +2,8 @@
 
 public final class DocumentFormatterUtil {
 
+    public static final String D_REGEX = "\\D+";
```
kingthorin, line 5: This seems unused

snuyanzin: thanks, sems like forget  to remove it

kingthorin (approved):

kingthorin:
Thanks!

PR #934, opened 2023-08-23
Extract SimpleFileVisitor, replace isDirectory with isRegularFile and put this condition to the end

snuyanzin:
I noticed that `Files.isDirectory` depends on the system and for me it takes a lot since it involves io operations...

The idea is to apply string operations (`endswith` and others first and only after io )

also extract creation of visitor from loop's iteration since it is always same

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Locality.java
+++ b/src/main/java/net/datafaker/providers/base/Locality.java
@@ -55,33 +55,35 @@ public class Locality extends AbstractProvider<BaseProviders> {
         Set<String> locales = new HashSet<>();
         final SimpleFileVisitor<Path> visitor = new SimpleFileVisitor<>() {
 
-            @Override
-            public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
-                final String filename;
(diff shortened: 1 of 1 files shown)


src/main/java/net/datafaker/providers/base/Locality.java
```diff
@@ -53,52 +53,49 @@ public List<String> allSupportedLocales(Set<String> fileMasks) {
...
-                            if (langs.contains(parentFileName)) {
-                                locales.add(parentFileName);
-                            } else {
-                                locales.add(filename.substring(0, filename.indexOf('.')));
+        final SimpleFileVisitor<Path> visitor = new SimpleFileVisitor<>() {
+
+            @Override
+            public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+                final String filename;
+                if (((filename = file.getFileName().toString().toLowerCase(Locale.ROOT)).endsWith(".yml") || filename.endsWith(".yaml")) && Files.isRegularFile(file) && Files.isReadable(file)) {
```
kingthorin, line 61: Move the assignment from the condition to the declaration? (I think that might just simplify how it reads for future maintenance).

snuyanzin: makes sense

kingthorin:
Thanks!

PR #935, opened 2023-08-23
Make TireTest more stable

snuyanzin:
It seems prefix could contain 3 digits as well

like here https://github.com/datafaker-net/datafaker/actions/runs/5948774981/job/16133232260?pr=934#step:4:971


kingthorin (approved):
Thank you!

PR #937, opened 2023-08-25
Optimize Twitter#twitterId

snuyanzin:
Several optimizations like extraction of logger to `static final`, use of `DateTimeFormatter` instead of `SimpleDateFormat`, usage of `StringBuilder`

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Twitter.java
+++ b/src/main/java/net/datafaker/providers/base/Twitter.java
@@ -102,7 +102,6 @@ public class Twitter extends AbstractProvider<BaseProviders> {
      */
     public String text(String[] keywords, int sentenceMaxLength, int wordMaxLength) {
         if (wordMaxLength <= 2) {
-            LOGGER.setLevel(Level.WARNING);
             LOGGER.warning("Word length less than 2 is dangerous. Exceptions can be raised.");
         }
         ArrayList<String> text = new ArrayList<>();
@@ -150,7 +149,6 @@ public class Twitter extends AbstractProvider<BaseProviders> {
      */
     public String getLink(String username, int extraLength) {
         if (extraLength <= 4) {
-            LOGGER.setLevel(Level.WARNING);
             LOGGER.warning("Extra length <=4 can cause collision.");
         }
         RandomService random = faker.random();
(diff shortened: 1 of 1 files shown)


src/main/java/net/datafaker/providers/base/Twitter.java
```diff
@@ -15,7 +17,9 @@
  * @since 0.9.0
  */
 public class Twitter extends AbstractProvider<BaseProviders> {
-
+    private static final Logger LOGGER = Logger.getLogger(Twitter.class.getName());
```
kingthorin, line 20: Full disclosure I only have experience with Log4j. Is a string needed here or is passing class fine?

src/main/java/net/datafaker/providers/base/Twitter.java
```diff
@@ -51,9 +56,8 @@ public Date createdTime(boolean forward, Date base, Date constrains) {
      */
     public String twitterId(int expectedLength) {
         if (expectedLength <= 6 || expectedLength >= 25) {
-            Logger logger = Logger.getLogger(Twitter.class.getName());
-            logger.setLevel(Level.WARNING);
-            logger.warning("expectedLength <= 6 may easily cause twitter id collision. And expectedLength >= 25" +
+            LOGGER.setLevel(Level.WARNING);
```
kingthorin, line 59: Does the level really have to be set on each use?

kingthorin: I see it below as well

src/main/java/net/datafaker/providers/base/Twitter.java
snuyanzin: In case of `java.util.logging.Logger` it requires `String`. To be sure it is unique usually people provide classnames

kingthorin (approved):

src/main/java/net/datafaker/providers/base/Twitter.java
snuyanzin: good question...
I have no idea, I've just faced this code and don't know why it ishould log like that...

Let's assume that at least `WARN` is always switched on and remove setting of the log level

PR #938, opened 2023-08-25
Minimize amount of substrings

snuyanzin:


src/main/java/net/datafaker/providers/base/Lorem.java
```diff
@@ -168,11 +168,11 @@ public String maxLengthSentence(final int fixedLength) {
...
 
-        String sentence = this.sentence(fixedLength);
-        String endOfSentence = sentence.substring(fixedLength - 1, fixedLength);
-        while (" ".equals(endOfSentence)) {
-            sentence = this.sentence(fixedLength);
-            endOfSentence = sentence.substring(fixedLength - 1, fixedLength);
+        String sentence = sentence(fixedLength);
+        final String space = " ";
+        while (space.regionMatches(
+            space.length(), sentence, fixedLength - 1, fixedLength)) {
```
kingthorin, line 174: Why have it process the length of space when it's known to be 1?

snuyanzin: yes, probably checking just last character would be better

kingthorin (approved):

PR #939, opened 2023-08-26
Use DateTimeFormatter instead of SimpleDateFormat

snuyanzin:
In `net.datafaker.service.FakeValuesServiceTest` in tests every time it creates `SimpleDateFormat` instance... It is more effectively to use `DateTimeFormatter` which could be created once and reused

### DIFF
--- a/src/test/java/net/datafaker/service/FakeValuesServiceTest.java
+++ b/src/test/java/net/datafaker/service/FakeValuesServiceTest.java
@@ -26,7 +26,6 @@ import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.temporal.ChronoField;
 import java.util.*;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -40,8 +39,6 @@ import static org.mockito.Mockito.when;
 
 class FakeValuesServiceTest extends AbstractFakerTest {
 
-    private static final Long SECONDS_IN_5_HOURS = TimeUnit.HOURS.toSeconds(5);
-    private static final Long SECONDS_IN_10_DAYs = TimeUnit.DAYS.toSeconds(10);
     private static final DateTimeFormatter DATE_TIME_FORMATTER =
         new DateTimeFormatterBuilder().appendPattern("yyyy-MM-dd HH:mm:ss")
             .appendFraction(ChronoField.MILLI_OF_SECOND, 0, 3, true).toFormatter();
@@ -306,7 +303,7 @@ class FakeValuesServiceTest extends AbstractFakerTest {
     @RepeatedTest(100)
     void futureDateExpression() {
         LocalDateTime now = LocalDateTime.now(OffsetDateTime.now(ZoneId.systemDefault()).getOffset());
-        LocalDateTime nowPlus10Days = now.plusSeconds(SECONDS_IN_10_DAYs);
+        LocalDateTime nowPlus10Days = now.plusDays(10);
         String expression = fakeValuesService.expression("#{date.future '10','TimeUnit.DAYS'}", faker, context);
         LocalDateTime date = LocalDateTime.parse(expression, DATE_TIME_FORMATTER);
         assertThat(date).isStrictlyBetween(now, nowPlus10Days);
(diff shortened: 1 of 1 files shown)


src/test/java/net/datafaker/service/FakeValuesServiceTest.java
```diff
@@ -35,8 +39,11 @@
 
 class FakeValuesServiceTest extends AbstractFakerTest {
 
-    private static final Long MILLIS_IN_AN_HOUR = 1000 * 60 * 60L;
-    private static final Long MILLIS_IN_A_DAY = MILLIS_IN_AN_HOUR * 24;
+    private static final Long SECONDS_IN_AN_HOUR = 60 * 60L;
+    private static final Long SECONDS_IN_A_DAY = SECONDS_IN_AN_HOUR * 24;
```
kingthorin, line 43: Are there not java constants for this in ChromoUnit or TimeUnit? Or couldn't it be simplified to like `TimeUnit.HOURS.toSeconds(1)` and `TimeUnit.DAYS.toSeconds(1)`

snuyanzin: good point, thanks
now `TimeUnit` is used for that

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
```diff
@@ -35,8 +40,11 @@
 
 class FakeValuesServiceTest extends AbstractFakerTest {
 
-    private static final Long MILLIS_IN_AN_HOUR = 1000 * 60 * 60L;
-    private static final Long MILLIS_IN_A_DAY = MILLIS_IN_AN_HOUR * 24;
+    private static final Long SECONDS_IN_AN_5_HOURS = TimeUnit.HOURS.toSeconds(1) * 5;
+    private static final Long SECONDS_IN_A_10_DAYs = TimeUnit.DAYS.toSeconds(1) * 10;
```
kingthorin, line 44: The "A" and "AN" can be dropped

kingthorin: And the 5 and 10 can replace the 1

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
```diff
@@ -35,8 +40,11 @@
 
 class FakeValuesServiceTest extends AbstractFakerTest {
 
-    private static final Long MILLIS_IN_AN_HOUR = 1000 * 60 * 60L;
-    private static final Long MILLIS_IN_A_DAY = MILLIS_IN_AN_HOUR * 24;
+    private static final Long SECONDS_IN_5_HOURS = TimeUnit.HOURS.toSeconds(5);
+    private static final Long SECONDS_IN_10_DAYs = TimeUnit.DAYS.toSeconds(10);
```
bodiam, line 44: was the lowercase `s` her intentional?

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
```diff
@@ -296,32 +304,27 @@ void expressionWithValidFakerObjectValidMethodInvalidArgs() {
...
-
-        Date now = new Date();
-        Date nowPlus10Days = new Date(now.getTime() + MILLIS_IN_A_DAY * 10);
-
-        Date date = dateFormat.parse(fakeValuesService.expression("#{date.future '10','TimeUnit.DAYS'}", faker, context));
-
-        assertThat(date.getTime()).isStrictlyBetween(now.getTime(), nowPlus10Days.getTime());
+    void futureDateExpression() {
+        LocalDateTime now = LocalDateTime.now(OffsetDateTime.now(ZoneId.systemDefault()).getOffset());
+        LocalDateTime nowPlus10Days = now.plusSeconds(SECONDS_IN_10_DAYs);
```
bodiam, line 309: why not do: 

`LocalDateTime nowPlus10Days = now.plusDays(10)` ?

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
snuyanzin: don't think so, it is just a private const, so seems just a misprint

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
snuyanzin: yea, that makes sense, probably another with hours should go same way
thanks

kingthorin (approved):

snuyanzin:
thanks for taking a look

PR #950, opened 2023-09-02
Speed validEsMXSsn by extraction vars

snuyanzin:
There is a number of vars like lists, regexps, datetimeformatter which could be extracted on constant level and as a result improve speed

### DIFF
--- a/src/main/java/net/datafaker/idnumbers/EsMXIdNumber.java
+++ b/src/main/java/net/datafaker/idnumbers/EsMXIdNumber.java
@@ -25,12 +25,12 @@ public class EsMXIdNumber implements IdNumbers {
         "QWDF414424HNSDVAW4",
         "AKDF144424MDEFVFA1"
     };
-    private static final String[] CONSONANT = {"B", "C", "C", "D", "F",
-        "G", "H", "J", "K", "L", "L", "M", "N", "N",
-        "P", "Q", "R", "S", "T", "V", "W", "X", "Y", "Z"
+    private static final char[] CONSONANT = {'B', 'C', 'C', 'D', 'F',
+        'G', 'H', 'J', 'K', 'L', 'L', 'M', 'N', 'N',
+        'P', 'Q', 'R', 'S', 'T', 'V', 'W', 'X', 'Y', 'Z'
     };
 
-    private static final String[] VOWEL = {"A", "E", "I", "O", "U"};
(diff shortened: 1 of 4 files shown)


src/test/java/net/datafaker/idnumbers/EsMXIdNumberTest.java
```diff
@@ -1,29 +1,31 @@
...
 import org.junit.jupiter.api.RepeatedTest;
 
 import java.util.Locale;
+import java.util.regex.Pattern;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 class EsMXIdNumberTest {
 
+    private static final Pattern REGEX = Pattern.compile("[A-Z][A-Z][A-Z][A-Z]\\d{6}[HM]" +
```
kingthorin, line 14: Why not simplify [A-Z]{4} and then 5 for the second set?

src/main/java/net/datafaker/idnumbers/EsMXIdNumber.java
```diff
@@ -47,26 +49,24 @@ public class EsMXIdNumber implements IdNumbers {
...
-        String v4 = faker.options().option(VOWEL);
-
-        String ranNum = (Integer.parseInt(birthDay.substring(0, 4)) <= 1999) ?
-            "0" : faker.options().option(CONSONANT);
-
-        String ssn = c1 + v1 + c2 + c3 + birthDay.substring(2, 8) + sex + state + v2 + v3 + v4 + ranNum;
-        ssn = ssn + getChecksum(ssn);
-
-        return ssn;
+        String ssn = options.option(CONSONANT) +
```
kingthorin, line 56: Would this chunk be any faster with a StringBuilder using append?

src/test/java/net/datafaker/idnumbers/EsMXIdNumberTest.java
snuyanzin: yes, that makes sense, thanks

src/main/java/net/datafaker/idnumbers/EsMXIdNumber.java
snuyanzin: It seems there is even faster way.
Since the length of the result is always same there could be used just array of chars without any concatenation

kingthorin (approved):
Thanks!

PR #951, opened 2023-09-03
Use refs to classes rather than hardcoded names

snuyanzin:
It's better to refer to classes and use their names/packages rather than hardcode them and try to not forget while refactoring

### DIFF
--- a/src/test/java/net/datafaker/script/ProviderGenerator.java
+++ b/src/test/java/net/datafaker/script/ProviderGenerator.java
@@ -1,7 +1,14 @@
 package net.datafaker.script;
 
+import net.datafaker.providers.base.AbstractProvider;
+import net.datafaker.providers.entertainment.EntertainmentFakerTest;
+import net.datafaker.providers.entertainment.EntertainmentProviders;
+import net.datafaker.providers.videogame.VideoGameFakerTest;
+import net.datafaker.providers.videogame.VideoGameProviders;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


src/test/java/net/datafaker/script/ProviderGenerator.java
```diff
@@ -77,14 +84,14 @@ private void createCreator(File file, String key, Set<String> strings, ProviderT
 
         System.out.println("package " + providerType.getPackageName() + ";");
         System.out.println();
-        System.out.println("import net.datafaker.providers.base.AbstractProvider;");
+        System.out.println("import " + AbstractProvider.class.getName() + ";");
         System.out.println();
         System.out.println("/**");
-        System.out.println(" * @since 1.8.0");
+        System.out.println(" * @since 2.0.2");
```
kingthorin, line 90: I guess the behaviour is slightly different but the functionality was still introduced in 1.8

kingthorin: Or is this something that's tweaked (or supposed to be) on every use?

snuyanzin: this is a generator of a boiler plate for new provider, so in theory it should contain actual version since it is invoke for actual version.
May be in future it will be better to use something like maven -properties plugin to put version from there or something like that to make it bumping automatically here

kingthorin (approved):
Thanks

PR #953, opened 2023-09-03
Speed up json generation by minimizing of Stringbuilder append invocations

snuyanzin:
The idea is to invoke `StringBuilder` for larger peaces than just one symbol if possible to minimize amount of checks while `StringBuilder#append"

### DIFF
--- a/src/main/java/net/datafaker/transformations/JsonTransformer.java
+++ b/src/main/java/net/datafaker/transformations/JsonTransformer.java
@@ -129,10 +129,11 @@ public class JsonTransformer<IN> implements Transformer<IN, Object> {
             final int length = val.length();
             for (int i = 0; i < length; i++) {
                 final char c = val.charAt(i);
-                if (ESCAPING_MAP.get(c) == null) {
+                final String escapedValue = ESCAPING_MAP.get(c);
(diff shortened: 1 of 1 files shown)


src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -125,9 +125,19 @@ private static void value2String(Object value, StringBuilder sb) {
...
                 sb.append("\"");
             }
-            for (char c : String.valueOf(value).toCharArray()) {
-                sb.append(ESCAPING_MAP.getOrDefault(c, String.valueOf(c)));
+            int start = 0;
+            final int length = val.length();
+            for (int i = 0; i < length; i++) {
+                final char c = val.charAt(i);
+                final String escapedValue = ESCAPING_MAP.get(c);
+                if (escapedValue == null) {
```
kingthorin, line 133: I know it means using negative logic but if you check not null then no other condition/branch is needed.

snuyanzin: makes sense, done

kingthorin (approved):

PR #960, opened 2023-09-08
Speed up EnZAIdNumber#validSsn

snuyanzin:
There is no need to transform int to string and then to parse it again...
Also amount of substrings could be decreased

### DIFF
--- a/src/main/java/net/datafaker/idnumbers/EnZAIdNumber.java
+++ b/src/main/java/net/datafaker/idnumbers/EnZAIdNumber.java
@@ -21,7 +21,6 @@ public class EnZAIdNumber implements IdNumbers {
      * @return a valid social security number on faker
      */
     public String getValidSsn(BaseProviders f) {
-
         String ssn = "";
         while (!validSsn(ssn)) {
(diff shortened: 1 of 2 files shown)


src/main/java/net/datafaker/idnumbers/EnZAIdNumber.java
```diff
@@ -75,9 +74,8 @@ boolean validSsn(String ssn) {
             return false;
         }
 
-        int calculatedChecksum = calculateChecksum(ssn.substring(0, 12));
-        int checksum = Integer.parseInt(ssn.substring(12, 13));
-        return (checksum == calculatedChecksum);
+        int calculatedChecksum = calculateChecksum(ssn, 12);
+        return ssn.charAt(12) - '0' == calculatedChecksum;
```
kingthorin, line 78: Is there any reason for the intermediate store (assignment)? Why not just combine the two lines?

snuyanzin: nothing special, could be combined
done

kingthorin (approved):

PR #961, opened 2023-09-10
Speed up Twitter#getLink

snuyanzin:


src/test/java/net/datafaker/providers/base/TwitterTest.java
```diff
@@ -91,29 +92,36 @@ void testTextKeyWords() {
 
     @Test
     void username() {
+        final Twitter twitter = faker.twitter();
```
kingthorin, line 95: Does this need to be instantiated for each method or should it just be a class member?

snuyanzin: yes, makes sense to extract it

kingthorin (approved):

PR #962, opened 2023-09-10
Speed up Lorem#sentence

snuyanzin:
The idea is to capitalize just one word to minimize substring.
Also no need to create list of strings, just concatenate them as soon as possible

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Lorem.java
+++ b/src/main/java/net/datafaker/providers/base/Lorem.java
@@ -116,7 +116,7 @@ public class Lorem extends AbstractProvider<BaseProviders> {
         for (int i = 1; i < totalWordCount; i++) {
             sb.append(" ").append(word());
         }
-        return sb + ".";
+        return sb.append(".").toString();
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/datafaker/providers/base/Lorem.java
```diff
@@ -107,7 +108,15 @@ public String sentence(int wordCount) {
...
-        return capitalize(String.join(" ", words(wordCount + numberOfWordsToAdd)) + ".");
+        final int totalWordCount = wordCount + numberOfWordsToAdd;
+        StringBuilder sb = new StringBuilder();
+        if (totalWordCount > 0) {
+            sb.append(WordUtils.capitalize(word()));
+        }
+        for (int i = 1; i < totalWordCount; i++) {
+            sb.append(" ").append(word());
+        }
+        return sb + ".";
```
kingthorin, line 119: Use `.append` again?

snuyanzin: done

kingthorin (approved):

PR #967, opened 2023-09-16
chore(deps): Downgrade codecov/codecov-action action to v3.1.4

kingthorin:
Revert https://github.com/datafaker-net/datafaker/pull/964 because it isn't working. I suspect they incorrectly/temporarily had a v4 tag.


kingthorin:
Side note: Is there a reason we aren't including 2.x?

```yaml
  push:
    branches: [ main, 1.x ]
  pull_request:
    branches: [ main, 1.x ]
```

PR #970, opened 2023-09-18
Flip equals to protect from possible NPE

snuyanzin:


bodiam:
Where's the test?

src/main/java/net/datafaker/service/FakeValuesContext.java
```diff
@@ -69,7 +69,7 @@ private static String getFilename(Locale locale) {
      * This addresses that unfortunate condition.
      */
     private static String language(Locale l) {
-        if (l.getLanguage().equals("iw")) {
+        if ("iw".equals(l.getLanguage())) {
```
bodiam, line 72: Can this really return a null?

src/main/java/net/datafaker/providers/videogame/Hearthstone.java
```diff
@@ -31,7 +31,7 @@ public int battlegroundsScore() {
 
     public String standardRank() {
         String rank = resolve("games.hearthstone.rank");
-        if (rank.equals("Legend")) {
+        if ("Legend".equals(rank)) {
```
bodiam, line 34: How can this return a null?

src/main/java/net/datafaker/providers/base/Aviation.java
```diff
@@ -110,7 +110,7 @@ public String engineType() {
      */
     public String flight(String type) {
         String airline;
-        if (type.equals("ICAO")) {
+        if ("ICAO".equals(type)) {
```
bodiam, line 113: Maybe a non-null annotation would be better?

src/main/java/net/datafaker/service/FakeValuesContext.java
snuyanzin: this one probably not
however imho it's better to have constant first to avoid situations when someone uses this code as example however don't think about NPE

src/main/java/net/datafaker/providers/videogame/Hearthstone.java
snuyanzin: this one probably not
however imho it's better to have constant first to avoid situations when someone uses this code as example however don't think about NPE

src/main/java/net/datafaker/providers/base/Aviation.java
snuyanzin: for this one probably both options are ok

at the same time there are still a number of other public methods taking args which potentially could be null

and there we might think about annotations or something like that... 
however the main thing is to make them tested and returning a human friendly message in case of problem

src/main/java/net/datafaker/service/FakeValuesContext.java
bodiam: Yeah, I don't agree, but that's fine. I rather be explicit about these things, instead of hiding the null behavior.

src/main/java/net/datafaker/providers/videogame/Hearthstone.java
snuyanzin: yes, however here there is the opposite case:
in code we have only defined case for some specific string value and everything else...
With such approach I would consider null value as everything else

snuyanzin:
Thanks for the feedback
I added tests to reproduce NPE for some of the cases
I'm going to merge it

PR #978, opened 2023-10-02
Use array of chars instead of StringBuilder for Company#domainName

snuyanzin:


src/test/java/net/datafaker/providers/base/CompanyTest.java
```diff
@@ -7,9 +7,12 @@
 
 import java.util.List;
 import java.util.Collection;
+import java.util.regex.Pattern;
 
 class CompanyTest extends BaseFakerTest<BaseFaker> {
 
+    public static final Pattern URL_PATTERN = Pattern.compile("(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])");
```
kingthorin, line 14: I know this is just a copy/paste but could you make it consistent in lower then caps for the letter sets?

snuyanzin: yes, that makes sense thanks

PR #980, opened 2023-10-02
Collection of populated elements should not be always same

### LINKED ISSUE #979: Java Object transformation to list
Is Java object transformation to list possible??

From the documentation here https://www.datafaker.net/documentation/schemas/#java-object-transformation it seems only singleton is posssible unless i missed something.

Please advise. Here is my iomplementation:

```java
Faker faker = new Faker();

                Field.field("name", () -> faker.name().firstName()),
                Field.field("email", () -> faker.internet().emailAddress())
        );

BaseFaker bfaker = new BaseFaker();
Client client = bfaker.populate(Client.class, schema); //Here i would like to have a collection of fake Clients instead
```

snuyanzin:
This allows to have different elements in collection of populated objects and should fix #979


bodiam:
Thanks for fixing!

PR #982, opened 2023-10-03
Build on Java 21 (temurin doesn't support Java 21 yet, so moved to Co…

bodiam:
…rretto for the time being)

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -16,7 +16,7 @@ jobs:
   build:
     strategy:
       matrix:
-        java-version: [ 17, 20 ]
+        java-version: [ 17, 21 ]
         runs-on: [ ubuntu-latest, macos-latest, windows-latest ]
     name: Jdk ${{ matrix.java-version }}, os ${{ matrix.runs-on }}
     runs-on: ${{ matrix.runs-on }}
@@ -26,7 +26,7 @@ jobs:
         uses: actions/setup-java@v3
         with:
           java-version: ${{ matrix.java-version }}
-          distribution: 'temurin'
+          distribution: 'corretto'
           cache: maven
       - name: Build with Maven
         run: ./mvnw $mvn_options verify -PnoGpg --file pom.xml


kingthorin (approved):

.github/workflows/maven.yml
```diff
@@ -10,7 +10,7 @@ on:
     branches: [ main, 1.x ]
 
 env:
-  mvn_options: --batch-mode -Dstyle.color=always -Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss.SSS
+  mvn_options: --batch-mode -Dnet.bytebuddy.experimental=true -Dstyle.color=always -Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss.SSS
```
snuyanzin, line 13: no need fo `=true`
this should be enough
`-Dnet.bytebuddy.experimental`

bodiam: I just stole it from here: https://github.com/raphw/byte-buddy/issues/1396

snuyanzin: +1

bodiam: Seems this Windows Build doesn't want to start or finish. I'll kill it

kingthorin: 👍

I re-ran failed (which includes cancelled). Fingers crossed it actually runs this time.

bodiam: haha, thanks @kingthorin !

bodiam: Hmm, it seems the Windows version doesn't like the build for some reason. Anyone suggestions?

kingthorin: I can look into it after work.

bodiam: I cancelled the build, triggered it again, but seems no luck.

snuyanzin: let's wait till the end to see exception
i guess there should be timeout something about an hour

snuyanzin: seems I found the reason

1. `mockito` (5.5.0)  depends on `byte-buddy` 1.14.6
2. `mvn dependency:tree` shows that it uses `byte-buddy` 1.12.11
3. it comes from `assertj-core`

As a result exclusion of `byte-buddy`  from `assertj-core` solves the issue and makes it using `byte-buddy` from `mockito`
as a bonus after that there is no need for `-Dnet.bytebuddy.experimental` at all

kingthorin: Nice! Thanks for tackling that.

kingthorin (approved):

PR #987, opened 2023-10-06
SvSe id number should contain valid date

### LINKED ISSUE #986: Faker v2.0.2 generates invalid Sweden SSN
Hi there 👋 

We've been using `net.datafaker.providers.base.IdNumber#validSvSeSsn` for quite some time but it stopped working with the `2.0.2` version. Our application resolves users' birth date from their Swedish SSN which, according to the SSN _specification_ should be always possible. Unfortunately, the faker started returning random numbers, for example, `295659+4036`. Can you refer to that? The method `validSvSeSsn` suggests that the number would follow the rules so maybe there is a regression bug in the `2.0.2`?

snuyanzin:
Also adds test to check whether it contains a valid date or not
should fix #986

### DIFF
--- a/src/main/java/net/datafaker/idnumbers/SvSEIdNumber.java
+++ b/src/main/java/net/datafaker/idnumbers/SvSEIdNumber.java
@@ -68,7 +68,7 @@ public class SvSEIdNumber implements IdNumbers {
                 }
             }
         }
-        return false;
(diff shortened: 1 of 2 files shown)


bodiam:
Thanks for your fix @snuyanzin !

PR #989, opened 2023-10-07
Use generation of birthday instead of loop

snuyanzin:
There is no need for repeatedly generation of new id until the lucky one...
Instead just generate the correct one with the first attempt

also allows to speed up generation as a result

kind of follow up for #987

### DIFF
--- a/src/main/java/net/datafaker/providers/base/DateAndTime.java
+++ b/src/main/java/net/datafaker/providers/base/DateAndTime.java
@@ -258,6 +258,7 @@ public class DateAndTime extends AbstractProvider<BaseProviders> {
      * @param minAge the minimal age
      * @param maxAge the maximal age
      * @return a random birthday between {@code minAge} and {@code maxAge} years ago from now.
+     * Negative {@code minAge} and {@code maxAge} are supported.
      */
     public Timestamp birthday(int minAge, int maxAge) {
         final LocalDate localDate = LocalDate.now();
@@ -278,6 +279,7 @@ public class DateAndTime extends AbstractProvider<BaseProviders> {
      * @param minAge the minimal age
      * @param maxAge the maximal age
      * @return a random birthday between {@code minAge} and {@code maxAge} years ago from now.
+     * Negative {@code minAge} and {@code maxAge} are supported.
      */
     public LocalDate birthdayLocalDate(int minAge, int maxAge) {
         final LocalDate localDate = LocalDate.now();
--- a/src/main/java/net/datafaker/providers/base/DateAndTime.java
+++ b/src/main/java/net/datafaker/providers/base/DateAndTime.java
@@ -234,6 +234,15 @@ public class DateAndTime extends AbstractProvider<BaseProviders> {
     }
 
     /**
+     * Generates a random birthday between 65 and 18 years ago from now as LocalDate.
+     *
+     * @return a random birthday between 65 and 18 years ago from now.
+     */
+    public LocalDate birthdayLocalDate() {
+        return birthdayLocalDate(DEFAULT_MIN_AGE, DEFAULT_MAX_AGE);
+    }
+
+    /**
      * Generates a string representation of a random birthday between 65 and 18 years ago from now.
      *
      * @param pattern date time pattern to convert to string.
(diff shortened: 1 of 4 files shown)


src/main/java/net/datafaker/providers/base/DateAndTime.java
```diff
@@ -264,6 +273,25 @@ public Timestamp birthday(int minAge, int maxAge) {
             LocalDateTime.of(date, LocalTime.ofNanoOfDay(faker.number().numberBetween(0, DAYS_NANOS))));
     }
 
+    /**
+     * Generates a random birthday between two ages from now as LocalDate.
+     *
+     * @param minAge the minimal age
+     * @param maxAge the maximal age
+     * @return a random birthday between {@code minAge} and {@code maxAge} years ago from now.
+     * @throws IllegalArgumentException if the {@code maxAge} is lower than {@code minAge}.
```
kingthorin, line 282: Does this need to try/catch/rethrow?

src/main/java/net/datafaker/providers/base/DateAndTime.java
```diff
@@ -264,6 +273,25 @@ public Timestamp birthday(int minAge, int maxAge) {
...
+     */
+    public LocalDate birthdayLocalDate(int minAge, int maxAge) {
+        final LocalDate localDate = LocalDate.now();
+        final LocalDate from = localDate.minusYears(maxAge);
+        if (minAge == maxAge) {
+            return from;
+        }
```
kingthorin, line 289: Do this before instantiating any other cars?

kingthorin: Should we check that neither value is negative?

src/main/java/net/datafaker/providers/base/DateAndTime.java
snuyanzin: good catch, seem it is outdated and not related anymore to this method

src/main/java/net/datafaker/providers/base/DateAndTime.java
snuyanzin: it works ok with negative numbers (will return birthday in future)

kingthorin: Maybe note that in the javadoc then?

kingthorin: Also fixed my original note here should have said “vars” instead of “cars” 🤦‍♂️

snuyanzin: added to javadoc

src/main/java/net/datafaker/providers/base/DateAndTime.java
```diff
@@ -264,6 +273,25 @@ public Timestamp birthday(int minAge, int maxAge) {
...
+     * @param maxAge the maximal age
+     * @return a random birthday between {@code minAge} and {@code maxAge} years ago from now.
+     * Negative {@code minAge} and {@code maxAge} are supported.
+     */
+    public LocalDate birthdayLocalDate(int minAge, int maxAge) {
+        final LocalDate localDate = LocalDate.now();
+        final LocalDate from = localDate.minusYears(maxAge);
+        if (minAge == maxAge) {
+            return from;
+        }
```
kingthorin, line 289: I'd still suggest moving this before the var declarations, no point creating them if they won't be used.

kingthorin: OMG disregard, you're returning one of the declared values 🤦‍♂️

kingthorin (approved):

src/main/java/net/datafaker/providers/base/DateAndTime.java
snuyanzin: I didn't get it
```java
final LocalDate localDate = LocalDate.now();
final LocalDate from = localDate.minusYears(maxAge);
if (minAge == maxAge) {
    return from;
}
```
the return value is based on the 2 values declared above since first need to know current date and need to substract the input age

kingthorin: It’s fine. I was being brain dead for a sec. Sorry.

snuyanzin: ok, no problem , thanks for taking a look

PR #994, opened 2023-10-17
Add enforcer plugin to get more human friendly error message for wrong jdk

snuyanzin:
Currently if anyone will try to build it with jdk 8
it will print a huge exception which requires time understand what is the issue...

with this approach it will just print 
```
[ERROR] Rule 0: org.apache.maven.enforcer.rules.version.RequireJavaVersion failed with message:
[ERROR] Detected JDK version 1.8.0-372 (JAVA_HOME=<...>/jdk8/jre) is not in the allowed range [17,).

```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,6 +37,7 @@
         <maven.compiler.source>17</maven.compiler.source>
         <maven.compiler.target>17</maven.compiler.target>
         <maven-compiler-plugin.version>3.11.0</maven-compiler-plugin.version>
+        <maven-enforcer-plugin.version>3.4.1</maven-enforcer-plugin.version>
         <maven-failsafe-plugin.version>3.1.2</maven-failsafe-plugin.version>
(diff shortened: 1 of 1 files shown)


kingthorin:
Are the mismatched brackets really the output from the plugin or was that a bad copy/paste/touchup?

snuyanzin:
`<...>` - these brackets are the result of my edits
`[17,)` - these brackets are the result of plugin's output where square bracket means including and round bracket excluding. Since for the left part there is 17 then it means starting from 17 including and  since there is nothing on the right part then it means till infinity. Both boumdaries could be configured in pom

kingthorin:
That's fine, just leave oks funny mixed square and round. For my two  cents `[17,]` makes it clear but whatever it's out of your control.

snuyanzin:
it could be squared if specified both ends IIRC

thanks for the feedback

PR #1008, opened 2023-11-02
Manually relocate snakeyaml multi-release classes.

### LINKED ISSUE #1007: Relocated snakeyaml package not correct in META-INF/versions
**Describe the bug**
The relocated snakeyaml package is not correctly transformed for a multi-release jar setup. The package path is still `META-INF/versions/9/org/yaml/snakeyaml/internal`  but the  `Logger.class` valid for Java version 9 and higher (inside this folder) uses `package net.datafaker.shaded.snakeyaml.internal` as package name. This is probably caused by the Maven shaded plugin but you might have a workaround to fix it. The not correctly shaded path leads to collisions with not shaded snakeyaml variants when executing multi-release jars.

**To Reproduce**
Open (unpack) the datafaker jar and find the `META-INF/versions/9` folder and the snakeyaml package.

**Expected behavior**
The multi-release variant package path shoud be correctly shaded `META-INF/versions/9/net/datafaker/shaded/snakeyaml/internal`

**Versions:**
 - JRE >= 9
 - Faker Version 2.0.2

schmist:
The multi-release classes for the relocated snakeyaml dependency need to be relocated manually due to a bug in the Maven Shade plugin. The plugin only relocates the content but not the package structure itself. See https://issues.apache.org/jira/browse/MSHADE-406. This fixes #1007.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -295,8 +295,9 @@
                              needed due to a bug in the Maven Shade plugin, which only relocates the content but not
                              the package structure itself, see https://issues.apache.org/jira/browse/MSHADE-406. -->
                         <relocation>
-                            <pattern>META-INF/versions/9/org.yaml.snakeyaml</pattern>
-                            <shadedPattern>META-INF/versions/9/net.datafaker.shaded.snakeyaml</shadedPattern>
+                            <pattern>META-INF/versions/(\d+)/org/yaml/snakeyaml</pattern>
+                            <shadedPattern>META-INF/versions/$1/net/datafaker/shaded/snakeyaml</shadedPattern>
+                            <rawString>true</rawString>
                         </relocation>
                         <relocation>
                             <pattern>com.mifmif.common.regex</pattern>
--- a/pom.xml
+++ b/pom.xml
@@ -291,6 +291,13 @@
                             <pattern>org.yaml.snakeyaml</pattern>
                             <shadedPattern>net.datafaker.shaded.snakeyaml</shadedPattern>
                         </relocation>
+                        <!-- Manually relocate shaded snakeyaml multi-release classes to the correct package. This is
+                             needed due to a bug in the Maven Shade plugin, which only relocates the content but not
+                             the package structure itself, see https://issues.apache.org/jira/browse/MSHADE-406. -->
+                        <relocation>
+                            <pattern>META-INF/versions/9/org.yaml.snakeyaml</pattern>
+                            <shadedPattern>META-INF/versions/9/net.datafaker.shaded.snakeyaml</shadedPattern>
+                        </relocation>
                         <relocation>
                             <pattern>com.mifmif.common.regex</pattern>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -291,6 +291,13 @@
                             <pattern>org.yaml.snakeyaml</pattern>
                             <shadedPattern>net.datafaker.shaded.snakeyaml</shadedPattern>
                         </relocation>
+                        <!-- Manually relocate shaded snakeyaml multi-release classes to the correct package. This is
+                             needed due to a bug in the Maven Shade plugin, which only relocates the content but not
+                             the package structure itself, see https://issues.apache.org/jira/browse/MSHADE-406. -->
+                        <relocation>
+                            <pattern>META-INF/versions/9/org.yaml.snakeyaml</pattern>
+                            <shadedPattern>META-INF/versions/9/net.datafaker.shaded.snakeyaml</shadedPattern>
```
snuyanzin, line 299: I'm curious whether there is a more generic solution which could handle other versions once they appear...
Otherwise any new version will lead to the same issue

schmist: There might be an (undocumented way) to do this, I am currently looking at the Maven plugin code.

schmist: The follwing commit should address your concern. Now all versions of the `org.yaml.snakeyaml` package are being relocated. Although new versions will not come around very often and I hope that the actual bug in the Maven plugin will get fixed sooner or later so that this workaround can be removed.

snuyanzin (commented):
I noticed that in shaded classes (e.g. Logger) there is a package name 
`package META-INF.versions.9.org.yaml.snakeyaml.internal;`
i.e. it contains `-` in `META-INF` which is not supported symbol in package names...
I wonder how it is supposed to work?

schmist:
> package META-INF.versions.9.org.yaml.snakeyaml.internal

Where do you see this? Inside the packaged jar?

snuyanzin:
yes, inside the packaged jar
yep, probably you are right 
however right now i'm not able to test end2end case.
Can you confirm that this change makes it working?

schmist:
It probably works because the relocation I added is only used to move the classes into the correct package. The package name inside the class files is not replaced since the provided relocation path (including META-INF) does not match. Since the buggy Maven Shade plugin has done the renaming already (but not the relocation into the correct package, see linked bug report), we should be on the save side. I agree that it is not a clean solution but the result is a working and correct multi-release jar. This workaround can be removed as soon as the bug in the Plugin is fixed.

I can also confirm that with this fix, the issue is gone for our use case and datafaker is still doing its job.

schmist:
It seems that they are currently working on a fix: https://github.com/apache/maven-shade-plugin/pull/202

snuyanzin:
thanks for checking
yes, probably it would make sense to remove this from pom file once there will be a new shade plugin version with the fix available

snuyanzin (approved):

PR #1026, opened 2023-11-24
feature: add joke provider

hardikSinghBehl:
This PR add a joke provider class that supports the following methods
* `new Faker().joke().pun();`
* `new Faker().joke().knockKnock();`

This can be extended in the future to support more joke categories.

Functionality has been validated and tested locally.


kingthorin (approved):
Thanks!

PR #1037, opened 2023-12-05
added module-info.java

### LINKED ISSUE #1020: Datafaker as a module
It would be great if you can add a module-info.java and make it available from modular projects.

pbankovic:
Support for modular projects, fixes #1020

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -362,6 +362,7 @@
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <version>${maven-javadoc-plugin.version}</version>
                 <configuration>
+					<sourcepath>src/main/java</sourcepath>
                     <detectJavaApiLink>false</detectJavaApiLink>
                     <additionalOptions>-Xdoclint:all -Xdoclint:-missing</additionalOptions>
                 </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -362,7 +362,7 @@
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <version>${maven-javadoc-plugin.version}</version>
                 <configuration>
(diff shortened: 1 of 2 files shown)


bodiam:
I understand why you did it, but should we really expose the internal package?

pbankovic:
Actually, we shouldn't. But after I saw that the build was failing, I thought not exposing the internal package caused it.
I'm gonna remove it again. 
Would yo say that net.datafaker.internal.helper is the only package that should stay internal, or is there anything else that shouldn't be referenced from outside?

bodiam:
It seems the build is still failing. I wouldn't worry too much about exposing the helper package right now, we can address that later, but I think it's indeed the only package which ideally should stay hidden

pbankovic:
Yes, I see that it fails to generate javadoc, but I just can't see how it could be affected by the module declaration file.

pom.xml
```diff
@@ -362,6 +362,7 @@
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <version>${maven-javadoc-plugin.version}</version>
                 <configuration>
+					<sourcepath>src/main/java</sourcepath>
```
kingthorin, line 365: Could you indent this to be 'inside' configuration?

pbankovic: Oh, yes, I guess my IDE is configured differently

pbankovic:
I had to add the source path to the javadoc plugin, and now it's fine.
Fixing the indention, and that should be it  😄

kingthorin (approved):

bodiam:
I know, I know. I need to refresh that key. Happy if you want to merge it though

kingthorin:
Okay, thanks!

PR #1045, opened 2023-12-22
Use String#replace whenever possible

snuyanzin:
If there is no need to replace by regexp then there is `String#replace` which is much faster than `replaceAll`, thus it's better to use it instead

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Internet.java
+++ b/src/main/java/net/datafaker/providers/base/Internet.java
@@ -41,7 +41,8 @@ public class Internet extends AbstractProvider<BaseProviders> {
     public String username() {
         StringBuilder result = new StringBuilder();
         final Name name = faker.name();
-        final String firstName = name.firstName().toLowerCase(faker.getContext().getLocale()) + "." + name.lastName().toLowerCase(faker.getContext().getLocale());
+        final String firstName = name.firstName().toLowerCase(faker.getContext().getLocale())
+            + "." + name.lastName().toLowerCase(faker.getContext().getLocale());
         for (int i = 0; i < firstName.length(); i++) {
             final char c = firstName.charAt(i);
             if (c == '\'' || Character.isWhitespace(c)) {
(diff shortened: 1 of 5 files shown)


src/main/java/net/datafaker/providers/base/Internet.java
```diff
@@ -91,7 +91,7 @@ public String domainName() {
     }
 
     public String domainWord() {
-        return FakerIDN.toASCII(SINGLE_QUOTE.matcher(faker.name().lastName().toLowerCase()).replaceAll(""));
+        return FakerIDN.toASCII(faker.name().lastName().toLowerCase().replace("'", ""));
```
kingthorin, line 94: Shouldn't lowercase here at least use Locale.ROOT?

src/main/java/net/datafaker/providers/base/Internet.java
```diff
@@ -145,7 +145,7 @@ public String webdomain() {
             "www",
             ".",
             FakerIDN.toASCII(
-                SINGLE_QUOTE.matcher(faker.name().firstName().toLowerCase()).replaceAll("") +
+                faker.name().firstName().toLowerCase().replace("'", "") +
```
kingthorin, line 148: Here too

src/main/java/net/datafaker/providers/base/Internet.java
snuyanzin: here it's name generated base on locale set by user, so in case it iis e.g. TR it should be being lower cased to turkish lowercase. Thus i think here it is ok to have locale from faker's context

kingthorin: Is it using fakers context when not passed though?

snuyanzin: yes, good catch, 
added locale from context
thanks

src/main/java/net/datafaker/providers/base/Internet.java
snuyanzin: also added locale from context

kingthorin (approved):

kingthorin:
Thanks!

PR #1048, opened 2023-12-27
Update for tr.yml

CengizhanAral:
I added more data for the tr.yml file.


bodiam:
Nice work! Thanks for this addition!

kingthorin:
This is great, thanks 🙂

PR #1062, opened 2024-01-19
Use bitwise operations

snuyanzin:
Replace some operations with bitwise operations since they in general a bit faster


kingthorin (approved):
LGTM and all tests remain passing 🙂

snuyanzin:
thanks for taking a look

PR #1066, opened 2024-01-24
Remove duplicates from yaml values

### LINKED ISSUE #1064: Lorem words are not unique
**Describe the bug**

There are several repeated words in `lorem.words` and `lorem.supplemental`, and given the unique provider doesn't check values but works with index, it returns the same value multiple times.

I can submit a PR just wanted to check in if there is any reason for the duplicates.
This can be solved by either removing duplicates or improving the unique provider to not return duplicates even if the yaml contains them. This latter approach is a bit more work but safer, more robust.

**To Reproduce**

```
faker.unique().fetchFromYaml("lorem.words")
faker.unique().fetchFromYaml("lorem.supplemental")
```

**Expected behavior**

Unique values are returned.

**Versions:**
 - Faker Version 2.1.0

TamasSzigeti:
Fixes #1064


kingthorin:
@TamasSzigeti would you be up for adding tests to this or is that asking too much? (Not a trick we can do it separately if that's outside your interest/skillset.)

TamasSzigeti:
Sure, gladly, will make time for it in the coming days.

kingthorin:
Even just something like:

Get the List from the yaml: `List<String> loremWords = getBaseList("lorem.words");`
Create a Set from the List (since Sets in Java are unique values): `Set<String> uniqueWords = new HashSet<>(loremWords);`

Then assert that the two sizes should be equal, then we know there's no duplicates.
Ummm actually you could even do a parameterized test that passes "lorem.words" and "lorem.supplemental" as key String, use "Terms" instead of "Words" for the variable names, then it's just a single test method like `wordListsShouldNotHaveDuplicates` or something like that.

kingthorin:
Your method content will be different but here's an example how to annotate and set things up:
https://github.com/datafaker-net/datafaker/blob/8166a0f2e2c63c8763bdf00b3d5b87bba0e27c80/src/test/java/net/datafaker/providers/base/DateAndTimeTest.java#L194-L199

TamasSzigeti:
@kingthorin I did a bit more than that, have a look pls

kingthorin (approved):
LGTM

kingthorin:
Thanks!

kingthorin:
@snuyanzin @bodiam one of you wanna make a second call on this. I'm not 100% sure all lists should be unique. (I don't see a reason why not, but don't wanna kill something being unaware.)

src/test/java/net/datafaker/providers/base/BaseFakerTest.java
```diff
@@ -90,6 +90,20 @@ protected void testProviderList(TestSpec testSpec, TestInfo testInfo) {
...
+    void testNoDuplications(TestSpec testSpec) {
+        if (testSpec.isDummy) {
+            return;
+        }
+
+        var terms = getBaseList(testSpec.key);
+
+        assertThat(new HashSet<>(terms).size())
+                .as("Check no duplications in " + testSpec.key)
+                .isEqualTo(terms.size());
```
snuyanzin, line 104:
```suggestion
               .hasSameSizeAs(terms);;
```
could be simplified

snuyanzin:
thanks for the contribution
i think there are a couple of thing need to be done
1. apply spotless to remove unused imports
2. there are several tests failing after this change, they should be adopted

kingthorin:
Thanks!

snuyanzin (approved):

bodiam:
@TamasSzigeti a snapshot release should have been published automatically, I can make a new official release soon if you need.

kingthorin:
https://s01.oss.sonatype.org/content/repositories/snapshots/net/datafaker/datafaker/

TamasSzigeti:
Thanks, I can use the snapshot for now, no problem

PR #1074, opened 2024-01-27
Update README - Code fence snapshot location

kingthorin:
So that users don’t try to visit it and get lost in the vast number of folders etc

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ dependencies {
 
 You can also use the snapshot version (`2.1.1-SNAPSHOT`), which automatically gets published
 after every push to the main branch of this repository. Binary repository URL for snapshots download is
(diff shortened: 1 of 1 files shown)


snuyanzin (approved):

bodiam:
I'm not sure if this change is ideal. To use snapshots in gradle, and afaik, also in maven, you need to  specify a url for snapshots. That was the original url. The new url is pointing to the Datafaker url specifically, but I'm not really sure what the benefit of that is, nobody would download the jar manually, or use this url for anything, so I believe the old url was perhaps better?

kingthorin:
Okay if it's meant to be a documentation thing instead of a destination for users how about we put code fence syntax on it?

Issue #1076, opened 2024-01-30
aws().vpcId() generates invalid id


axelfontaine:
**Describe the bug**
`aws().vpcId()` generates invalid id such as `vpc-8e6792c23ea07abf`

**To Reproduce**
`faker.aws().vpcId()`

**Expected behavior**
valid vpc id such as `vpc-08e6792c23ea07abf` (note the extra `0` !)

**Versions:**
 - OS: Win11 x64
 - JDK 21
 - Faker Version 2.0.

**Additional context**
same issue with `subnetId()` and `securityGroupId()`

kingthorin:
Is that a length requirement or a character requirement? (Meaning does it have to be a leading 0 or just something in the existing set)

axelfontaine:
Both

kingthorin:
Well either it has to be zero or not. It can’t really be both.

axelfontaine:
The length is wrong (16 hex chars instead of 17) and the first char has to be 0.

kingthorin:
There we go, that actually answers the question. Thank you.

kingthorin:
https://aws.amazon.com/blogs/compute/longer-resource-ids-in-2018-for-amazon-ec2-amazon-ebs-and-amazon-vpc/

kingthorin:
So is it literally 0 followed by 16hex or are they generated serially and the first character simply hasn’t rolled yet?

All the examples I see are zero but I couldn’t find anything definitive.

bodiam:
Seems like a reasonable straightforward fix, would you be willing to provide a fix @axelfontaine ?

bodiam:
I don't think they should start with a 0. If you look at the aws documentation, this is what it says:

https://aws.amazon.com/about-aws/whats-new/2018/02/longer-format-resource-ids-are-now-available-in-amazon-ec2/

The example used there is:

vpc-1234567890abcdef0

kingthorin:
Can anyone point to documentation that supports `subnetId()` and `securityGroupId()` being 17 as well?

axelfontaine:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/resource-ids.html

kingthorin:
Thank  you, I'll tweak my PR.

PR #1079, opened 2024-02-01
AWS vpc/subnet/securitygroup IDs now 17char

kingthorin:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/resource-ids.html

Fixes datafaker-net/datafaker#1076

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Aws.java
+++ b/src/main/java/net/datafaker/providers/base/Aws.java
@@ -59,22 +59,26 @@ public class Aws extends AbstractProvider<BaseProviders> {
     }
 
     public String securityGroupId() {
-        return "sg-" + randHex();
+        return "sg-" + randHex(17);
     }
 
     public String subnetId() {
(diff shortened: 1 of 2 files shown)


src/main/java/net/datafaker/providers/base/Aws.java
```diff
@@ -59,21 +59,25 @@ public String route53ZoneId() {
...
     private String appName() {
         return faker.app().name().toLowerCase().replaceAll("\\W+", "");
     }
 
+    private String randHex(int length) {
+        return faker.random().hex(length, false);
+    }
+
     private String randHex() {
         return faker.random().hex(16, false);
```
snuyanzin, line 82: I would suggest to replace then the implementation of `randHex` with something like
```java
private String randHex() {
        return randHex(16);
}        
```

kingthorin:
Tweaked

snuyanzin (approved):

PR #1083, opened 2024-02-05
Use indexof string

snuyanzin:
Suddenly I realized that `indexOf(<String>)` is faster than `indexOf(<char>)` since the first one has jvm intrinsic...

So the PR makes use `indexOf(<String>)` 

it gives some speed e.g. could be checked with this test
```java
   private static final Name name = new Faker().name();
    @Test
    void test() {
        for (int i = 0; i < 500_000_000; i++) {
            name.lastName();
        }
    }
```

for my machine before change it took 16-18 sec
after 9-10 sec (almost 2 times)

### DIFF
--- a/src/main/java/net/datafaker/annotations/FakeResolver.java
+++ b/src/main/java/net/datafaker/annotations/FakeResolver.java
@@ -56,7 +56,8 @@ public class FakeResolver<T> {
     private Schema<Object, T> getSchema(String pathToSchema) {
         if (pathToSchema != null) {
             try {
-                final int sharpIndex = pathToSchema.indexOf('#');
+                // indexOf(<String>) is faster than indexOf(<char>) since it has jvm intrinsic
+                final int sharpIndex = pathToSchema.indexOf("#");
                 final Class<?> classToCall;
                 final String methodName;
                 if (sharpIndex >= 0) {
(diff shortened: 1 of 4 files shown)


bodiam:
I love what you come up with :-) Approved of course!

snuyanzin:
thanks a lot!

bodiam:
I've executed the test myself btw, I made a small change to make sure the JIT didn't inline the code, to something like this:

```
    @Test
    void test() {
        for (int i = 0; i < 500_000_000; i++) {
            String s = name.lastName();

            if(s.length()> 1000) {
                System.out.println("xx");
            }
        }
    }
```

Before the change, executing it took 16 seconds (MacBook Pro M2, Corretto-17.0.10). After the change, it took 15.8 seconds :-)

snuyanzin:
interesting, i tested with openjdk 17
Also If I run with profiler before change I'm able to see `net.datafaker.service.FakeValuesService#resolveExpression(java.lang.String, java.lang.Object, net.datafaker.providers.base.ProviderRegistration, net.datafaker.service.FakerContext)` with `charAt` among consumers
since this code is moved now and in most of the cases (especially repeating) it should not be invoked, I assume that in your profiler there should be something else
Screenshot from 2024-02-05 12-43-37

PR #1092, opened 2024-02-22
Use RgxGen to cope with StackOverFlow

### LINKED ISSUE #1091: StackOverflowError for regexify
for some of the `regexify` it fails with StackOverflow e.g.

```java
new Faker().regexify("^arn:.+:.+:.*:([0-9]{12}):(.+)$");
```
This is an  issue in generex https://github.com/mifmif/Generex/issues/56
There was a suggestion to use an alternative e.g. https://github.com/curious-odd-man/RgxGen

snuyanzin:
fixes #1091

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
         <commons-lang3.version>3.14.0</commons-lang3.version>
         <commons-text.version>1.11.0</commons-text.version>
         <commons-validator.version>1.8.0</commons-validator.version>
-        <generex.version>1.0.2</generex.version>
+        <rgxgen.version>2.0</rgxgen.version>
         <google.javaformat.version>1.17.0</google.javaformat.version>
         <jacoco-maven-plugin.version>0.8.11</jacoco-maven-plugin.version>
         <jbanking.version>4.2.0</jbanking.version>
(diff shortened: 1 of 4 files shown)


kingthorin (approved):

bodiam:
I don't think we can depend on snapshot versions. If they publish a new version, we would have no idea what kind of code we get into our library. I don't think this would be an acceptable  situation.

snuyanzin:
based on the commit history they are not going to do it...
the latest commit was more than 8 months ago...

bodiam:
I don't really care how often they release, I care that we don't depend on snapshots for production releases. Maybe they change the api in a year, then it would break every Datafaker release depending on the snapshot, including older versions. If they would introduce something malicious, we would download this too automatically. So no, let's not do this.

kingthorin:
Wouldn't that be true of any dependency?

kingthorin (commented):
Assuming a release happens and this PR goes forward then I think `src/main/java/module-info.java` will also need updating.

snuyanzin:
it is already a a part of the PR or do you mean something else?

kingthorin:
All good, I totally overlooked it 🤦‍♂️

bodiam:
2.0.0 has been released!

snuyanzin:
thanks for letting know, updated to use released version

snuyanzin:
Merging since now there is no SNAPSHOT version in deps

PR #1106, opened 2024-03-02
Added pronouns provider

eliasnogueira:
## Description

This PR adds a new provider: Pronouns.

### Question 1

I don't know the release process, so I added this new provider as a 2.1.0 version.
Should we change it to 2.1.1, if it gets merged, or wait until the next release?

### Question 3

Should we add it as part of the `Name` provider?


src/main/resources/en/pronouns.yml
```diff
@@ -0,0 +1,39 @@
...
+        - "Her"
+        - "Him"
+        - "Them"
+        - "Hir"
+        - "Xem"
+        - "Vir"
+        - "Tem"
+        - "Em"
+      possessive:
+        - "Her"
```
kingthorin, line 23: Hers

src/main/resources/en/pronouns.yml
```diff
@@ -0,0 +1,39 @@
...
+        - "His"
+        - "Their"
```
kingthorin, line 25: Theirs

src/main/resources/en/pronouns.yml
eliasnogueira: Fixed

src/main/resources/en/pronouns.yml
eliasnogueira: Fixed

snuyanzin:
it seems it should be rebased since there are conflicts

eliasnogueira:
@snuyanzin should we start a discussion about the artifact release?

From my view, this PR will get merged but not be available until you guys release a new artifact.

Should the version increase automatically and be released when the PR is merged?
Should we wait for a formal release?

I'm curious about the release process, and also happy to help.

kingthorin:
On merge it'll be snapshot yes and available that way. Otherwise yes it'll have to wait for a release.

docs/releases/2.1.0-SNAPSHOT.md
kingthorin, line 1: This file shouldn't be changed. 2.1.0 was already released

docs/documentation/providers.md
```diff
@@ -197,6 +197,7 @@ Datafaker comes with a total of 233 data providers:
...
 | [Princess Bride](https://javadoc.io/doc/net.datafaker/datafaker/latest/net/datafaker/providers/entertainment/PrincessBride.html)                                |                                                                                                                                                                                                             | Entertainment | 0.8.0 |
 | [Programming Language](https://javadoc.io/doc/net.datafaker/datafaker/latest/net/datafaker/providers/base/ProgrammingLanguage.html)                             |                                                                                                                                                                                                             | Base          | 0.8.0 |
+| [Pronouns](https://javadoc.io/doc/net.datafaker/datafaker/latest/net/datafaker/providers/base/Pronouns.html)                                                    | Gender pronouns are the set of pronouns that an individual uses to reflect that person's own gender identity                                                                                                | Base          | 2.1.0 |
```
kingthorin, line 200: This shouldn't be modified this is part of the release tasks (IIRC)

src/main/java/net/datafaker/providers/base/Pronouns.java
```diff
@@ -0,0 +1,44 @@
+package net.datafaker.providers.base;
+
+/**
+ * This class generates gender neutral pronouns
+ * @since 2.1.0
```
kingthorin, line 5: I believe the next release is slated to be 2.2.0

kingthorin:
You could add it to the 'full list' in the readme.

docs/documentation/providers.md
bodiam: Correct, but it doesn't hurt.

docs/releases/2.1.0-SNAPSHOT.md
bodiam: Seems this change was maybe done on a outdated branch?

src/main/resources/en/pronouns.yml
```diff
@@ -0,0 +1,39 @@
+en:
+  faker:
+    pronouns:
+      subjective:
+        - "She"
+        - "He"
+        - "They"
```
bodiam, line 7: I'm wondering if these things should be uppercased?

src/main/java/net/datafaker/providers/base/Pronouns.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public class Pronouns extends AbstractProvider<BaseProviders> {
```
bodiam, line 7: Any reason why a new class is used, and not add it to Gender? Gender seems like a reasonable place for this, or not?

docs/releases/2.1.0-SNAPSHOT.md
eliasnogueira: That's right. I just pushed the merge conflict fix

docs/documentation/providers.md
eliasnogueira: That's one of the points about release. The code, doc, and version will be there without the artifact generated and publicly available.

I can follow any recommendation.

src/main/java/net/datafaker/providers/base/Pronouns.java
eliasnogueira: That's a good question.
I'm trying to follow the same pattern as the other " self-contained " providers. They have not implemented the fluent interface approach as I could call the class itself or another class.

Example:
```java
faker.gender().pronouns().subjective();
```

The example above would be implemented in this way:
* The `Pronouns` class still exists
* The `Gender` class would have a pronouns() method returning its new instance

```java
public class Gender extends AbstractProvider<BaseProviders> {

    // ignoring already existing code
    public Pronouns pronouns() {
        return new Pronouns(faker);
    }
}
```
* The `Pronouns` class as a default `toString()` in the case the consumer would call only its method
```java
faker.gender().pronouns()
```

I'm open to any implementation ideas :-)

src/main/java/net/datafaker/providers/base/Pronouns.java
eliasnogueira: Waiting for the resolution of this comment https://github.com/datafaker-net/datafaker/pull/1106#discussion_r1510218013

src/main/java/net/datafaker/providers/base/Pronouns.java
eliasnogueira: @bodiam any feedback about this last comment?

snuyanzin: same could achieved with 
```java
    public Pronouns pronouns() {
        return faker.getPronouns();
    }
```
however no need to make GC happy by creation a new object each time there is a call to `pronouns`

bodiam: @eliasnogueira it's a bit comment, not sure what to say :-) I'm not a fan of the fluent interface approach that much, so if you want to keep the Pronouns as a separate class, that's fine.

docs/documentation/providers.md
```diff
@@ -27,7 +27,7 @@ Number of providers per Datafaker version
 | 2.0.0   | 3                       | 233                       |
 | 2.1.0   | 4                       | 237                       |
 
-Datafaker comes with a total of 233 data providers:
+Datafaker comes with a total of 237 data providers:
```
bodiam, line 30: this document is auto generated, no need to update.

bodiam:
Anyone objecting if I merge this?

kingthorin:
Go for it

bodiam:
Thanks @eliasnogueira et all for the work on this!

eliasnogueira:
Thank you guys!

PR #1114, opened 2024-03-04
Updated the correct generation of Javadoc links.

bodiam:
Current documentation links are broken.

The old generated names are:
`https://javadoc.io/doc/net.datafaker/datafaker/latest/net/datafaker/providers/base/Name.html`

While the correct names are:
`https://javadoc.io/doc/net.datafaker/datafaker/latest/net.datafaker/net/datafaker/providers/base/Name.html`


bodiam:
(I've published the new documentation with the new links to Datafaker.net already, so they work again, but before this, all links published here were broken: https://www.datafaker.net/documentation/providers/#provider-groups)

kingthorin (approved):

Issue #1112, opened 2024-03-03
Duplicate faker groups (Money and Currency)


panilya:
It seems we have 2 faker groups with similar fakers: [src/main/java/net/datafaker/providers/base/Money.java](https://github.com/datafaker-net/datafaker/blob/main/src/main/java/net/datafaker/providers/base/Money.java) and [src/main/java/net/datafaker/providers/base/Currency.java](https://github.com/datafaker-net/datafaker/blob/main/src/main/java/net/datafaker/providers/base/Currency.java).

The best solution I see to this issue is to deprecate one of the faker groups. I want to hear your thoughts and what faker group we should deprecate.

After discussion, I will be happy to contribute to this change.

bodiam:
Good point. Might make sense to deprecate Currency,  delegate everything to Money, and we will remove Currency in 2.4 or 3.0. Would that work?

panilya:
Yes, I think this will work.

+1 to remove deprecated Currency faker group in the 3.0 version

snuyanzin:
I wonder why  have this class at all?
there is a java class `java.util.Currency` which could be reused instead of duplicating logic with `yaml` file which is slower

bodiam:
@snuyanzin good question. I think faker.Currency is pre Java 8, while util.Currency was introduced in Java 8. Might make sense to migrate it.

snuyanzin:
In fact it is present in Java since 1.4 ... 
In javadocs it is stated that it done based on ISO 4217, so probably would be better to use it

snuyanzin:
@bodiam , @panilya am I right that this one now can be closed?

panilya:
I think so.

Maybe it also makes sense to leave somewhere a reminder to delete Currency faker in the 3.0.0 release

bodiam:
I think it will be deleted in version 2.4.0 or 2.5.0. I know it's not entirely "semver", but 3.0 to me sounds like a huge milestone, not just the removal of a class which is is probably used by only a small set of our users.

kingthorin:
The code/javadoc have the removal todo details.

PR #1118, opened 2024-03-04
Use java.util.Currency in currency related fakers

panilya:
Issue: #1112


src/main/java/net/datafaker/providers/base/Currency.java
```diff
@@ -2,18 +2,25 @@
 
 /**
  * @since 0.8.0
+ * @deprecated since 2.3.0. For removal in 3.0.0 version. This faker is deprecated due to the existence
+ * of duplicate {@link Money} faker with similar methods. Use {@link Money} instead.
  */
+@Deprecated(since = "2.3.0", forRemoval = true)
```
kingthorin: 2.2.0 is next I believe

src/main/java/net/datafaker/providers/base/Country.java
```diff
@@ -27,12 +34,22 @@ public String capital() {
...
     }
 
+    /**
+     * @see Currency#getDisplayName()
+     * @return a random detailed ISO 4217 currency display name
+     */
     public String currency() {
-        return resolve("country.currency");
+        int randomIndex = faker.random().nextInt(availableCurrencies.size());
+        return availableCurrencies.get(randomIndex).getDisplayName();
```
kingthorin, line 43: Is there a reason for the intermediate store when it could just be inlined?

src/main/java/net/datafaker/providers/base/Currency.java
```diff
@@ -2,18 +2,25 @@
...
 public class Currency extends AbstractProvider<BaseProviders> {
 
     public Currency(BaseProviders faker) {
         super(faker);
     }
 
+    /**
+     * @deprecated since 2.3.0. For removal in 3.0.0 version. Use {@link Money#currency} instead.
```
kingthorin: IMHO if you're going to javadoc this then also include:
`@Deprecated(since = "2.2.0", forRemoval = true)`

Same throughout.

bodiam (approved):

src/main/java/net/datafaker/providers/base/Currency.java
```diff
@@ -2,32 +2,29 @@
...
+@Deprecated(since = "2.2.0", forRemoval = true)
 public class Currency extends AbstractProvider<BaseProviders> {
 
     public Currency(BaseProviders faker) {
         super(faker);
     }
 
+    /**
+     * @deprecated since 2.2.0. For removal in 3.0.0 version. Use {@link Money#currency} instead.
+     */
```
kingthorin, line 17: If you're going to JavaDoc that much I'd also include: `@Deprecated(since = "2.2.0", forRemoval = true)`

src/main/java/net/datafaker/providers/base/Money.java
```diff
@@ -1,31 +1,65 @@
...
+        int randomIndex = faker.random().nextInt(availableCurrencies.size());
+        return availableCurrencies.get(randomIndex).getCurrencyCode();
+    }
+
+    /**
+     * <p>Returns the 3 digit ISO 4217 numeric code of a random currency. Read more: </p>
+     *
+     * @see Currency#getNumericCodeAsString()
+     * @return an ISO 4217 currency numeric code
+     * @since 2.3.0
```
kingthorin, line 48: 2.2.0

src/main/java/net/datafaker/providers/base/Money.java
```diff
@@ -1,31 +1,65 @@
...
+     */
+    public String currencyNumericCode() {
+        int randomIndex = faker.random().nextInt(availableCurrencies.size());
+        return availableCurrencies.get(randomIndex).getNumericCodeAsString();
+    }
+
+    /**
+     * @see Currency#getSymbol()
+     * @return an ISO 4217 currency symbol
+     * @since 2.3.0
```
kingthorin, line 58: 2.2.0

src/main/java/net/datafaker/providers/base/Country.java
bodiam: That would be quite a long line don't you think?

src/main/java/net/datafaker/providers/base/Currency.java
bodiam: I think if the whole class is marked for removal, that  should be enough no?

src/main/java/net/datafaker/providers/base/Country.java
kingthorin: 🤷‍♂️ spotless should take care of it

src/main/java/net/datafaker/providers/base/Currency.java
kingthorin: I guess, I'm not sure if inherit is the default. Figured it was best to be specific for users/implementers. Not a blocker I guess

snuyanzin (changes requested):
I tend to think that if we remove `money.yml` then we need to remove it from 
`net.datafaker.service.files.EnFile` as well

src/main/java/net/datafaker/providers/base/Money.java
```diff
@@ -1,31 +1,65 @@
...
      */
     public String currency() {
-        return resolve("money.currency");
+        int randomIndex = faker.random().nextInt(availableCurrencies.size());
+        return availableCurrencies.get(randomIndex).getDisplayName();
     }
 
     /**
-     * Method returns a randomly generated currency value like EUR, USD, etc.
+     * <p>Returns a random ISO 4217 currency code (ex. EUR, USD)</p>
```
kingthorin, line 33: The paragraph tags seem unnecessary

panilya: Fixed

src/main/java/net/datafaker/providers/base/Currency.java
panilya: Make sense. Resolved.

thank you for the review

kingthorin (commented):
Seems to be a lot of duplicate code, why not just have money and country call the methods in currency?

Also could use List.of instead of new ArrayList for the available currencies.

Edit: I also don't see the enfile change.

snuyanzin:
> Seems to be a lot of duplicate code, why not just have money and country call the methods in currency?
Also could use List.of instead of new ArrayList for the available currencies.

+1

Also how about making deprecating `Currency` exntending `Money` and remove al the methods from `Currency` then it will automatically cal methods from `Money` ?

bodiam:
> Also could use List.of instead of new ArrayList for the available currencies.

-1. There's no such constructor to turn a Set into a List.

bodiam:
> Seems to be a lot of duplicate code, why not just have money and country call the methods in currency?

I'm not sure if something changed in the meantime, but there's hardly any code in Currency, which duplicate code do you mean?

snuyanzin:
>>Also could use List.of instead of new ArrayList for the available currencies.

>-1. There's no such constructor to turn a Set into a List.

in fact there is such, e.g. here 
```java
this.availableCurrencies = new ArrayList<>(Currency.getAvailableCurrencies());
```

and it could be replaced with 
```java
this.availableCurrencies =Lis.of(Currency.getAvailableCurrencies());
```

May be `duplicated` code is not the right word here however e.g. `Currency#code` is
```java
    public String name() {
        return resolve("currency.name");
    }
```
which does the same  or almost the same thing as
```java
public String currencyCode() {
        int randomIndex = faker.random().nextInt(availableCurrencies.size());
        return availableCurrencies.get(randomIndex).getCurrencyCode();
    }
```

bodiam:
> this.availableCurrencies = List.copyOf(Currency.getAvailableCurrencies());

`copyOf` would work, didn't think of that, nice! (not really sure how that would be much better than `new ArrayList`)

> Regarding duplication

Sure, but that Currency thing is going to be deleted one day anyway, does it matter much? It's a genuine question, I'm just wondering if that should hold up this PR, that's all.

panilya:
>  I've just looked through the code once more and it looks like there is the duplicated logic for Money and for Country e.g.
> Money#currencyCode and Country#currencyCode same for currency method. Why do we need this?

I agree about duplicate code and will push the fix soon.

I thought about this (duplicate code or reuse Money faker) when I first implemented these changes, but I thought it might be better to not reuse Money (don't remember exactly the reason why I thought so, sorry). But it is better to reuse Money, indeed.

Thank you all for reviewing.

bodiam:
> Money#currencyCode and Country#currencyCode

Don't know, but we have these methods for years (one is from 2019, the other from 2022). And at least the currency codes are now gone from the YML, and these currency codes in money.yml and country.yml were only half duplicated. At least now they're in sync. Small bit of duplication; we could remove one of the methods, or have one of them call the other. Or we leave them like they are. I lean towards leaving them like they are, but if there's good arguments for a different approach, I'm fine with that too. Ideally I don't want to break the API too much, Currency being a bit of an exception perhaps.

kingthorin:
RE: Duplication for my two cents these are basically all the same:

```
    public String currencyCode() {
        return resolve("country.currency_code");
        int randomIndex = faker.random().nextInt(availableCurrencies.size());
        return availableCurrencies.get(randomIndex).getCurrencyCode();
    }
```

They could be genericized to a single method that gets the index and accept a function to execute.

kingthorin (approved):
LGTM

snuyanzin (approved):
Thanks for addressing feedback

PR #1124, opened 2024-03-12
Update doc to fit the latest source code version

benallard:
The `static` only makes sense when the class is defined as an inner class, but that is not mentioned, and was confusing me as well.


bodiam (approved):

bodiam:
Hi @benallard , thanks for the fix, looks good!

PR #1128, opened 2024-03-16
Added Ticker datasets for NSE and LSE exchanges

ChiragMangukia:
Added Scrips from the National Stock Exchange (NSE) India and the London Stock Exchange


bodiam:
Seems these files are getting a bit big. Your contribution is most welcome though, I'll merge it in! Thanks!

ChiragMangukia:
Thank you @bodiam! I have a question. I noticed that we have Nasdaq and NYSE tickers in the Finance provider as well. Is it intentional to keep them in both providers?

bodiam:
@ChiragMangukia hips point. Probably not, and it might be good to deprecate the one in Finance. We usually don't introduce breaking changes without a warning depreciation first, but there's generally not a great reason to have the same list in 2 places.

ChiragMangukia:
Noted!

Addressed it in an another PR #1129

Thank you!

PR #1129, opened 2024-03-16
Deprecate fin tickers from Finace provider

ChiragMangukia:
I've deprecated the NSDQ, NYSE, and stock market methods (as suggested in #1128) in the Finance provider. Furthermore, I've added the stockMarket method (renaming it as 'exchanges') to the Stocks provider as it's more relevant to have it there.


bodiam (approved):
Great, looks good to me. Thanks for addressing this!

Issue #1109, opened 2024-03-03
Possible JSON generation error


eliasnogueira:
## Description

The [JSON Schema Transformation](https://www.datafaker.net/documentation/schemas/#json-transformation) seems to generate an invalid JSON output when it generates more than 2 objects.

## Problem

When it generates 2 or more objects, the JSON starts as an Object instead of an Array.

## Steps to reproduce

Getting the example in the datafaker documentation from [JSON Schema Transformation](https://www.datafaker.net/documentation/schemas/#json-transformation) we have the following code:

```java
Schema<Object, ?> schema = Schema.of(
    field("Text", () -> faker.name().firstName()),
     field("Bool", () -> faker.bool().bool())
 );

JsonTransformer<Object> transformer = JsonTransformer.builder().build();
 String json = transformer.generate(schema, 2);
```

### Expected result

As it's generating 2 objects, the output must start as an Array

```json
[
   {
      "Text":"Nathalie",
      "Bool":false
   },
   {
      "Text":"Tamika",
      "Bool":true
   }
]
```

### Actual result

The output starts as an object.

```json
{
   {
      "Text":"Nathalie",
      "Bool":false
   },
   {
      "Text":"Tamika",
      "Bool":true
   }
}
```

### How to validate

There are two easy ways to validate.

#### Transform the output into a JSON object

1. Import the [json](https://mvnrepository.com/artifact/org.json/json/20240205) Java library
2. Use the same code as in the docs,  using the generated JSON output to create a JSON object

```java
 Schema<Object, ?> schema = Schema.of(
  field("Text", () -> faker.name().firstName()),
  field("Bool", () -> faker.bool().bool())
);

JsonTransformer<Object> transformer = JsonTransformer.builder().build();
String json = transformer.generate(schema, 2);

 new JSONObject(json); // an exception will happen here
```

#### Validate the output in any online JSON formatted

1. Access https://jsonchecker.com/
4. Copy the output from the _Actual results_ section
5. Check the output

## Versions

 - OS: MacOS 14.3.1
 - JDK: openjdk version "21" 2023-09-19
 - Faker version: 2.1.0

## Additional information

[...]

PR #1134, opened 2024-03-22
Changes in the Json Transform

eliasnogueira:
## Description

This PR changes the `JsonTransformer` by generating the output as valid JSON data (except by the without comma approach). The output will always be valid when one or more data is generated.

Added extra `prettyPrint()` method to enable a more readable output.

## Changed

- Added the `prettyPrint()` method into `JsonTransformer` class along with the `json` library
  - This is set by default and will output the JSON value in a pretty print format
  - When disabled, one line String is generated
  - When using the `withCommaBetweenObjects`, the `prettyPrint()` is "disabled" (set to `false`) due to a parse error
    - parse error when there's no comma for multiple items
- Refactored `JsonTest`, `FakeStreamTest`, and `FakeCollectionTest` for methods using the `JsonTransformer` class

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,6 @@
                          locale. Turkish is a notoriously tricky locale. -->
         <surefire.argline>@{argLine} -Duser.language=TR -Duser.country=tr</surefire.argline>
         <kotlin.version>1.9.23</kotlin.version>
-        <json.version>20240303</json.version>
     </properties>
     <licenses>
         <license>
@@ -176,11 +175,6 @@
             <artifactId>kotlin-test</artifactId>
             <version>${kotlin.version}</version>
             <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.json</groupId>
-            <artifactId>json</artifactId>
-            <version>${json.version}</version>
         </dependency>
     </dependencies>
     <distributionManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -55,6 +55,7 @@
                          locale. Turkish is a notoriously tricky locale. -->
         <surefire.argline>@{argLine} -Duser.language=TR -Duser.country=tr</surefire.argline>
         <kotlin.version>1.9.23</kotlin.version>
+        <json.version>20240303</json.version>
     </properties>
     <licenses>
         <license>
@@ -175,6 +176,11 @@
             <artifactId>kotlin-test</artifactId>
             <version>${kotlin.version}</version>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.json</groupId>
+            <artifactId>json</artifactId>
+            <version>${json.version}</version>
         </dependency>
     </dependencies>
     <distributionManagement>
--- a/src/main/java/net/datafaker/transformations/JsonTransformer.java
+++ b/src/main/java/net/datafaker/transformations/JsonTransformer.java
@@ -1,6 +1,8 @@
 package net.datafaker.transformations;
 
 import net.datafaker.sequence.FakeSequence;
+import org.json.JSONArray;
+import org.json.JSONObject;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
@@ -8,18 +10,18 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.Map;
-import java.util.StringJoiner;
 
 public class JsonTransformer<IN> implements Transformer<IN, Object> {
 
     private static final Map<Character, String> ESCAPING_MAP = createEscapeMap();
-
-    private final char[] wrappers;
+    private static final char[] WRAPPERS = "[]".toCharArray();
     private final boolean commaBetweenObjects;
+    private final boolean prettPrint;
 
-    private JsonTransformer(char[] wrappers, boolean commaBetweenObjects) {
-        this.wrappers = wrappers;
+
+    private JsonTransformer(boolean commaBetweenObjects, boolean prettyPrint) {
         this.commaBetweenObjects = commaBetweenObjects;
+        this.prettPrint = prettyPrint;
     }
 
     public static <IN> JsonTransformer.JsonTransformerBuilder<IN> builder() {
@@ -39,9 +41,7 @@ public class JsonTransformer<IN> implements Transformer<IN, Object> {
             } else {
                 applyValue(input, sb, ((SimpleField) fields[i]).transform(input));
             }
-            if (i < fields.length - 1) {
-                sb.append(", ");
-            }
+            if (i < fields.length - 1) sb.append(", ");
         }
         sb.append("}");
         return sb.toString();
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -55,6 +55,7 @@
                          locale. Turkish is a notoriously tricky locale. -->
         <surefire.argline>@{argLine} -Duser.language=TR -Duser.country=tr</surefire.argline>
         <kotlin.version>1.9.23</kotlin.version>
+        <json.version>20240303</json.version>
```
bodiam, line 58: Besides the pretty printing, what do you need this library for?

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -157,17 +169,7 @@ private static Map<Character, String> createEscapeMap() {
...
-            // covered by map.put('\b', "\\b");
-            // map.put('\u0009', "\\u0009");
-            // covered by map.put('\t', "\\t");
-            // map.put((char) 10, "\\u000A");
-            // covered by map.put('\n', "\\n");
             Map.entry('\u000B', "\\u000B"),
-            // map.put('\u000C', "\\u000C");
-            // covered by map.put('\f', "\\f");
-            // map.put((char) 13, "\\u000D");
-            // covered by map.put('\r', "\\r");
```
snuyanzin, line 170: I don't think we need to remove it
this clarifies why we don't have `map.put('\u0008', "\\u0008");` and thers

otherwise each time this question will pop up

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -53,31 +53,43 @@ public String generate(Iterable<IN> input, Schema<IN, ?> schema) {
...
+                sb.append(apply(null, schema, i));
+                if (commaBetweenObjects && i < limit - 1) sb.append(",");
             }
         }
-        return limit > 1 ? wrappers[0] + LINE_SEPARATOR + sb + LINE_SEPARATOR + wrappers[1] : sb.toString();
+
+        String result = "" + WRAPPERS[0] + sb + WRAPPERS[1];
+
+        return prettPrint ? result.startsWith("{")
+            ? new JSONObject(result).toString(2) : new JSONArray(result).toString(2) : result;
```
snuyanzin, line 87: This is a breaking change 
and this is not clear why we need it.

The main idea behind was that the output was in JSON Lines format (https://jsonlines.org/)

And the main feature is that it could be easily copied and pasted into BigQuery as stated in their doc (https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json)

Now i don't see how it could be possible

pom.xml
eliasnogueira: Directly for the pretty print. Indirectly to ensure that the JSON output is valid when using the `JSONObject` and `JSONArray` classes.

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -1,25 +1,27 @@
...
-import java.util.StringJoiner;
 
 public class JsonTransformer<IN> implements Transformer<IN, Object> {
 
     private static final Map<Character, String> ESCAPING_MAP = createEscapeMap();
-
-    private final char[] wrappers;
+    private static final char[] WRAPPERS = "[]".toCharArray();
     private final boolean commaBetweenObjects;
+    private final boolean prettPrint;
```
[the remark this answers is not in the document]
eliasnogueira: Fixed

src/test/java/net/datafaker/formats/JsonTest.java
```diff
@@ -5,25 +5,18 @@
...
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
-import java.util.AbstractMap;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Random;
+import java.util.*;
```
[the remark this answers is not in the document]
eliasnogueira: Fixed

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -39,9 +41,7 @@ public String apply(IN input, Schema<IN, ?> schema) {
             } else {
                 applyValue(input, sb, ((SimpleField) fields[i]).transform(input));
             }
-            if (i < fields.length - 1) {
-                sb.append(", ");
-            }
+            if (i < fields.length - 1) sb.append(", ");
```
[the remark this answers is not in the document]
eliasnogueira: Rolled back.

src/main/java/net/datafaker/transformations/JsonTransformer.java
eliasnogueira: Reverting to the previous state.

Wouldn't be better to add this as a comment in the method instead of commented code?
Something like this:

```java
    /*
     * The following entries are not added to the map because they are covered by another existing one
     *
     * map.put('\u0008', "\\u0008"); -> covered by map.put('\b', "\\b");
     * map.put('\u0009', "\\u0009"); -> covered by map.put('\t', "\\t");
     * map.put((char) 10, "\\u000A"); -> covered by map.put('\n', "\\n");
     * map.put('\u000C', "\\u000C"); -> covered by map.put('\f', "\\f");
     * map.put((char) 13, "\\u000D"); -> covered by map.put('\r', "\\r");
     */
```

src/main/java/net/datafaker/transformations/JsonTransformer.java
eliasnogueira: I understand your concern @snuyanzin, and I do agree it's a breaking change as the way to generate large datasets is now missing.

First, the "regular" JSON output was wrong as it wasn't taking the correct format, based on its semantics. See [JSON](https://www.json.org/json-en.html).
The output was, all the time, an object even when the output was an array, and this change fixed it.

Now, from my point of view, the way this feature was created this biased in only one usage, which might be not the common usage for the Datafaker consumers. Datafaker has delivered this feature thinking about generating JSON output to be processed for large external datasets or tools, like logging systems and the mentioned BigQuery example. And it's awesome!

As an end user, my first way of thinking is that Datafaker will generate a valid JSON for most of the use cases.
Nonetheless, it wasn't creating a valid JSON for any other purpose. The https://github.com/datafaker-net/datafaker/issues/1109 puts some light on it.

I think that a decision should be made, in terms of which format Datafaker will support it.
It can be one, it can be both.

Maybe a new builder in the `JsonTransformerBuilder` can be added to take care of the large dataset output.

I will follow any decision you guys make, but with the note that the current approach seems not valid for most Datafaker users.

snuyanzin: May be i was not clear enough, sorry
My point not about valid not valid json

Fixing json is ok

The problem is that here we change output format: now there either everything in one line or pretty

at the same time `jsonline` assumes that there should be one top level entity per line and I don't know how can we have with the proposed solution. 

So do we really need this `json ` lib usage? I guess we could fix wrong json issue without breaking existing approach

eliasnogueira: Sorry for my misunderstanding.

The library is not that important if we make sure the JSON is correctly created. I intended to have this as a double-check, plus the pretty output.

I would say that we can keep all, making the pretty not default. By the way, when we apply the approach without commas, which is the JSON line solution, the pretty is not applied, but I have changed the code to have it in one line.

Can I propose to remove the one-line approach and keep the pretty output?

bodiam: Ideally, I would like the library to be removed. If that means no pretty printing, then I don't mind, I think that's a nice to have, and I don't think that's the responsibility of Datafaker.

snuyanzin: The problem with pretty print: measurement show that it 3 times slow for 1, 10, 1000 entities
```
Benchmark                              Mode  Cnt     Score    Error   Units
DatafakerSimpleMethods.json1          thrpt   10  2517.912 ± 43.562  ops/ms
DatafakerSimpleMethods.json10         thrpt   10   252.746 ±  3.304  ops/ms
DatafakerSimpleMethods.json100        thrpt   10    25.772 ±  0.437  ops/ms
DatafakerSimpleMethods.json100Pretty  thrpt   10     8.306 ±  0.013  ops/ms
DatafakerSimpleMethods.json10Pretty   thrpt   10    77.777 ±  2.426  ops/ms
DatafakerSimpleMethods.json1Pretty    thrpt   10  2513.785 ± 14.620  ops/ms
```

eliasnogueira: No problem.
I will remove the library and other changes, keeping the original fix.

kingthorin:
Don't open a new PR, just keep pushing to this branch.

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -53,31 +55,43 @@ public String generate(Iterable<IN> input, Schema<IN, ?> schema) {
             throw new IllegalArgumentException("The sequence should be finite of size");
         }
 
-        StringJoiner data = new StringJoiner(LINE_SEPARATOR);
+        StringBuilder sb = new StringBuilder();
```
snuyanzin, line 58: Why do we need to remove usage of line separator here?

I agree that there might be cases when it is preferred to get result without it however sometimes it is better to have it.

If you really need to have result without it then i would suggest to add dedicated method to json transformer builder to set lineseparator, then both old and new behavior will be supported

eliasnogueira:
@snuyanzin @kingthorin @bodiam Finally the changes are here!
I've removed the JSON library along with the pretty print, and the text block usage (it's causing issues on Java 22 + Windows).

All tests green!

kingthorin:
Sorry I missed this somehow.

@snuyanzin thanks for approving/merging!

PR #1154, opened 2024-04-13
Fix address for en-NZ locale

### LINKED ISSUE #1153: Fix address for 'en-NZ' locale
**Describe the bug**
The region and region_abbr do not seem to be getting picked up for the locale en-NZ.  This is because  [Address](https://github.com/datafaker-net/datafaker/blob/252125385b65b86aa489ad7cbdd87efe97ba3c5f/src/main/java/net/datafaker/providers/base/Address.java) does not have a concept of _region_ and _region_abbr_ and instead knows about [_state_](https://github.com/datafaker-net/datafaker/blob/252125385b65b86aa489ad7cbdd87efe97ba3c5f/src/main/java/net/datafaker/providers/base/Address.java#L93) and _[state_abbr](https://github.com/datafaker-net/datafaker/blob/252125385b65b86aa489ad7cbdd87efe97ba3c5f/src/main/java/net/datafaker/providers/base/Address.java#L97)_. 

**To Reproduce**
```
jshell> faker.address().fullAddress();
$3 ==> "5351 Scott Estate Dr, South Samuelland, AK 9311"

jshell> faker.address().state();
$6 ==> "Wyoming"

jshell> faker.address().stateAbbr();
$7 ==> "ND"

```

The data is not being picked up from [_en-NZ.yml_](https://github.com/datafaker-net/datafaker/blob/252125385b65b86aa489ad7cbdd87efe97ba3c5f/src/main/resources/en-NZ.yml)

**Expected behavior**
The data should be picked up from the relevant locale file. 

**Versions:**
 - OS:MacOS
 - JDK 17
 - Faker Version v2.1.0

**Additional context**
I've already made progress to fix this along with adding a bit more features for address in en-NZ locale.

k1w1dev:
resolves  #1153

### DIFF
--- a/src/main/resources/en-NZ.yml
+++ b/src/main/resources/en-NZ.yml
@@ -19,13 +19,75 @@ en-NZ:
     internet:
       domain_suffix: [co.nz, com, net.nz, net, org.nz, org, govt.nz, iwi.nz]
     address:
+      full_address:
+        - "#{urban_address}"
+        - "#{delivery_address}"
+        - "#{rural_address}"
+      country_code: ['NZ']
+      state_abbr: "#{region_abbr}"
+      state: "#{region}"
       region_abbr: [NTL, AUK, WKO, BOP, GIS, HKB, TKI, MWT, WGN, TAS, NSN, MBH, WTC, CAN, OTA, STL]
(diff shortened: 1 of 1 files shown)


bodiam:
Thanks for your contribution @k1w1dev , much appreciated! Looks good, merged!

PR #1159, opened 2024-04-16
Adding a regexify test case for metacharacters inside a character set

### LINKED ISSUE #1158: Using metacharacters on a character set with regexify does not produce the expected text
# Description

Using `\d` instead of `0-9` or `\w` instead of `a-zA-Z0-9_` inside a character set in the `regexify` method generates either the expected character or an `[` randomly, plus a `]` character.

# Sample code

```java
import net.datafaker.Faker;

class Scratch {
    public static void main(String[] args) {
        var faker = new Faker();

        faker.regexify("[0-9]"); // a random digit
        faker.regexify("\\d");   // also a random digit
        faker.regexify("[\\d]"); // 2 characters, the first randomly being either a digit or `[`, and a `]`

        faker.regexify("[a-zA-Z0-9_]"); // a word characters
        faker.regexify("\\w");          // also a word characters
        faker.regexify("[\\w]");        // 2 characters, the first randomly being either a word character or `[`, and a `]`
    }
}
```

My guess is that `\d` and `\w` are being replaced to their respective character set directly without checking if it's already on a character set, generating a regex like `[[0-9]]` for `[\d]` and `[[a-zA-Z0-9_]]` for `[\w]`, which would explain the actual behavior.

# Expected behavior

Using `\d` and `\w` should have the same behavior as using `0-9` and `a-zA-Z0-9_` respectively, regardless of being in a character set.

# Versions

- OS: Linux
- JDK: zulu 17.44.17
- Faker Version: 2.1.0

# Workaround

As shown in the sample code, using `0-9` instead of `\d` and `a-zA-Z0-9_` instead of `\w` when inside a character set works normally.

# Additional context

I've also tested on the `main` branch, and after replacing the [Generex](https://github.com/mifmif/Generex) dependency with [RgxGen](https://github.com/curious-odd-man/RgxGen) in [this commit](https://github.com/datafaker-net/datafaker/commit/84674dc756a6c4b48f589d6890d85e94a991df47), it seems to be working normally.

Since the root problem is on a dependency that the project is already switching away from and there's a reasonable workaround to work with in the meantime, maybe it's better to just add a test case for this instead of fixing it on Generex (which looks like it's not maintained anymore anyway) or working around it here in the repo.

eduhoribe:
Closes #1158

### DIFF
--- a/src/test/java/net/datafaker/FakerTest.java
+++ b/src/test/java/net/datafaker/FakerTest.java
@@ -109,7 +109,8 @@ class FakerTest extends AbstractFakerTest {
     Test case for issue https://github.com/datafaker-net/datafaker/issues/1091
      */
     @ParameterizedTest
-    @ValueSource(strings = {"\\d",
+    @ValueSource(strings = {
+        "\\d",
         "\\w",
         "[aeiou]{2,3}",
         "[a-z]{2,3}",
@@ -123,7 +124,9 @@ class FakerTest extends AbstractFakerTest {
         "(aeiou)",
         "(a|b)",
         "\\.\\*\\?\\+",
-        "^arn:.+:.+:.*:([0-9]{12}):(.+)$"})
+        "^arn:.+:.+:.*:([0-9]{12}):(.+)$",
+        "\\s[\\s]\\d[\\d]\\w[\\w]",
+    })
     void testRegexify(String input) {
         var faker = new Faker();
         assertThat(faker.regexify(input)).matches(input);


src/test/java/net/datafaker/FakerTest.java
```diff
@@ -14,13 +14,7 @@
...
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Random;
-import java.util.Set;
+import java.util.*;
```
snuyanzin, line 17: please do not use `import *`

bodiam:
Thanks for adding this PR. Happy to accept it if you could please revert the `*` import.

eduhoribe:
Import changes reverted, thanks for the review!

kingthorin (approved):

snuyanzin (approved):

snuyanzin:
Thanks for addressing feedback

Issue #1170, opened 2024-04-24
NullPointerException when using Faker#locality


f-cramer:
**Describe the bug**
I get a `NullPointerException` when calling `Faker#locality`. The message is `java.lang.NullPointerException: Cannot invoke "java.nio.file.Path.toString()" because the return value of "java.nio.file.Path.getFileName()" is null`

**To Reproduce**
https://github.com/f-cramer/datafaker-npe
Just run `gradlew test` to reproduce

**Expected behavior**
Initialization of Locality should complete without throwing an exception

**Versions:**
 - OS: Windows 11, could also reproduce in a Github Action running on ubuntu-latest
 - JDK 17
 - Faker 2.2.1, but seems to happen since at leat 2.1.0

bodiam:
Thanks for reporting, looking into it!

bodiam:
2.2.2 has been released to address this issue.

f-cramer:
That great. Thanks for the very quick fix.

Issue #1168, opened 2024-04-24
v2.2.0: Exceptions when calling certain Faker methods


bpicode:
**Describe the bug**
Observed the following exceptions:
* `java.lang.RuntimeException: Unable to resolve #{IDNumber.ssn_valid} directive for FakerContext FakerContext{, locale=SingletonLocale{locale=en}, randomService=net.datafaker.service.RandomService@1}.`
* `java.lang.NoClassDefFoundError: net/datafaker/shaded/curiousoddman/rgxgen/RgxGen`

**To Reproduce**
To reproduce, call Faker as follows
* `(new Faker()).idNumber().valid()`
* `(new Faker()).regexify("[a-z]{6}")`

**Expected behavior**
No exceptions thrown.

**Versions:**
 - JDK: OpenJDK Runtime Environment Corretto-17
 - Faker Version: 2.2.0

**Additional context**
With version 2.1.0 it was still working.

snuyanzin:
Thanks for reporting
yeah, shading of hyphens might be tricky...

the issue is not reproduced in datafaker tests since shading is happening after that
I was able to reproduce it with [datafaker-gen](https://github.com/datafaker-net/datafaker-gen) where datafaker is used a separate lib

after this fix it starts working again
@bpicode could you please double check that this fix also helps with your setup?
https://github.com/datafaker-net/datafaker-gen/actions/runs/8814194154/job/24193516670?pr=9

bpicode:
@snuyanzin How can I check the fix? Should I build this lib with the commit https://github.com/datafaker-net/datafaker/pull/1169/commits/7a5fb07715600ad39e99bf204b3550c18243ea5e locally?

snuyanzin:
yes, currently building it locally and publish to local maven repo (e.g. `mvn clean install`) and then use it as dependency (version `2.2.1-SNAPSHOT`)  seems to be the only way to check

bodiam:
Hi all, thanks for the quick action on this. If this is confirmed to be fixed, happy to release a new version.

bpicode:
Yes I can confirm that this is fixed. Thanks for the quick responses.

snuyanzin:
thanks for checking
yep, i think it make sense to have a bug fix release since this issue makes all expression related functionality non working...
sorry about that

PR #1171, opened 2024-04-24
Add tests against downstream project (datafaker-gen)

snuyanzin:
the idea is to get main repo from datafaker-gen and build it and then run tests against latest datafaker version to be able to detect issue appearing only in downstream projects like e.g. #1170 or #1168 

currently this check is done only in case of jdk21 and ubuntu (in fact i do not see a real reason to test it with the whole matrix)

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -12,6 +12,8 @@ on:
 env:
   # Quotes are here since Windows has issues with parsing of maven args with dots
   mvn_options: --batch-mode -q "-Dstyle.color=always" "-Dorg.slf4j.simpleLogger.showDateTime=true" "-Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss.SSS"
+  jdk_for_downstream: 21
+  os_for_downstream: ubuntu-latest
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -21,6 +21,8 @@ jobs:
         runs-on: [ ubuntu-latest, macos-latest, windows-latest ]
     name: Jdk ${{ matrix.java-version }}, os ${{ matrix.runs-on }}
     runs-on: ${{ matrix.runs-on }}
+    outputs:
+      output1: ${{ steps.getProjectVersion.outputs.test }}
```
kingthorin, line 25: Why not actually call it version or something meaningful?

snuyanzin: still playing to get something working
after that will rename it

snuyanzin:
//cc @bodiam since you were asking about something like this

.github/workflows/maven.yml
```diff
@@ -30,6 +32,20 @@ jobs:
           distribution: 'temurin'
           cache: maven
       - name: Build with Maven
-        run: ./mvnw $mvn_options verify -PnoGpg --file pom.xml
+        run: ./mvnw $mvn_options install -PnoGpg --file pom.xml
```
snuyanzin, line 35: we need `install` since during tests with downstream project these artifacts will be reused

bodiam: What would you think of turning all the tests of Datafaker into a test artifact, and run them as part of the downstream project?

kingthorin:
Sorry I missed the request on this one. It does look good 🙂

.github/workflows/maven.yml
snuyanzin: I guess most of them are independent whether they are running within Datafaker code or some downstream projects.
There are only few of them where it works differently.

Also there will be another challenge: how to make it possible
I see several solutions and none of them is perfect
1. copy-paste of tests however it means we need to maintain 2 repos synced at least from tests point of view
2. we could publish `test-jar` and reuse it in downstream project however anyway need to create a bunch of child classes and keep it sync (e.g. each time a new test class is appearing in Datafaker we need to create a corresponding child in third party)
3. something else that I don't know yet

bodiam: Sorry, I don't understand point 2. Why would we need to create child classes? The idea I had was to indeed create a test artifact, put all the tests in there, and use surefire to run the tests against the jar. I'm not sure if we do some magic which prevents this?

Alternatively, I think we need only 1 test, which is very similar to the repeatability test : just call every in the public api and make sure it returns something. That would have stopped a few bugs from being released.

snuyanzin: May be I missed something however I thought that tests will not run if we just add a `test-jar` dependency without any usage of that

bodiam: Well, here it says we could run them if we package them up and give them to surefire, unless I'm misunderstanding something:

https://stackoverflow.com/questions/10496846/run-junit-tests-contained-in-dependency-jar-using-maven-surefire

snuyanzin: may be makes sense to try it once more 
not sure when will have time for this

PR #1172, opened 2024-04-25
Parent filename could be null for resources in jar

### LINKED ISSUE #1170: NullPointerException when using Faker#locality

snuyanzin:
This should fix #1170 

to reproduce use datafaker-gen, especially this PR this https://github.com/datafaker-net/datafaker-gen/pull/10

to check that it is fixed use same PR and run it with (assuming this fixed is built and installed in local maven repo)
```
./mvnw clean install -Ddatafaker.version=2.2.2-SNAPSHOT
```

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Locality.java
+++ b/src/main/java/net/datafaker/providers/base/Locality.java
@@ -53,8 +53,8 @@ public class Locality extends AbstractProvider<BaseProviders> {
         final String filename = file.getFileName().toString().toLowerCase(Locale.ROOT);
         if ((filename.endsWith(".yml") || filename.endsWith(".yaml")) && Files.isRegularFile(file) && Files.isReadable(file)) {
             final Path parent = file.getParent();
-            final String parentFileName = parent == null ? null : parent.getFileName().toString();
-            if (langs.contains(parentFileName)) {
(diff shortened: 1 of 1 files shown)


bodiam:
Hi @snuyanzin , what's the root cause for this? Any chance we could have a unit test for this?

snuyanzin:
yep, that's the idea of this draft PR https://github.com/datafaker-net/datafaker/pull/1171
Unfortunately haven't succeed yet...

The idea is first to build datafaker and then build datafker-gen as a downstream project with datafaker E2E tests

the root cause: when resource files are in jar  it might happen that the parent for some of them is just `/` and that means the parentname is null... at the same time it is not the case for datafaker code itself, only for downstream projects

PR #1174, opened 2024-04-25
Remove outdated jdk11 profile

snuyanzin:


kingthorin:
I have nothing against this change, I don't know maven well enough. Should we just be changing this to 17+?

Issue #1137, opened 2024-03-28
Extend medical providers to include more medicine-related values


filipowm:
Currently medical providers contain only medicine names, disease names, hospitals, symptoms, diagnosis and procedure codes, medical professions. Since I'm working in healthcare space, I would like to leverage more providers for:
- medicine codes (like RxNorm)
- laboratory results codes (like LOINC)
- anatomical sites
- allergies
- personal characteristics (eg. eye color, hair color)

I would be happy to help and start adding those by myself. Also, due to growing number of medical providers, I would consider extracting existing medical providers from `net.datafaker.providers.base.Medical` to dedicated `net.datafaker.providers.medical.*`, where we could have dedicated `Medication`, `Diagnosis`, `Diagnostics` etc..

bodiam:
Hi @filipowm , great suggestion, we would absolutely appreciate a contribution from someone who has experience in the healthcare domain, so yes, a PR would be very welcome, and happy to help if you need it!

filipowm:
Fantastic. So let me play around it a bit in upcoming days/weeks 😉

filipowm:
I started from moving around existing code to suit new structure (and grouping some stuff together), and I started thinking if it wouldn't be worthwhile to a bit "break" how providers are structured and make API more fluent, by providing here intermediate grouping provider.

Details below ↓

### The idea

Logically group providers, in this case under `healthcare`providers group there will be other providers related to medical stuff.

#### As-is

Example code how it works today:

```java
var medicine = faker.medical().medicineName();
var symptom = faker.medical().symptoms();
var blood = faker.bloodtype().bloodGroup();
```

#### To-be

Let's compare with previous API:

```java
var medicine = faker.healthcare().medication().drugName(); // well, it might be another long discussion about drug vs medicine ;-)
var symptom = faker.healthcare().condition().symptom();
var blood = faker.healthcare().blood().group();
```

Looking at the class diagram, it would simply be about adding grouping providers interface like `HealthcareProvidersGroup`, that `Faker` would implement.

```mermaid
classDiagram
    Faker ..|> HealthcareProvidersGroup
    HealthcareProvidersGroup --|> ProviderRegistration : I'm not sure if this implements would\n make sense here, cause it's just\n grouping of other ProviderRegistration(s)
    HealthcareProviders --|> ProviderRegistration
    HealthcareProviders  --> HealthcareProvidersGroup
    Blood --> HealthcareProviders
    Medication --> HealthcareProviders 
[...]

PR #1179, opened 2024-04-28
feat: migrate part of existing medical provider under specific healthcare providers (#1137)

filipowm:
This is initial PR to migrate existing `Medical` provider. The next one will be to migrate `BloodType` and `Disease` providers. Then I will move on adding new ones.

I marked `Medical` provider as deprecated since 2.4.0, yet I'm not sure whether it shouldn't be 2.3.0 or some other.


src/main/java/net/datafaker/providers/healthcare/CareProvider.java
```diff
@@ -0,0 +1,28 @@
+package net.datafaker.providers.healthcare;
+
+import net.datafaker.providers.base.AbstractProvider;
+
+/**
+ * Generate random types of health care providers
```
erikpragt-connectid, line 6: All providers require a "since" property, so I think a since 2.3.0 would be good, even though we might release this as part of 2.2.3, not sure yet.

src/main/java/net/datafaker/providers/base/Medical.java
```diff
@@ -30,11 +45,19 @@ public String diagnosisCode() {
...
+     * @deprecated since 2.4.0. For removal in 3.0.0 version. Use {@link net.datafaker.providers.healthcare.MedicalProcedure#icd10()} instead.
+     */
+    @Deprecated(since = "2.4.0", forRemoval = true)
     public String procedureCode() {
         String regex = resolve("medical.procedure_code.icd10");
         return faker.regexify(regex);
     }
 
+    /**
+     * @deprecated since 2.4.0. For removal in 3.0.0 version. Use {@link net.datafaker.providers.healthcare.CareProvider#medicalProfession()} instead.
```
erikpragt-connectid, line 58: I probably wouldn't add the "for removal in". The 3.0.0 can be released next week, or it can be done 5 years from now, that's hard to say, we usually keep these things around 1-2 released, and then remove them.

src/test/java/net/datafaker/providers/healthcare/MedicalProcedureTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class MedicalProcedureTest extends HealthcareFakerTest {
+
+    private final MedicalProcedure medicalProcedure = getFaker().medicalProcedure();
+
+    @RepeatedTest(100)
+    void testProcedureCodes() {
+        // will use icd-10-pcs - https://regex101.com/library/nJ1wC4
```
erikpragt-connectid, line 13: There is icd10-pcs, there's icd10-cm, and there's also icd10-am. Maybe it would be good to be explicit in the `icd10()` method which one we mean?

src/main/java/net/datafaker/providers/healthcare/CareProvider.java
filipowm: Got it. I added it in all new classes and interface.

src/main/java/net/datafaker/providers/base/Medical.java
filipowm: Ok, adjusted.

src/test/java/net/datafaker/providers/healthcare/MedicalProcedureTest.java
filipowm: To clarify it -- in the context here, that is `MedicalProcedure.icd10()`, we're testing for `icd-10-pcs` regex.

In future there would be `Disease.icd10()` which would test for `icd-10` regex (which is the same for all local variants).

bodiam: Ah yes, that makes sense now, sorry, I missed the class name, all good!

src/main/java/net/datafaker/providers/healthcare/CareProvider.java
```diff
@@ -0,0 +1,30 @@
...
+ * @since 2.3.0
+ */
+public class CareProvider extends AbstractProvider<HealthcareProviders> {
+    protected CareProvider(HealthcareProviders faker) {
+        super(faker);
+    }
+
+    /**
+     * Generate random hospital name
+     * @return A hospital name
```
snuyanzin, line 17: Why do we need javadoc which just repeats the actual method name here and in the rest?
IMHO: if javadoc can not bring more info than existing name then probably we don't need that javadoc

src/main/java/net/datafaker/providers/base/Medical.java
```diff
@@ -2,13 +2,20 @@
 
 /**
  * @since 0.8.0
+ * @deprecated since 2.3.0. This faker is deprecated due to migration
+ * to healthcare-specific aggregated into {@link net.datafaker.providers.healthcare.HealthcareFaker} fakers with similar methods.
  */
+@Deprecated(since = "2.3.0", forRemoval = true)
```
snuyanzin, line 8: If we are going to remove it
what is the replacement for `Medical#diseaseName` ?

src/main/java/net/datafaker/providers/healthcare/CareProvider.java
kingthorin: Agreed

src/main/java/net/datafaker/providers/base/Medical.java
filipowm: I will add it in next PR, cause it touches `Disease` which I mentioned.

src/main/java/net/datafaker/providers/healthcare/CareProvider.java
filipowm: Yeah, I agree too. However existing codebase is inconsistent, cause in some places there are such dummy javadocs, in other not. So I preferred to add it cause I didn't know which one you prefer.

So I will remove those.

bodiam:
I think this PR is good to go right? Any objections if I merge it?

kingthorin (approved):
LGTM

bodiam:
Thanks @filipowm , it's merged!

PR #1180, opened 2024-04-28
POC for 1178

snuyanzin:
There is still a couple of tests which should be handled


src/main/resources/en/internet.yml
```diff
@@ -15,7 +15,7 @@ en:
       - "Visit #{Country.name}"
       - "Want to play a game?"
       - "Since we can’t all win the lottery..."
-      - "#{FamousLastWords}"
+      - "#{FamousLastWords.lastWords}"
```
snuyanzin, line 18: This is a dangerous usage
it could potentially change the behavior once there will be more methods added
that's why there should be a specific method mentioned

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
```diff
@@ -149,60 +151,33 @@ void resolveKeyToPropertyWithAPropertyWithoutAnObject() {
         verify(dummy).hello();
     }
 
-    @Test
-    void resolveKeyToPropertyWithAPropertyWithAnObject() {
```
snuyanzin, line 153: mocked tests is painful to maintain
at the same time there are several normal (non mocked tests) covered this functionality
e.g. tests with `InsectFromFile`

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
```diff
@@ -149,60 +151,33 @@ void resolveKeyToPropertyWithAPropertyWithoutAnObject() {
...
-        // when
-        final String actual = fakeValuesService.resolve("property.resolutionWithList", dummy, mockedFaker, mockedFaker.getContext());
-
-        // then
-        assertThat(actual).isEqualTo("Yo!");
-        verify(dummy).hello();
-    }
-
-    @Test
-    void resolveKeyWithMultiplePropertiesShouldJoinResults() {
```
snuyanzin, line 188: mocked tests is painful to maintain
at the same time there are several normal (non mocked tests) covered this functionality
e.g. tests with `InsectFromFile`

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
```diff
@@ -149,60 +151,33 @@ void resolveKeyToPropertyWithAPropertyWithoutAnObject() {
...
-        assertThat(actual).isEqualTo("Luke Cage");
-        verify(mockedFaker).superhero();
-        verify(person).name();
-    }
-
     @Test
     void resolveKeyToPropertyWithAList() {
         // property.resolutionWithList -> #{hello}
         // #{hello} -> DummyService.hello
-
```
snuyanzin, line 173: mocked tests is painful to maintain
this one is rewritten in a way without mocks

bodiam:
@snuyanzin do you think this one is ready? I'm happy to try it out in my project by using the snapshot.

snuyanzin:
@bodiam it looks like that
at least most of the times its green

 would be great to get some feedback from downstream rojects

PR #1187, opened 2024-05-03
Small renames in the code.

bodiam:


snuyanzin (approved):

src/main/java/net/datafaker/service/FakeValuesContext.java
```diff
@@ -58,40 +58,41 @@ public static FakeValuesContext of(Locale locale, String filename, String path,
...
+        return singletonLocale.getLocale();
     }
 
     public String getFilename() {
         return filename;
     }
 
-    public Set<String> getPath() {
-        return pathes;
+    public Set<String> getPaths() {
```
kingthorin, line 90: I'm m not against changing this but it is technically a breaking change to the API.

(Renaming public methods others may depend on.)

kingthorin: Same with the setter

bodiam: That's a fair comment. Is this part of our public api? I can deprecate this one and introduce the second method, that's no problem.

kingthorin: That's safest. Tag the existing one deprecated for removal. It can even then simply call the new method(s) (to avoid duplication).

snuyanzin: I guess with `FakeValueContext` it is still kind of ok since there is not so many cases 1if not none.
And the cases are very advanced (even not covered in doc =), probably another point to improve)
So if there is any user who were able to use datafaker for such use case, i'm pretty sure that method rename isnot a problem then...

however agree, if we are talking about much stricter approach then may be ...

bodiam: I also think the chance of breaking is probably low. I'll just merge it as is.

PR #1188, opened 2024-05-04
Migrate disease provider under healthcare providers (#1137)

filipowm:
Hi,

I started migrating `Disease` provider under `healthcare`, however I'm not sure if this wouldn't break backward compatibility, because:
1. `Disease` is moved from `BaseProviders` to `HealthcareProviders`
2. `disease.yml` was changed to be prepended keys with `healthcare` prefix.

What are your thoughts on this? Is this approach safe enough to move on?


bodiam:
To me, though I only had a quick look and didn't test it myself, this looks very good. I don't think the API is broken (maybe we should have a library API tester?), since the decease only moved, and the public API seems quite similar, so on a quick glance, it seems the API hasn't been broken.

kingthorin:
Completely changing the location of methods is API breaking. Users have to re-code things to make them work again. While it is minimal effort to adapt it's still a breaking change.

src/main/java/net/datafaker/providers/healthcare/Disease.java
```diff
@@ -0,0 +1,113 @@
...
+    public String anyDisease() {
+        List<Supplier<String>> providers = List.of(
+            this::internalDisease,
+            this::neurology,
+            this::surgery,
+            this::paediatrics,
+            this::gynecologyAndObstetrics,
+            this::ophthalmologyAndOtorhinolaryngology,
+            this::dermatology
+        );
```
snuyanzin, line 45: I don't see a reason to create this list on each method call, it could be extracted in a const imho

src/main/java/net/datafaker/providers/healthcare/Disease.java
```diff
@@ -0,0 +1,113 @@
...
+ */
+public class Disease extends AbstractProvider<HealthcareProviders> {
+
+    static final String INTERNAL_DISEASE_KEY = "healthcare.disease.internal_disease";
+    static final String NEUROLOGICAL_DISEASE_KEY = "healthcare.disease.neurology";
+    static final String SURGICAL_DISEASE_KEY = "healthcare.disease.surgery";
+    static final String PAEDIATRIC_DISEASE_KEY = "healthcare.disease.paediatrics";
+    static final String GYNECOLOGY_AND_OBSTETRICS_DISEASE_KEY = "healthcare.disease.gynecology_and_obstetrics";
+    static final String OPHTHALMOLOGY_AND_OTORHINOLARYNGOLOGY_DISEASE_KEY = "healthcare.disease.ophthalmology_and_otorhinolaryngology";
+    static final String DERMATOLOGY_DISEASE_KEY = "healthcare.disease.dermatology";
```
snuyanzin, line 21: Probably it's better to use enum for this

src/test/java/net/datafaker/providers/healthcare/DiseaseTest.java
```diff
@@ -0,0 +1,58 @@
...
+            TestSpec.of(disease::dermatology, DERMATOLOGY_DISEASE_KEY));
+    }
+
+    @RepeatedTest(100)
+    void testAnyDisease() {
+        // given
+        Set<String> allDiseases = Stream.of(INTERNAL_DISEASE_KEY, NEUROLOGICAL_DISEASE_KEY, SURGICAL_DISEASE_KEY, PAEDIATRIC_DISEASE_KEY, GYNECOLOGY_AND_OBSTETRICS_DISEASE_KEY, OPHTHALMOLOGY_AND_OTORHINOLARYNGOLOGY_DISEASE_KEY, DERMATOLOGY_DISEASE_KEY)
+            .map(this::getBaseList)
+            .flatMap(Collection::stream)
+            .collect(Collectors.toSet());
```
snuyanzin, line 41: if we use repeated test, there is no need to create same list for each test attempt

bodiam:
> Completely changing the location of methods is API breaking. Users

I might be missing something, but I don't see it? They will still do `faker.disease().icd10()`, why would any user care if the `disease()` is part of the BaseProviders or HealthcareProviders, it's still the same API call.

kingthorin:
I'll have to pull the branch and play with it.

If, as you've said, the methods are still accessed the same way then I totally agree it's non-breaking.

I took bullet one from the PR description to mean that access had changed.

kingthorin:
👂 I'm being brain dead. The unit tests would have been different if there were API changes. I know they did move but the faker usage is the same. So ignore me 😔

src/main/java/net/datafaker/providers/healthcare/Disease.java
filipowm: Done

src/test/java/net/datafaker/providers/healthcare/DiseaseTest.java
filipowm: Done

src/main/java/net/datafaker/providers/healthcare/Disease.java
bodiam: I don't mind if you keep it as is, these are mostly (only?) internal things anyway, we can always refactor it later.

snuyanzin: if you use enum, then you don't need `private final List<Supplier<String>> allDiseaseProviders` at all
and `anyDisease` could be implemented as `Options.option..`

And each time there will be a new decease, then with current approach you need to add noth string var for this decease + maintain `allDiseaseProviders`. With enums you just need to update enums...
Thus finally less code

filipowm: Hm, I need to play with that, cause `Options.option` is new to me. Thanks for that hint, sounds reasonable now.

kingthorin:
I haven't looked but is this PR branched after that fix or does it need rebasing?

filipowm:
> I haven't looked but is this PR branched after that fix or does it need rebasing?

@kingthorin I branched out on saturday, so it might require rebasing. I will do that in the evening.

bodiam:
Hi all, what should we do with this PR?

snuyanzin:
I guess addressing of this comment is still in progress
@filipowm do you need help with it?

src/main/java/net/datafaker/providers/healthcare/Disease.java
bodiam: Hi all, hope you don't mind, but I've addressed these comments in a commit on this PR, I didn't want to let these efforts go to waste. Thanks @filipowm for this improvement. @snuyanzin , can you please have another look at the PR?

snuyanzin (approved):

PR #1189, opened 2024-05-06
Added locations.

bodiam:
I took them from here: https://improwiki.com/en/wiki/improv/list_of_places

PR #1192, opened 2024-05-08
Add further locations

kingthorin:
Most of these are just from brain storming.

Follow-up to: datafaker-net/datafaker#1189


bodiam:
Looks great, let's merge!

PR #1196, opened 2024-05-18
show actual values in error message

asolntsev:
... instead of just throwing an empty runtime exception.

### Problem
Before this change, test might fail with an unclear message:

```java
java.lang.RuntimeException
	at net.datafaker.Issue759.issue759Test(Issue759.java:93)
```

or the output of multiple failing tests might be mixed:
```java
 0 0 0 0 0
 0 0 0 0 0
 0 0 0 0 0 0
 0 0 0 0 0 0 0 0 0 0 0 0 0 0
 0 0 0 0 0
 0 0 0
 0 0 0
 0 0 0
 0

 0 0 0 0 0

java.lang.RuntimeException
	at net.datafaker.Issue759.issue759Test(Issue759.java:93)
```

### Solution
Now the error message is clear:
```java
java.lang.AssertionError: Not all of s[0, 0, 0, 0, 0] are not equal to 20000

	at net.datafaker.Issue759Test.issue759Test(Issue759Test.java:85)
```

### P.S.
To reproduce the issue, just put breakpoint, say, to any field of `WorkerThread`, run `Issue759` test in debug mode and resume the execution after stopping on the breakpoint.

### DIFF
--- a/src/test/java/net/datafaker/Issue759Test.java
+++ b/src/test/java/net/datafaker/Issue759Test.java
@@ -59,7 +59,7 @@ class Issue759Test {
     @RepeatedTest(10)
     void issue759Test() throws InterruptedException {
         final int numThreads = 5;
-        final int iterationsPerThread = 20000;
+        final int iterationsPerThread = 60000;
 
         final Faker faker = new Faker();
 
@@ -73,16 +73,23 @@ class Issue759Test {
         }
 
         int[] lastIters = getIterations(threads);
+        final int delayMs = 100;
+        final int maxIntervalWithoutChangesMs = 1000;
+        int intervalWithoutChanges = 0;
         while (true) {
-            Thread.sleep(100);
+            Thread.sleep(delayMs);
             int[] iters = getIterations(threads);
             if (Arrays.equals(lastIters, iters)) {
                 // Either all threads are done, or something is probably stuck
-                System.err.println();
                 if (allElementsEqual(iters, iterationsPerThread)) {
                     break;
+                } else if (Arrays.equals(lastIters, iters)) {
+                    intervalWithoutChanges += delayMs;
+                    if (intervalWithoutChanges > maxIntervalWithoutChangesMs) {
+                        throw new AssertionError("Not all of %s are equal to %s".formatted(Arrays.toString(iters), iterationsPerThread));
+                    }
                 } else {
-                    throw new AssertionError("Not all of %s are equal to %s".formatted(Arrays.toString(iters), iterationsPerThread));
+                    intervalWithoutChanges = 0;
                 }
             }
             lastIters = iters;
--- a/src/test/java/net/datafaker/Issue759Test.java
+++ b/src/test/java/net/datafaker/Issue759Test.java
@@ -2,29 +2,28 @@ package net.datafaker;
 
 import org.junit.jupiter.api.RepeatedTest;
 
-import java.util.Arrays;
+import java.util.concurrent.CountDownLatch;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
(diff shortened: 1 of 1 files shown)


src/test/java/net/datafaker/Issue759Test.java
```diff
@@ -89,8 +82,7 @@ void issue759Test() throws InterruptedException {
                 if (allElementsEqual(iters, iterationsPerThread)) {
                     break;
                 } else {
-                    printIterations(iters);
-                    throw new RuntimeException();
+                    throw new AssertionError("Not all of s%s are not equal to %s".formatted(Arrays.toString(iters), iterationsPerThread));
```
bodiam, line 85: Sorry, this error message, it's doing my head in. Not all of %s are not equal to .... 

Does this mean: `Not all of %s are different to %s`, or something else?

asolntsev: Ooops, one "s" and one "not" were not needed.

The intended variant was:
> Not all of [1,2,3,4,3,6,3,7] are equal to 3.

But we can simplify it, of course.

snuyanzin:
Thanks for your contribution @asolntsev 

it seems there was a long standing issue with this test, which apparently started to reproduce with your PR [1] on Windows only...
I submitted a fix for that (a bit more description is below)
If you are ok with it, we can merge it, otherwise please express your concerns

About the issue:
It looks the problem was with the code 
```java
        int[] lastIters = getIterations(threads);
        while (true) {
            Thread.sleep(100);
            int[] iters = getIterations(threads);
            if (Arrays.equals(lastIters, iters)) {
                // Either all threads are done, or something is probably stuck
                System.err.println();
                if (allElementsEqual(iters, iterationsPerThread)) {
                    break;
                } else {
                    printIterations(iters);
                    throw new RuntimeException();
                }
            }
```
On Windows it could happen that we reach the line `if (allElementsEqual(iters, iterationsPerThread)) {` before the threads are even started... as a result there was none iterations executed...
The idea of fix is to measure amount of time of being unchanged and react on the fact that this amount of time is exceeded some limit.

src/test/java/net/datafaker/Issue759Test.java
bodiam: Ah, that reads a lot better indeed! Thanks for improving this!

asolntsev:
@snuyanzin Thank you for sharing the information.
Now I looked at the test once more and realized it can be simplified using Java built-in class `CountDownLatch`.
Look my last commit. :)

src/test/java/net/datafaker/Issue759Test.java
```diff
@@ -0,0 +1,62 @@
...
+            assertThat(county).isNotEqualTo(zipCode);
+        } catch (Exception ignore) {
+        }
+    }
+
+    @RepeatedTest(10)
+    void issue759Test() throws InterruptedException {
+        int numThreads = 5;
+        int iterationsPerThread = 20000;
+        CountDownLatch countDownLatch = new CountDownLatch(numThreads * 20000);
```
snuyanzin, line 45: nit: i guess we could reuse the second var as well
```suggestion
        CountDownLatch countDownLatch = new CountDownLatch(numThreads * iterationsPerThread);
```

snuyanzin:
@asolntsev Indeed it is better approach
Thanks a lot!
It look great

I left one tiny comment

src/test/java/net/datafaker/Issue759Test.java
asolntsev: Oops, thank you. Fixed. @snuyanzin

snuyanzin (approved):

PR #1197, opened 2024-05-18
add Estonian id number generator

asolntsev:


snuyanzin:
Great to see the number of locales increasing
thanks for your contribution @asolntsev 
it looks good to me

Could you please also add the locale to README as others at https://github.com/datafaker-net/datafaker/blob/91aaac7c7e5ff2496d99861971dd915e248006e1/README.md?plain=1#L463
?

asolntsev:
@snuyanzin Sure, I've added "et" to README.

snuyanzin (approved):

asolntsev:
@snuyanzin Oops, now I realized that the project already had Estonian support. 
But it was in file [ee.yml](https://github.com/datafaker-net/datafaker/blob/main/src/main/resources/ee.yml). While I added another file [et.yml](https://github.com/datafaker-net/datafaker/blob/main/src/main/resources/et.yml).

So what is the file name: country code or language code?
For Estonia, "ee" is country code, and "et" is language code.

snuyanzin:
yep, sorry I should have checked that before...

there are 2 ways
1. just language code, in that case it should be `et`
2. language code - country code e.g `en-CA.yml`, `en-GB.yml`, `en-IND.yml`, `en-MS.yml`

asolntsev:
@snuyanzin Ok, then I will create a PR for merging "et.yml" and "ee.yml" into one file?

snuyanzin:
yep, I think that would make sense
+ may be update `README.md` if it is required after the merge

asolntsev:
Cool :)

Here it is: https://github.com/datafaker-net/datafaker/pull/1206

Issue #1190, opened 2024-05-07
Intermitted expressions failures during testing


bodiam:
**Describe the bug**

It seems there's another or new bug in our tests:

```
 10:28:42.078 [ERROR] Failures: 
10:28:42.078 [ERROR] net.datafaker.providers.base.RestaurantTest.name
10:28:42.078 [ERROR]   Run 18: RestaurantTest.name:33 
Expecting actual:
  "## Bakery"
not to contain:
  ["#", "?"]
but found:
  ["#"]
```

**To Reproduce**

I cannot.

**Expected behavior**

The test should pass

**Versions:**
 - Faker Version 2.2.x-snapshot

snuyanzin:
That looks strange, it looks like sometimes it doesn't take into account possible expressions inside values.
I'm not sure whether it is COW, currently it doesn't look like that

bodiam:
I've ran the test locally quite a few times, and I'm not able to reproduce it, but I will have some time over the weekend to look into this.

snuyanzin:
If I'm not mistaken @RVRhub is also looking at it, please correct me if I'm wrong

RVRhub:
Yes, I started looking into this problem.
I have determined that this is because sometimes resolveExpression uses a Supplier that retrieves data from a file.

snuyanzin:
@bodiam it might happen that @RVRhub has fixed it within #1198  (not 100%)
however now it looks much more stable, so if you know how to reproduce it would worth to check it against the latest main

bodiam:
@snuyanzin no idea how to reproduce it, I just know it happens sometimes. Let's monitor it for a week or so, and then close this issue if it's fixed.

PR #1198, opened 2024-05-20
Modify the enrichment provider classes map in BaseFaker 

RVRhub:
PR Modify the enrichment provider classes map in BaseFaker to store a complete list of generated FakerContext.
: #1190

This allows you to store multiple locals for one class in the net.datafaker.providers.base.BaseFaker#CLASSES map.
Before this, every call to the net.datafaker.providers.base.BaseFaker#getProvider(java.lang.Class<AP>, java.util.function.Function<PR,AP>, PR) method would overwrite the map for one class name, making the cache ineffective.


snuyanzin (approved):
Thanks for the contribution @RVRhub 

I've run about 25 times (each run contains 9 jobs) the ci and all the runs are passed
https://github.com/datafaker-net/datafaker/actions/runs/9165081403
So it looks pretty stable

PR #1203, opened 2024-05-22
[refactoring] simplify COWMap

asolntsev:
I have no idea what "COW" means, but it seems we just need an old good synchronized map (provided by Java SDK).

In future, we could even replace all usages of COWMap by `synchronizedMap`.

### DIFF
--- a/src/main/java/net/datafaker/internal/helper/COWMap.java
+++ b/src/main/java/net/datafaker/internal/helper/COWMap.java
@@ -5,13 +5,13 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Supplier;
 
+import static java.util.Collections.synchronizedMap;
+
 public class COWMap<K, V> implements Map<K, V> {
-    private volatile Map<K, V> map;
-    private final Supplier<Map<K, V>> mapSupplier;
+    private final Map<K, V> map;
 
     public COWMap(Supplier<Map<K, V>> mapSupplier) {
-        this.mapSupplier = mapSupplier;
-        this.map = mapSupplier.get();
+        this.map = synchronizedMap(mapSupplier.get());
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


kingthorin:
I believe COW == CopyOnWrite

snuyanzin (changes requested):
Thanks for looking into that
IIRC the idea here was that these maps are filled only once per provider in case the provider is in use. After that the provider related info is only read from those maps. For that reason copy on write is used here.

The problem with synchronized map is that they make things drastically slower

I made some basic measurements with help of https://github.com/datafaker-net/datafaker-benchmark
like
```
mvn clean package exec:exec -Dbenchmarks="DatafakerSimpleMethods" -Ddatafaker.version=2.2.3-SNAPSHOT
```

before this PR
```
DatafakerSimpleMethods.firstname      thrpt   10   4782.533 ±  20.500  ops/ms
DatafakerSimpleMethods.fullname       thrpt   10   1697.407 ±  25.530  ops/ms
DatafakerSimpleMethods.numberBetween  thrpt   10  45964.335 ± 237.500  ops/ms
DatafakerSimpleMethods.streetAddress  thrpt   10   1246.463 ±  21.650  ops/ms
```

After
```
DatafakerSimpleMethods.firstname      thrpt   10   3465.447 ±  78.319  ops/ms
DatafakerSimpleMethods.fullname       thrpt   10   1270.090 ±  38.539  ops/ms
DatafakerSimpleMethods.numberBetween  thrpt   10  45742.122 ± 624.010  ops/ms
DatafakerSimpleMethods.streetAddress  thrpt   10    974.074 ±  49.535  ops/ms
```

and it looks like it starts being up to 30% slower...

asolntsev:
@snuyanzin Wow, interesting... Thank you for sharing the information.
Well, probably this information could be added to COWMap javadoc?

bodiam:
@asolntsev yes, and probably renaming that class to CopyOnWriteMap wouldn't hurt either, you're not the first one to ask that it is.

snuyanzin:
> Well, probably this information could be added to COWMap javadoc?
> and probably renaming that class to CopyOnWriteMap wouldn't hurt either, you're not the first one to ask that it is.

yes, that probably makes sense, done  at https://github.com/datafaker-net/datafaker/pull/1207

asolntsev:
@bodiam No problems. In fact, we should rather delete this PR because it was replaced by https://github.com/datafaker-net/datafaker/pull/1207

PR #1204, opened 2024-05-22
Fix flaky tests

asolntsev:


snuyanzin (approved):
Thanks for improving it
LGTM

PR #1205, opened 2024-05-23
catch only specific exceptions

asolntsev:
...instead of catching too generic exceptions (which is considered a bad practice because some unexpected problems might be ignored).


kingthorin (approved):
Looks good to me. Thanks!

snuyanzin (approved):

PR #1208, opened 2024-05-24
add Albanian id number generator

asolntsev:


src/main/java/net/datafaker/idnumbers/AlbanianIdNumber.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public String getInvalid(BaseProviders faker) {
+        String pin = getValid(faker);
+        int invalidMonth = faker.number().numberBetween(93, 99);
+        return pin.substring(0, 2) + invalidMonth + pin.substring(4);
+    }
+
+    public String getValid(BaseProviders faker) {
+        LocalDate birthDate = faker.date().birthdayLocalDate(0, 200);
+        boolean female = Math.random() >= 0.5;
```
snuyanzin, line 22:
```suggestion
        boolean female = faker.bool().bool();
```
This seems to be the problem and the reason for failing tests.
Especially `FakerRepeatabilityIntegrationTest`. The idea is that datafaker allows to have repeatable execution which could be helpful to reproduce some test failure. Among others faker has input arg seed or `Random` with seed and for the same seed and same sequence of actions the output will be same. This is what `FakerRepeatabilityIntegrationTest` is checking for. 
As a result we can not use `Math.random()` or some other similar code where we are not the master of seed

bodiam:
@asolntsev Thank for this contribution, and thanks @snuyanzin for giving feedback. I've merged the feedback in, and I'll merge it.

src/main/java/net/datafaker/idnumbers/AlbanianIdNumber.java
snuyanzin: yeah, there is a number of tools allowing such checks, e.g. https://github.com/policeman-tools/forbidden-apis
also allows to do that by specifying of methods which are not allowed in config and failing while compilation
example of usage  in Calcite https://github.com/apache/calcite/blob/main/src/main/config/forbidden-apis/signatures.txt

Feel free to submit a PR for that
I'm ok with any tool if it allows easily add more exceptions once we have such

snuyanzin: One drawback. of archunit I faced in the past
sometimes it generates different set of violations depending on java bytecode
https://github.com/TNG/ArchUnit/issues/1124

bodiam: I'm not for or against something like ArchUnit, but I don't really care if you Math.random() or not, as long as the execution of Datafaker is repeatable. And if you can do that using Math.random(), all good. 

I think that in effect, the repeatable test case covers things already, and ArchUnit is a bit too low level for me personally (it gives me same feeling as mocking, which I'm not a fan of either). But if you see value, and are willing to submit a PR, happy to try it out here.

asolntsev: To be honest, I don't understand why are you curious about repeatability.
When a test fail, error message should always explicitly say what happened:
```java
java.lang.AssertionError: 
Expecting actual:
  "38809266059"
to match pattern:
  "ZZZ[1-6][0-9]{10}"
```

So it's not needed to re-run tests in the same order. I already see from the error message what was the wrong value. And I can re-run Faker with this specific value.

kingthorin: Repeatability in use not necessarily in tests. In the faker you (as a user/dev/implementor) can control the seed so that you get repeatable results. If Math.random is used the library no longer controls the seed.

snuyanzin: >I don't understand why are you curious about repeatability.
When a test fail, error message should always explicitly say what happened:

it is clear for simple junit test with more or less deterministic input.
The problem could be with floating issues. And here I'm talking not about datafaker as a user of it, I'm talking about projects who uses datafaker as a thirdparty lib for their tests. In case they've got a failure with help of datafaker they should be able to reproduce it again without extra dancing

Issue #1183, opened 2024-05-01
Breaking DateAndTime


bodiam:
Hi all,

Another proposal: I'm thinking of breaking DateAndTime.

At the moment, it's using `java.sql.Timestamp` for `future()`, `past()` and `birthday()` methods. I don't think anyone would use a SQL Timestamp for these kind of things, and I'm thinking of changing it. I'm not sure how to deprecate them since I want to change the return type, but keep the method names and the provider it's in (I could make a new one and deprecate this one, but I'm not sure how to call it. `TimeAndDate`? `DateAndTime2`? I also plan on removing all references to `java.util.Date`. 

The plan is to make `birthday()` return a LocalDate, and `future` and `past` I'm not sure of yet. Perhaps a `ZonedDateTime`, or an `OffsetDateTime`?

Suggestions welcome, if no objections, I will implement the above.

kingthorin:
I agree that getting rid of `java.sql.Timestamp` usage is likely a good move. Is there a reason to move away from `java.util.Date`? (Not that I'm against it, I just want to understand the direction a bit better.)

Both `ZonedDateTIme` and `OffSetDateTime` support `plus` and `minus` methods so that seems like a simple way to handle `future` and `past`.

bodiam:
I think `java.util.Date` is not something anyone should use with the introduction of `java.time`, and that's now more than 10 years ago! Also, Gavin King says you shouldn't: https://in.relation.to/2024/04/22/stop-using-date/

Also:

![image](https://github.com/datafaker-net/datafaker/assets/46468/27e9a75d-be80-47bc-a75f-976e4d1665b8)

(source: https://stackoverflow.com/questions/32437550/whats-the-difference-between-instant-and-localdatetime/32443004#32443004)

So, possibly, we should be using `Instant` instead.

kingthorin:
Thanks, great resources. Makes sense.

asolntsev:
I guess this issue can be closed because it was implemented in https://github.com/datafaker-net/datafaker/pull/1210 ?

PR #1210, opened 2024-05-25
Migrated DateAnTime to TimeAndDate to support Java8 Time, and depreca…

bodiam:
…te Timestamps and java.util.Date code.

Should address https://github.com/datafaker-net/datafaker/issues/1183

### DIFF
--- a/src/main/java/net/datafaker/providers/base/TimeAndDate.java
+++ b/src/main/java/net/datafaker/providers/base/TimeAndDate.java
@@ -93,11 +93,8 @@ public class TimeAndDate extends AbstractProvider<BaseProviders> {
      * @return a future date relative to {@code referenceDate}.
      */
     public Instant future(int atMost, TimeUnit unit, Instant referenceDate) {
-        long upperBound = unit.toMillis(atMost);
-
-        long futureMillis = referenceDate.toEpochMilli();
-        futureMillis += 1 + faker.random().nextLong(upperBound - 1);
-
+        long upperBoundMillis = unit.toMillis(atMost);
+        long futureMillis = referenceDate.toEpochMilli() + 1 + faker.random().nextLong(upperBoundMillis - 1);
         return Instant.ofEpochMilli(futureMillis);
     }
 
@@ -181,14 +178,10 @@ public class TimeAndDate extends AbstractProvider<BaseProviders> {
      * @return a past date relative to {@code referenceDate}.
      */
     public Instant past(int atMost, TimeUnit unit, Instant referenceDate) {
-        long upperBound = unit.toMillis(atMost);
-
-        long futureMillis = referenceDate.toEpochMilli();
-        futureMillis -= 1 + faker.random().nextLong(upperBound - 1);
-
-        return Instant.ofEpochMilli(futureMillis);
+        long upperBoundMillis = unit.toMillis(atMost);
+        long pastMillis = referenceDate.toEpochMilli() - 1 - faker.random().nextLong(upperBoundMillis - 1);
+        return Instant.ofEpochMilli(pastMillis);
     }
-
     /**
      * Generates a string representation of a past date relative to the {@code referenceDate}.
      *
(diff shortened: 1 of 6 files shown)


src/main/java/net/datafaker/providers/base/TimeAndDate.java
```diff
@@ -0,0 +1,344 @@
...
+     * @param atMost        at most this amount of time past to the {@code referenceDate} exclusive.
+     * @param unit          the time unit.
+     * @param referenceDate the past date relative to this date.
+     * @return a past date relative to {@code referenceDate}.
+     */
+    public Instant past(int atMost, TimeUnit unit, Instant referenceDate) {
+        long upperBound = unit.toMillis(atMost);
+
+        long futureMillis = referenceDate.toEpochMilli();
+        futureMillis -= 1 + faker.random().nextLong(upperBound - 1);
```
kingthorin, line 187: pastMillis?

kingthorin (approved):

kingthorin:
Ugh, just one minor comment. Apparently I was half asleep and set approve anyway 🤷‍♂️

src/main/java/net/datafaker/providers/base/TimeAndDate.java
bodiam: Good point. Will fix!

bodiam: Fixed.

bodiam:
> Ugh, just one minor comment. Apparently I was half asleep and set approve anyway 🤷‍♂️

Appreciate the approval, I fixed the naming, and a few other minor improvements. If you want to have another look, that would be great!

src/main/java/net/datafaker/providers/base/TimeAndDate.java
```diff
@@ -0,0 +1,334 @@
...
+
+    /**
+     * Generates a random birthday between two ages from now.
+     *
+     * @param minAge the minimal age
+     * @param maxAge the maximal age
+     * @return a random birthday between {@code minAge} and {@code maxAge} years ago from now.
+     * Negative {@code minAge} and {@code maxAge} are supported.
+     */
+    public LocalDate birthday(int minAge, int maxAge) {
```
kingthorin, line 259: Make sure they're not negative or zero?

src/test/java/net/datafaker/providers/base/TimeAndDateTest.java
```diff
@@ -0,0 +1,229 @@
...
+        Instant future = timeAndDate.future(1, TimeUnit.SECONDS, now);
+        assertThat(future).isBetween(now, now.plusSeconds(1));
+    }
+
+    @RepeatedTest(1000)
+    void testFutureDateWithMinimum() {
+        Instant now = Instant.now();
+        Instant future = timeAndDate.future(5, 4, TimeUnit.SECONDS);
+        assertThat(future)
+            .isBetween(now.plusMillis(3500), now.plusMillis(5500));
```
kingthorin, line 43: Make constants for these magic numbers or add a comment that clarifies. I'm sure it makes sense if I read the code but less digging is better 😉

src/test/java/net/datafaker/providers/base/TimeAndDateTest.java
```diff
@@ -0,0 +1,229 @@
...
+        assertThatThrownBy(() -> timeAndDate.between(then, now))
+            .isInstanceOf(IllegalArgumentException.class)
+            .hasMessage("Invalid date range, the upper bound date is before the lower bound.");
+    }
+
+    @RepeatedTest(1000)
+    void testBirthday() {
+        final LocalDateTime now = LocalDateTime.now();
+        final LocalDate from = now.minusYears(18).toLocalDate();
+        final LocalDate to = now.minusYears(65).toLocalDate();
```
kingthorin, line 111: Use the constants from the new class?

kingthorin (commented):
Sorry found a few more minor things. Feel free to merge after these 😁

src/test/java/net/datafaker/providers/base/TimeAndDateTest.java
bodiam: I agree, but I mostly ported the old code to the new types. I'll have a look later.

bodiam:
@kingthorin I've merged this PR, but I will take your feedback into account. Only reason I merged is so I can ask @asolntsev to use this faker instead in the other PR.

src/main/java/net/datafaker/providers/base/TimeAndDate.java
bodiam: > Make sure they're not negative or zero?

Not sure, since "Negative {@code minAge} and {@code maxAge} are supported.". ?

kingthorin: Okay, wasn't sure that was intentional.

src/test/java/net/datafaker/providers/base/TimeAndDateTest.java
bodiam: This is actually great feedback. I'm reading this code now, and to me, it doesn't really make sense. Why are 5 and 4 flipped around? It would make more sense to me to have a future date between 4 and 5 seconds into the future, not between 5 and 4. What do you think? (the challenge is that the behaviour of this method would be slightly different from the old one, so maybe this should throw an exception if argument 2 is lower than argument 1? Thoughts?

kingthorin: Hahaha I was actually on about the 3500 & 5500. 

But ya I agree changing the item you're talking about seems better too and enforcing the condition as well.

PR #1213, opened 2024-05-25
Add Bulgarian id number

asolntsev:


src/main/java/net/datafaker/idnumbers/BulgarianIdNumber.java
```diff
@@ -0,0 +1,62 @@
...
+        return basePart + checksum(basePart);
+    }
+
+    public String getInvalid(BaseProviders faker) {
+        String basePart = basePart(faker);
+        return basePart + (checksum(basePart) + 1) % 10;
+    }
+
+    private String basePart(BaseProviders faker) {
+        LocalDate birthDate = faker.date().birthdayLocalDate(0, 200);
```
bodiam, line 26: Would it be an option to use the new TimeAndDate provider here? Also, why can someone's age be 199 or 200?

PR #1216, opened 2024-05-25
Fix flaky test `TimeAndDateTest.testPastDate`

asolntsev:
Also, add details to thrown errors.

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Locality.java
+++ b/src/main/java/net/datafaker/providers/base/Locality.java
@@ -102,8 +102,8 @@ public class Locality extends AbstractProvider<BaseProviders> {
         for (String s: paths) {
             try {
                 Files.walkFileTree(Paths.get(s).toAbsolutePath(), visitor);
-            } catch (Exception e) {
-                throw new RuntimeException(e);
+            } catch (IOException e) {
+                throw new RuntimeException("Failed to read path \"%s\"".formatted(s), e);
             }
         }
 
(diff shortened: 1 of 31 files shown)


src/main/java/net/datafaker/providers/base/Locality.java
```diff
@@ -102,8 +102,8 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
         for (String s: paths) {
             try {
                 Files.walkFileTree(Paths.get(s).toAbsolutePath(), visitor);
-            } catch (Exception e) {
-                throw new RuntimeException(e);
+            } catch (IOException e) {
+                throw new RuntimeException("Failed to read path " + s, e);
```
snuyanzin, line 106:
```suggestion
                throw new RuntimeException("Failed to read path '" + s + "'", e);
```
sometimes when string is empty such output with quotes could save time during debug
another version with formatting is also ok

src/test/java/net/datafaker/sequence/FakeStreamTest.java
```diff
@@ -128,7 +128,7 @@ void checkWrongArguments() {
                 .maxLen(5)
                 .generate())
             .isInstanceOf(IllegalArgumentException.class)
-            .hasMessage("Max length must be not less than min length and not negative");
+            .hasMessage("Max length (5) must be not less than min length (10) and not negative");
```
snuyanzin, line 131: wouldn't it be better to extract it into vars?

snuyanzin (approved):
LGTM

I have a couple of minor comments

src/test/java/net/datafaker/sequence/FakeStreamTest.java
asolntsev: I would say it makes this test more complex.
These values are pretty near and well-readable. After introducing variables, it will be harder to read this test.

src/main/java/net/datafaker/providers/base/Locality.java
asolntsev: Good point, fixed.

PR #1224, opened 2024-05-27
Minor tweaks and cleanup

kingthorin:
From the previous few PRs and conversations.


asolntsev (approved):

src/main/java/net/datafaker/providers/base/TimeAndDate.java
```diff
@@ -21,8 +21,8 @@
  */
 public class TimeAndDate extends AbstractProvider<BaseProviders> {
 
-    private static final int DEFAULT_MIN_AGE = 18;
-    private static final int DEFAULT_MAX_AGE = 65;
+    public static final int DEFAULT_MIN_AGE = 18;
```
bodiam, line 24: Probably package protected would be fine too?

kingthorin: Sure I'll adjust in another in the next few days

PR #1225, opened 2024-05-27
fix generation of phone numbers (domestic/international/mobile)

asolntsev:
I realized that PhoneNumberTest almost didn't test anything: it did 10 iterations, but checked that error count is < 25. :)

Now phone numbers are
* fully correct for et, bg, mk, md, al, ru
* partially correct for most countries (77% errors tolerance)
* other countries are marked as FIXME.


src/test/java/net/datafaker/providers/base/PhoneNumberTest.java
```diff
@@ -28,14 +34,12 @@ void testCellPhone_enUS() {
...
+    @RepeatedTest(COUNT)
     void testPhone_esMx() {
         final BaseFaker f = new BaseFaker(new Locale("es", "MX"));
         final PhoneNumber phoneNumber = f.phoneNumber();
-        for (int i = 0; i < 10; i++) {
-            assertThat(phoneNumber.cellPhone()).matches("(044 )?\\(?\\d+\\)?([- .]\\d+){1,3}");
-            assertThat(phoneNumber.phoneNumber()).matches("\\(?\\d+\\)?([- .]\\d+){1,3}");
-        }
+        assertThat(phoneNumber.cellPhone()).matches("(044 )?\\(?\\d+\\)?([- .]\\d+){1,3}");
+        assertThat(phoneNumber.phoneNumber()).matches("\\(?\\d+\\)?([- .]\\d+){1,3}");
```
kingthorin, line 42: The groupings may as well be non-capturing where possible.

src/test/java/net/datafaker/providers/base/PhoneNumberTest.java
```diff
@@ -57,103 +62,139 @@ void testPhone_CA() {
...
+                    log.fine("Invalid phone: %s".formatted(phoneNumber));
+                }
+                errorCount++;
+            }
+        }
+        assertThat(errorCount).isLessThanOrEqualTo(allowedErrorsCount);
+    }
+
+    private static Arguments args(Locale locale, String phoneNumberRegion) {
+        return args(locale, phoneNumberRegion, (int) (COUNT * 0.77));
```
kingthorin, line 128: Thought you said 66?

src/test/java/net/datafaker/providers/base/PhoneNumberTest.java
```diff
@@ -57,103 +62,139 @@ void testPhone_CA() {
...
+    }
+
+    private static Arguments args(Locale locale, String phoneNumberRegion, int allowedErrorsCount) {
+        return Arguments.of(locale, phoneNumberRegion, allowedErrorsCount);
```
snuyanzin, line 132: should we just `import static` for `Arguments` and then it could be shorter like
```java
of(locale, phoneNumberRegion, allowedErrorsCount);
```
?

src/test/java/net/datafaker/providers/base/PhoneNumberTest.java
snuyanzin: may be also extract into a var?

src/test/java/net/datafaker/providers/base/PhoneNumberTest.java
```diff
@@ -57,103 +62,139 @@ void testPhone_CA() {
...
+        final BaseFaker faker = new BaseFaker(locale);
+        final PhoneNumber phoneNumberProvider = faker.phoneNumber();
+        for (int i = 0; i < COUNT; i++) {
+            String phoneNumber = phoneNumberProvider.cellPhone();
+            Phonenumber.PhoneNumber proto = util.parse(phoneNumber, phoneNumberRegion);
+            if (!util.isValidNumberForRegion(proto, phoneNumberRegion)) {
+                if (log.isLoggable(FINE)) {
+                    log.fine("Invalid phone: %s".formatted(phoneNumber));
+                }
+                errorCount++;
```
snuyanzin, line 121:
```suggestion
                errorCount++;
                if (allowedErrorsCount > errorCount) {
                    break;
                }    
```
I guess there is no need to iterate if we already exceed the limit

src/test/java/net/datafaker/providers/base/PhoneNumberTest.java
asolntsev: > Thought you said 66?

Yes, initially I put 66%, but later had to increase this number to 77% because of flaky tests failed on CI. :)

asolntsev: > extract into a var

Yes, I've extract 77% to a constant.

src/test/java/net/datafaker/providers/base/PhoneNumberTest.java
asolntsev: It's shorter, but is it really readable?
When I read just "of(xxx)", how can I understand what this "of" means"? Of what?

Compared to that, `Arguments.of` is more readable.

src/test/java/net/datafaker/providers/base/PhoneNumberTest.java
asolntsev: Yep, I've added "break" at these places.
Though, this code is temporary anyway. We need to configure phone numbers to achieve 0 error tolerance. :)

src/test/java/net/datafaker/providers/base/PhoneNumberTest.java
snuyanzin: I guess it depends on the context
if we are talking about `of` in general then I would agree it is questionable, however here we are in test class, in arguments provider method, with this knowledge it is more readable then from my point of view

snuyanzin: anyway i'm not going to stop merging just because of this
I expressed my concerns and leave the final decision up to you

snuyanzin (approved):

src/test/java/net/datafaker/providers/base/PhoneNumberTest.java
asolntsev: @kingthorin I agree, but now I feel out of resources to refactor these old regexp as well. :(

kingthorin: No problem

src/test/java/net/datafaker/providers/base/PhoneNumberTest.java
bodiam: I'm a huge fan of static imports (AssertJ for example), but in this case, importing "of" doesn't really add to readability, so I'd lean towards keeping the classname there

bodiam: Oh, I'm too late it seems :)

src/test/java/net/datafaker/providers/base/PhoneNumberTest.java
bodiam: I think I tried that some time ago, to get close to zero tolerance, but while I managed for some countries, it was non trivial to get them all in shape. Happy if you give it a shot though

Issue #1221, opened 2024-05-27
Feature request: add support to generate a random PNG/SVG.


erikpragt-connectid:
We currently have a testcase in which we pass in some data, such as:

```
 "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAFAAAABQCAYAAACOEfKtAAAACXBIWXMAAA7EAAAOxAGVKw4bAAAR+ElEQVR42u1ceVDUx5f/fGc4dDiCIAoEFCQYwaOUwgPYaDxiiGWiCaJokp/......... much more here"
```

It might be nice if we could generate random images like this too, something like `faker.image().base64PNG()` or so.

See for references:
* https://github.com/darpandodiya/random-image-generator/blob/master/RandomImageGenerator.java
* https://codereview.stackexchange.com/questions/244123/fastest-way-to-create-random-pixel-image
* https://stackoverflow.com/questions/56825653/i-wanted-to-make-an-random-image-generator-but-it-does-not-work

and:

* https://stackoverflow.com/questions/30093680/how-to-loop-a-81-svg-imagegenerated-by-java-to-get-a-88-svg-image

(I'm happy to build this myself, I just leave it here as a reminder)

Design considerations/questions?
* It would perhaps be nice to use the same logic to generate vector and binary image formats.
* Perhaps generating a base64 and binary output should be possible?

snuyanzin:
should it be closed now or something else should be implemented?

PR #1227, opened 2024-05-28
Added support for random image generation, such as PNG, JPG, GIF and …

bodiam:
…SVG.

Should address https://github.com/datafaker-net/datafaker/issues/1221


src/main/java/net/datafaker/providers/base/Image.java
```diff
@@ -0,0 +1,97 @@
...
+        // Fill the image with white background
+        graphics.setColor(WHITE);
+        graphics.fillRect(0, 0, WIDTH, HEIGHT);
+
+        // Draw random colored boxes
+        for (int y = 0; y < HEIGHT; y += BOX_SIZE) {
+            for (int x = 0; x < WIDTH; x += BOX_SIZE) {
+                int r = faker.random().nextInt(256);
+                int g = faker.random().nextInt(256);
+                int b = faker.random().nextInt(256);
```
snuyanzin, line 59: probably usage of `net.datafaker.service.RandomService#nextRandomBytes` will be much faster since you can generate a number of bytes at once and as I see byte range perfectly suits to your case

src/main/java/net/datafaker/providers/base/Image.java
```diff
@@ -0,0 +1,97 @@
+package net.datafaker.providers.base;
+
+import net.datafaker.providers.entertainment.EntertainmentProviders;
+
+import javax.imageio.ImageIO;
+import java.awt.Color;
+import java.awt.*;
```
kingthorin, line 7: Started with specific imports but then went wildcard?

src/main/java/net/datafaker/providers/base/Image.java
```diff
@@ -0,0 +1,97 @@
...
+    private String generateBase64SVGImage() {
+        StringBuilder svg = new StringBuilder();
+        svg.append("<svg xmlns=\"http://www.w3.org/2000/svg\" width=\"").append(WIDTH).append("\" height=\"").append(HEIGHT).append("\">");
+
+        for (int y = 0; y < HEIGHT; y += BOX_SIZE) {
+            for (int x = 0; x < WIDTH; x += BOX_SIZE) {
+                int r = faker.random().nextInt(256);
+                int g = faker.random().nextInt(256);
+                int b = faker.random().nextInt(256);
+                String color = String.format("#%02x%02x%02x", r, g, b);
```
kingthorin, line 86: Since this is essentially the same for both methods, why not extract it and use a boolean to control the output or just always return a Color object and in the SVG case do: `Integer.toHexString(yourCcolor.getRGB());`

src/test/java/net/datafaker/RandomImageGenerator.java
```diff
@@ -0,0 +1,99 @@
+package net.datafaker;
+
+import javax.imageio.ImageIO;
+import java.awt.*;
+import java.awt.image.BufferedImage;
```
kingthorin, line 5: IMHO specific is better than wildcard.

src/main/java/net/datafaker/providers/base/Image.java
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Generates base64 encoded PNG, GIF, JPG and SVG images.
+ *
+ * @since 2.3.0
+ */
+public class Image extends AbstractProvider<EntertainmentProviders> {
+
+    private static final int WIDTH = 256;
+    private static final int HEIGHT = 256;
+    private static final int BOX_SIZE = WIDTH / 8;
```
kingthorin, line 24: Change the visibility to make it usable in the generator class?

Edit: If the generator class is being kept.

src/test/java/net/datafaker/RandomImageGenerator.java
kingthorin: Actually I don't understand this class (`RandomImaegGenerator`) at all, it seems to be a duplicate of `Image`

src/main/java/net/datafaker/providers/base/Image.java
bodiam: Hmm, interesting indeed. Let's fix that.

src/main/java/net/datafaker/providers/base/Image.java
```diff
@@ -0,0 +1,97 @@
...
+import java.awt.*;
+import java.awt.image.BufferedImage;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Base64;
+
+import static java.awt.Color.WHITE;
+
+/**
+ * Generates base64 encoded PNG, GIF, JPG and SVG images.
```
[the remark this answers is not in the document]
bodiam: Well, I haven't really decided yet. I can imagine you might want to generated 8x8, 500x500, etc, but right now, I only have a need to generate images, so I kept the implementation a bit minimal.

src/main/java/net/datafaker/providers/base/Image.java
bodiam: done!

src/main/java/net/datafaker/providers/base/Image.java
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @since 2.3.0
+ */
+public class Image extends AbstractProvider<EntertainmentProviders> {
```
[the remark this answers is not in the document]
bodiam: oops! Copy and paste!

src/test/java/net/datafaker/RandomImageGenerator.java
bodiam: yeah, I used it for a demo, I really should have removed it. I now moved it and fixed it, it's minimal now. Thanks for feedback, I really should have reviewed my work better!

src/main/java/net/datafaker/providers/base/Image.java
bodiam: generator is gone.

src/main/java/net/datafaker/providers/base/Image.java
bodiam: Done. I wanted to reuse the logic, but wasn't sure how, so I kept the duplication in, but I fixed it now. Thanks for that!

src/test/java/net/datafaker/script/RandomImageGenerator.java
kingthorin, line 1: I'm not necessarily against this but I don't think we have anything similar anywhere else in the lib.

kingthorin: I didn't go hunting, I'm reviewing  on a cell phone 🙂

bodiam: Well, I put it with the other generators. With have a generator for documentation, plus another one, that's why I put it here in the scripts folder

kingthorin (approved):

PR #1228, opened 2024-05-28
Add AWS services in AWS data provider

panilya:
Source: https://aws.amazon.com/documentation-overview/


kingthorin (approved):
Seems good to me. Did these come from a single source? Could you provide a link or something? (So that it's easier to update in the future.)

src/test/java/net/datafaker/providers/base/AwsTest.java
```diff
@@ -51,4 +51,11 @@ void testRegion() {
         String region = faker.aws().region();
         assertThat(region).matches("^[a-z]{2}-(south|east|north|west|northeast|central|southeast)-\\d$");
     }
+
+    @Test
```
asolntsev, line 55: Usually we re-run this test multiple times using annotation `@RepeatableTest(10)` instead of just `@Test`.

asolntsev (approved):

src/main/java/net/datafaker/providers/base/Aws.java
```diff
@@ -8,6 +8,7 @@
 public class Aws extends AbstractProvider<BaseProviders> {
 
     private final Text.TextRuleConfig configForRoute53ZoneId;
+    private final static String AWS_SERVICE_PREFIX = "AWS";
```
bodiam, line 11: I'm not sure where you got it from, but if I look here:

https://docs.aws.amazon.com/managedservices/latest/userguide/mq-comp.html

I can see that some services are prefixed with AWS, and some are prefixed with Amazon. (and some not prefixed at all). Did you use a different list perhaps?

panilya: You're right. Didn't pay attention. Thanks for noticing this!

https://stackoverflow.com/a/46088871

panilya:
> Seems good to me. Did these come from a single source? Could you provide a link or something? (So that it's easier to update in the future.)

Updated PR description with a link to the source

src/test/java/net/datafaker/providers/base/AwsTest.java
panilya: After replacing the list where prefixes are included in the actual list entries in the YAML file I guess there's no sense to make that test repeatable, correct?

src/test/java/net/datafaker/providers/base/AwsTest.java
```diff
@@ -51,4 +51,10 @@ void testRegion() {
...
+    void testService() {
+        assertThat(faker.aws().service()).isNotEmpty();
+    }
```
kingthorin, line 58: If it's simply going to check not empty then I'd suggest using this type of test: https://github.com/datafaker-net/datafaker/blob/3a57f20c6603adec778ecb85de5e8cc525a98df0/src/test/java/net/datafaker/providers/base/ZodiacTest.java#L8-L12

panilya: Sorry for the delay. 

Done

Issue #1214, opened 2024-05-25
Remove public methods from `IdNumber` that are not intended to be directly called


asolntsev:
**Describe the bug**
It's rather a design issue, not a bug.
I don't like the fact that class `net.datafaker.providers.base.IdNumber` has all these public methods:
* `validSvSeSsn()`
* `validEnZaSsn()`
* `inValidEnZaSsn()`
* `invalidSvSeSsn()`
* `validEsMXSsn()`
* `invalidEsMXSsn()`
* `validEstonianPersonalCode()`
* `invalidEstonianPersonalCode()`
* `validAlbanianPersonalCode()`
* `invalidAlbanianPersonalCode()`
* `validMoldovaPersonalCode()`
* `invalidMoldovaPersonalCode()`
* `validBulgarianPersonalCode()`
* `invalidBulgarianPersonalCode()`

#### Problem 1:
I assume they are NOT intended to be explicitly called from client's code.
Instead, clients should generate corresponding ID numbers using locale/country:
```java
  Faker faker = new Faker(new Locale("et", "EE"));
  String code = faker.idNumbers().valid();
```

#### Problem 2:
Having these methods as public makes it impossible to _extend_ functionality of DataFaker. Like, when user adds a custom JAR to classpath that implements yet another language/country.

**Expected behavior**
It should be possible to add "yet another country" without adding any new public methods.
It should be enough just to add file "LANG.yml" and probably some (package-private) classes like `BulgarianIdNumber`.

**Versions:**
 - Faker Version: 2.2.2

kingthorin:
If that's the intent then I'm in agreement. I haven't been working on the project long enough to know.

bodiam:
Makes sense. I don't think we have a super clear agreement about what is our public api vs our more internal. What would you propose to limit the visibility of these methods? Make them package protected? Or something else?

asolntsev:
I think we don't need these methods at all. 
What I suggest:
1. Remove public methods from `IdNumber` class:
   * `public String validMacedonianPersonalCode()`
   * `public String invalidMacedonianPersonalCode()`
   * etc.
2. Load all classes like `MacedonianIdNumber` as _service providers_ (a built-in mechanism in Java). Every provider knows what country it's responsible for. 

I can submit a minimal pull request to demonstrate this approach.

snuyanzin:
yes that probably makes sense

why it was done in current way - probably nobody knows since it was done in javafaker first and those contributors are not active in both communities... 

[...]

PR #1229, opened 2024-05-28
#1214 [refactoring] load country-specific ID number providers by country code

asolntsev:
**Sorry for too big PR.**
**I just wanted to remove unneeded public method before we release a new version and people start using them.**

Thus, we can add ID numbers for new countries just by implementing "IdNumbers" without need to add public methods like "validZhCNSsn()", "validEstonianPersonalCode()", "validBulgarianPersonalCode()" to class "IdNumber".

Now all country-specific ID number generators are loaded from file "src/main/resources/META-INF/services/net.datafaker.idnumbers.IdNumbers" using a standard Java mechanism "service loader".

+ rename existing "*IdNumber" classes to human-readable, e.g. "ZhCnIdNumber" to "ChineseIdNumber".


src/main/java/net/datafaker/idnumbers/IdNumbers.java
```diff
@@ -1,7 +1,24 @@
...
+
 import java.time.format.DateTimeFormatter;
 
 public interface IdNumbers {
     DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern("yyMMdd");
+
+    /**
+     * @return ISO-2 code of the country ("US" for America, "EE" for Estonia etc.)
+     */
+    String country();
```
bodiam, line 13: would countryCode be perhaps a better named method? A method like `country()` might return "America" or so.

bodiam: Also, shouldn't this class perhaps be called "IdNumber", instead of "IdNumbers" (so, without the `s`?)

bodiam:
I left a tiny comment, but really nice PR, thanks for that!

src/main/java/net/datafaker/idnumbers/AlbanianIdNumber.java
```diff
@@ -11,13 +11,18 @@ public class AlbanianIdNumber implements IdNumbers {
...
+    }
+
+    public String generateInvalid(BaseProviders faker) {
+        String pin = generateValid(faker);
         int invalidMonth = faker.number().numberBetween(93, 99);
         return pin.substring(0, 2) + invalidMonth + pin.substring(4);
     }
 
-    public String getValid(BaseProviders faker) {
+    public String generateValid(BaseProviders faker) {
```
kingthorin, line 25: This will break for anyone using it directly. We should deprecate for removal and it can call the new method.

kingthorin: Also shouldn't generate valid have an override annotation?

kingthorin: I guess my original comment here goes for any of these country classes

kingthorin (commented):
I didn't get through the whole PR

src/main/java/net/datafaker/providers/base/Options.java
```diff
@@ -104,6 +104,16 @@ public String option(String... options) {
         return options[faker.random().nextInt(options.length)];
     }
 
+    /**
+     * Returns a random String element from an varargs.
```
snuyanzin, line 108: I guess it could be not only `String`

src/main/java/net/datafaker/providers/base/Options.java
```diff
@@ -104,6 +104,16 @@ public String option(String... options) {
...
+     *
+     * @param options The varargs to take a random element from.
+     * @return A randomly selected element from the varargs.
+     */
+    public <T> T option(List<T> options) {
+        return options.get(faker.random().nextInt(options.size()));
+    }
```
snuyanzin, line 115: At the very end of this class there is already existing method 
```java
    /**
     * Returns a random element from a list.
     *
     * @param list The list to take a random element from.
     * @param <E>  The type of the elements in the list.
     * @return A randomly selected element from the list.
     */
    public <E> E nextElement(List<E> list) {
        return list.get(faker.random().nextInt(list.size()));
    }
```
will not it help?

We need a separate for `String` because it could be used in expressions, however it's impossible to use generics in expressions... at least right now

src/main/java/net/datafaker/idnumbers/IdNumbers.java
asolntsev: @bodiam I like both variants: `countryCode` and `IdNmberGenerator`. Will rename.

src/main/java/net/datafaker/providers/base/Options.java
asolntsev: > I guess it could be not only String

@snuyanzin Good point, fixed javadoc.

src/main/java/net/datafaker/idnumbers/AlbanianIdNumber.java
kingthorin: Great!

kingthorin (approved):
LGTM

src/main/java/net/datafaker/idnumbers/IdNumbers.java
bodiam: Nice, thanks!

bodiam:
LGTM2!

PR #1233, opened 2024-05-31
[refactoring] extract phone numbers configuration to country-specific files

asolntsev:
The phone numbers do not depend on language. They depend only on country.

### The motivation
Say, my goal is to generate an Albanian phone number. 
I don't care about the language. 

Both `new Locale("sq", "AL")` and `new Locale("en", "AL")` should generate Albanian phone numbers.

### The problem
But currently only `new Locale("sq", "AL")` generates Albanian phone number, while `new Locale("en", "AL")` generate American phone number. Unexpected.

P.S. Unfortunately, I didn't have enough time to refactor all countries, only part of them.


snuyanzin (approved):
I have only superficial knowledge about phone number generation
so I don't see issues here and tests are green

LGTM from my side

PR #1237, opened 2024-05-31
Add Large Language Model provider

panilya:


src/main/resources/en/largelanguagemodel.yml
kingthorin, line 1: No Grok? Is it based on one of these or are these only open models er something like that

bodiam (approved):

src/main/resources/en/largelanguagemodel.yml
```diff
@@ -0,0 +1,56 @@
...
+      - "Mistral Large"
+      - "Codestral"
+      - "Gemini 1.0 Pro"
+      - "Gemini 1.0 Ultra"
+      - "Gemini 1.5 Pro"
+      - "Gemini 1.5 Flash"
+      - "Gemma 7B"
+      - "LLaMA"
+      - "Llama 2"
+      - "Code Llama"
```
bodiam, line 42: maybe add Grok-1 and OLMo?

src/main/resources/en/largelanguagemodel.yml
panilya: I listed only popular models. Some are open-source models, and some aren't (like Llama by Meta vs GPT by <del>Closed</del>OpenAI).

You can check this website to see the ratings of LLM models: https://huggingface.co/spaces/lmsys/chatbot-arena-leaderboard. The leaderboard is based on people's ratings.

It's not based on anything. They say it was trained from scratch, uses a Mixture of expert techniques (like Mixtral 8x7B) etc.

src/main/resources/en/largelanguagemodel.yml
panilya: Added Grok for Grok's fans 😄

PR #1240, opened 2024-06-01
Another attempt to get automated releases working.

bodiam:
Copied from https://github.com/markdown-asciidoc/markdown-to-asciidoc/blob/master/.github/workflows/release.yml

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,19 +1,33 @@
-name: Release non-snapshot package to the Maven Central Repository
+name: Release New Version
 on:
-  release:
-    types: [created]
+  workflow_dispatch:
+    inputs:
+      releaseversion:
+        description: 'Release version'
+        required: true
 jobs:
   publish:
     runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,19 +1,33 @@
...
-          GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }}
-
       - uses: actions/checkout@v4
-
-      - name: Set up Maven Central Repository
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
+          distribution: 'adopt'
```
kingthorin, line 20: Why two different distros in the same workflow?

bodiam: Cause copy and paste? ;-)

bodiam: oh, it's even in the same file! I thought you meant the release has a different distro than the build. I'll fix this!

bodiam: fixed.

PR #1243, opened 2024-06-01
Fixed airport code.

bodiam:
Hi all,

I don't think the fix here is correct: https://github.com/datafaker-net/datafaker/pull/1241. While N55 might be a valid airport code, it's not a valid [ICAO](https://en.wikipedia.org/wiki/ICAO_airport_code) code, which is what the method returns:

```
    /**
     * Returns an airport ICAO code.
     * See also: <a href="https://en.wikipedia.org/wiki/List_of_airports_by_ICAO_code:_A">https://en.wikipedia.org/wiki/List_of_airports_by_ICAO_code:_A</a>
     */
    public String airport() {
        return resolve("aviation.airport");
    }
```

image


kingthorin (approved):
Thanks!

Sorry I didn't dig far enough.

PR #1244, opened 2024-06-01
Drop deprecated Xml

snuyanzin:
Probably a bit late however need to remove the latest old format approach for XML since it was also replaced by `XMLTransformer`, no need for that anymore


asolntsev (approved):

snuyanzin:
Thanks for taking a look

PR #1245, opened 2024-06-01
Mark Internet#pass deprecated

snuyanzin:
`Internet#password` is very limited with its functionality and we should encourage users to use `Text` instead which covers much more cases

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Internet.java
+++ b/src/main/java/net/datafaker/providers/base/Internet.java
@@ -176,26 +176,62 @@ public class Internet extends AbstractProvider<BaseProviders> {
         return resolve("internet.http_method");
     }
 
+    /**
+     * Consider usage of {@link net.datafaker.providers.base.Text#text} instead,
+     * look at net.datafaker.providers.base.TextTest#textShouldContain3RULowerCaseAnd5CustomSpecialSymbols()
+     * as an example.
+     */
+    @Deprecated(since = "2.3.0")
     public String password() {
         return password(8, 16);
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


src/main/java/net/datafaker/providers/base/Internet.java
```diff
@@ -176,26 +176,62 @@ public String httpMethod() {
         return resolve("internet.http_method");
     }
 
+    /**
+     * Consider usage of {@link net.datafaker.providers.base.Text#text} instead,
+     * look at net.datafaker.providers.base.TextTest#textShouldContain3RULowerCaseAnd5CustomSpecialSymbols()
```
kingthorin, line 181: RULower? Regular upper lower?

kingthorin (commented):
Do we bundle unit tests in the library? It seems strange to document that users should check unit tests.

src/main/java/net/datafaker/providers/base/Internet.java
snuyanzin: >textShouldContain3RULowerCaseAnd5CustomSpecialSymbols

means that text should contain 3 lower case letters from `RU` locale and 5 custom special symbols

snuyanzin:
>Do we bundle unit tests in the library?

no
> It seems strange to document that users should check unit tests.

yep, there are 2 ways I see: either put example in javadoc for `net.datafaker.providers.base.Text#text` or in site documentation

snuyanzin:
Ok, added to javadoc

kingthorin (approved):
Thanks, I think that's much more usable

bodiam:
I've reverted this change. Without these methods it's way to complex too create a password, and I don't see why these methods can't exists next to each other.

PR #1246, opened 2024-06-01
Add doc about advanced char sequence generation

snuyanzin:


docs/documentation/password.md
```diff
@@ -0,0 +1,31 @@
...
+Now it allows to specify different set of symbols which should be contained and minimum amount of symbols per such set.
+
+Imagine there is a need to generate a password with length 8 and containing minimum 3 digits and 2 upper case symbols from en locale.
+=== "Java"
+
+    ``` java
+         var faker = new Faker();
+         String password = faker.text().text(Text.TextSymbolsBuilder.builder()
+                             .len(8)
+                             .with(EN_UPPERCASE, 3)
```
kingthorin, line 13: 2 like the lead-in says?

kingthorin:
Downstream tests are failing but the issue seems unrelated to this PR

docs/documentation/password.md
snuyanzin: yep, thanks for noticing, fixed

snuyanzin:
>Downstream tests are failing but the issue seems unrelated to this PR

yeahh... that was the reason https://github.com/datafaker-net/datafaker-gen/pull/12
we added connector to BigQuery and in tests it requires service account...

Anyway we don't need such test on datafaker side, so just skip it within https://github.com/datafaker-net/datafaker/pull/1247

kingthorin (approved):

PR #1249, opened 2024-06-04
generate phone numbers using Google library "libphonenumber"

asolntsev:
This library already contains phone number patterns for all countries. It's reasonable to reuse them instead of re-inventing the wheel.

Now PhoneNumberTest is stable, and doesn't need all these "max allowed errors count" hacks.

P.S. Maybe it's a good idea **not** to include this PR into the nearest release. 
It's quite a big change, maybe postpone it to a next major release?


src/main/java/net/datafaker/providers/base/PhoneNumberGenerator.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    String randomPhoneNumber(String countryCodeIso2, PhoneNumberType type, PhoneNumberFormat format) {
+        Phonenumber.PhoneNumber phoneNumber = randomPhoneNumber(countryCodeIso2, type);
+        return libPhoneNumber.format(phoneNumber, format);
+    }
+
+    private Phonenumber.PhoneNumber randomPhoneNumber(String countryCodeIso2, PhoneNumberType type) {
+        String pattern = CACHE.computeIfAbsent(new CacheKey(countryCodeIso2, type),
+            (key) -> phoneNumberPattern(countryCodeIso2, type));
```
kingthorin, line 39: Is `CACHE` really a constant if it's modified/filled after instantiation?

src/main/resources/en-CA.yml
```diff
@@ -8,43 +8,6 @@ en-CA:
     internet:
       free_email: [gmail.com, yahoo.ca, hotmail.com]
       domain_suffix: [ca, com, biz, info, name, net, org]
-    phone_number:
-      country_code: ['1']
```
kingthorin, line 12: I don't necessarily think it's wrong I just want to understand. In some cases country code was left in some it was removed?

src/test/java/net/datafaker/FakerTest.java
```diff
@@ -277,10 +277,10 @@ void differentLocalesTest() {
     @Test
     @Timeout(value = 3, unit = TimeUnit.MINUTES)
     void issue883Test() throws InterruptedException {
-        for (int i = 0; i < 10_000_000; i++) {
+        for (int i = 0; i < 10_000; i++) {
             Faker f = new Faker();
             String s = f.ancient().god();
-            if (i % 1_000_000 == 0) {
+            if (i % 1_000 == 0) {
```
kingthorin, line 283: This was purposefully huge to ensure a timing/performance issue isn't re-introduced.

kingthorin: If it's really going to be dropped by a factor of 10 then the timeout should as well I guess.

kingthorin (commented):
Overall looks good to me. Nothing seems breaking to me either so I don't see why we'd wait.

src/main/java/net/datafaker/providers/base/PhoneNumberGenerator.java
asolntsev: Not sure if I understand your question correctly.
Yes, `CACHE` is `static final` field (of type `Map`). We don't want to assign it any other value.
But we want to store more and more values in this map.

src/main/resources/en-CA.yml
asolntsev: @kingthorin I left `country_code:` in those "*.yml" files where it was the only value. 
I just was not sure if I can leave these files empty, or should I delete those files. :)
It was easier to leave them untouched, just in case.

src/test/java/net/datafaker/FakerTest.java
asolntsev: We can restore `10_000_000`, of course. It just seemed to much for me. Was a bit slow on my machine.

src/main/java/net/datafaker/providers/base/PhoneNumberGenerator.java
kingthorin: If it's are being added to it over time then it's not really a constant 😉 it isn't really "final" (it's not preloaded with content)

src/main/resources/en-CA.yml
kingthorin: Okay that's fine.

src/test/java/net/datafaker/FakerTest.java
kingthorin: I believe that's the right move.

src/main/java/net/datafaker/providers/base/PhoneNumberGenerator.java
asolntsev: Well... It's a widely used pattern in Java..
For example, look at the existing class `net.datafaker.annotations.FakeResolver`:

```java
public class FakeResolver<T> {
    private static final Map<Class<?>, FakeResolver<?>> CLASS_2_FAKE_RESOLVER = new CopyOnWriteMap<>(IdentityHashMap::new);

    private static final Map<Class<?>, Schema<Object, ?>> DEFAULT_SCHEMA_CACHE = new CopyOnWriteMap<>(IdentityHashMap::new);
```

kingthorin: Alrighty, I don't like it but fine, you're right the code base already uses it.

erikpragt-connectid:
Sorry, I'd like to veto this PR. I'm against bringing in new production libraries into Datafaker. It took me a lot of time to get most of them out, and only the Regex one is a special case for this. If you really want this, we could discuss making an SPI for this or so, that when you add a dependency to datafaker-phone-numbers it would use this library instead?

src/test/java/net/datafaker/integration/FakerIntegrationTest.java
```diff
@@ -157,6 +150,8 @@ void testExceptionsNotCoveredInAboveTest(Locale locale, Random random) {
     private static Stream<Arguments> dataParameters() {
         List<Arguments> arguments = new ArrayList<>();
         arguments.add(Arguments.of(Locale.ENGLISH, new Random()));
+        arguments.add(Arguments.of(new Locale("en-US"), new Random()));
```
erikpragt-connectid, line 153: we really should stop using this `-` or `_` approach, it's not something we support officially, and it's removed from all documentation AFAIK.

asolntsev: Do you mean replacing `new Locale("en-US")` by `new Locale("en", "US")`?
Yes, why not. 

P.S. Starting from Java 19, this constructor is also deprecated. Should be replaced by `Locale.of("en", "US")`.

asolntsev:
> I'm against bringing in new production libraries into Datafaker.

Interesting take. :)
And what's the reason of not including USEFUL libraries? 
Do you really want just to do the same job again?

NB! Look at [PhoneNumberTest](https://github.com/datafaker-net/datafaker/blob/main/src/test/java/net/datafaker/providers/base/PhoneNumberTest.java#L147):
DataFaker _almost always_ **generates an invalid phone number**. 
For some countries, in 100% cases, and for most countries - in 77% cases. 

@erikpragt-connectid
Is there any reason in the world **why generating invalid data is better than adding a dependency**?

src/test/java/net/datafaker/integration/FakerIntegrationTest.java
erikpragt-connectid: Yes, the `new Locale("en-US")` is not an officially supported Locale, that's something Javafaker introduced, and ideally, it should be removed from the code, not added.

src/test/java/net/datafaker/FakerTest.java
snuyanzin: Before changing number here we need to have a good replacement for that
Just changing `10_000_000` to `1_000` makes test useless since with so small number it is unable to reproduce the original problem https://github.com/datafaker-net/datafaker/issues/883  for the code before it was fixed

snuyanzin:
Despite the fact I understand the rationale of usage `libphonenumber` I would agree that it's better not to add it. 
Datafaker provides a number of different generators and adding a library just to improve one of them could be overkill in future. For instance if we add a lib for phone numbers why not for addresses and etc... Then finally it will tend to be a super fat... At the same time not every user uses phone number generator at least once...

From the side as it was mentioned such lib could improve quality of generated data. 

Thus I would suggest to think towards kind of plugin api (SPI or whatever). We already have custom provider support we allows to create a provider with better quality. May be we need to think about more fine-grained api allowing to use more accurate custom code for some providers (currently just brainstorming)

asolntsev:
@snuyanzin @erikpragt-connectid Ok, would you mind creating a new repository like https://github.com/datafaker-net/datafaker-phones? Who will be responsible for releasing new versions then? Someone will need to always remember to release `datafaker-phones.jar` every time when they release `datafaker.jar` It's easy to forget. Are you sure it's convenient? 

P.S. I sincerely don't understand why you want to avoid dependencies.
There is absolutely no reasons for this. 
Sorry, it's not normal. All projects have dependencies. 
Look around, the world is global. Countries use each other's products. Projects use each other as dependencies. Etc.

snuyanzin:
>Ok, would you mind creating a new repository like https://github.com/datafaker-net/datafaker-phones? Who will be responsible for releasing new versions then? Someone will need to always remember to release datafaker-phones.jar every time when they release datafaker.jar It's easy to forget. Are you sure it's convenient?

should we have a new repo? I guess having another module depending on a library would be ok.
Since then there could be a core module without extra dependencies and some other with. At the same time since everything in one repo then all could be released together and there is no issue with forgetting to release something

asolntsev:
Ok, you suggest to make a multi-module project from datafaker? Yes, sure, why not. I can modify the PR accordingly.

bodiam:
Well, do you really need the library at runtime if all you need is the formats? As in, if we would use the lib phone number at  build  time to extract the formats, would that be an option?

kingthorin:
libohonenumber's data is available as an xml file. I guess we could create a GitHub workflow to pull it and update our yaml(s) from time to time.

Though that seems like a lot of extra work to for some reason avoid including a 350k dependency. (I know you're worried about bloat but I suspect we could come up with some reasonable criteria/rationale to be applied in this and future cases.)

bodiam:
Javafaker had a ton of dependencies. Most of the dependencies had bugs or CVEs, which caused Javafaker to have CVEs too. Besides that, dependencies can have dependencies, and those dependencies can conflict with other dependencies. 

Maybe "no extra dependencies" is  too strict, but I wouldn't know where to draw the line. Just keep in mind we have around 500.000 downloads per month, so that's the impact we can have on people.

kingthorin:
> Maybe "no extra dependencies" is too strict, but I wouldn't know where to draw the line. Just keep in mind we have around 500.000 downloads per month, so that's the impact we can have on people.

That's a good stat to know, and definitely helps contextualize things.

As for criteria, based just on what's been mentioned so far.
- Minimal size (<1MB?)
    - libphonenumber is 350k. (On it's own, not compressed the library's metadata.xml is ~950k)
- No dependency tree (single level, the library itself)
    - libphonenumber has zero dependencies according to maven: https://mvnrepository.com/artifact/com.googlecode.libphonenumber/libphonenumber/8.13.38)
- Fills a need we can't do "simply".
    - Validating the formats for global phone numbers isn't a trivial task. We proven this with tests etc already.

bodiam:
Thanks for that writeup, appreciate it! Given that it doesn't introduce a whole tree of dependencies (see the merits of dependency free libraries? 😉) and the size is small, and all of the above, plus that we live in a democracy, and "code wins", I'm happy to change my mind on this. Maybe I'll maakt a small note in the contributing section (be mindful when introducing new libraries, or something like that). 

What do you think @snuyanzin ?

bodiam:
Small question though : is the result still repeatable? I think I saw something in the PR about having to generate a phone number a couple of times to create a valid one?

asolntsev:
> is the result still repeatable

Good question. 
But yes, the result is highly repeatable (all tests pass). 

But yes, inside we have a loop a'la "up to 100 retries".
The reason is that regex is not enough to generate a 100% valid phone number.
We generate a random phone using regex, but then need to validate it using "libphonenumber" because some generated numbers are still not valid because of area codes etc.

asolntsev:
@bodiam @snuyanzin You decide. I can prepare another PR which adds "libphonenumber" generation as a separate module, no problems. It may be even good to show others how they could add other modules as well. 

The only drawback is that such PR would move all existing sources to a subfolder, something like "core" or "engine" or just "datafaker". It's because Maven multi-module architecture requires the root project have type "pom", and submodules can have type "jar".

bodiam:
I think the current way of doing it is fine, no need to overcomplicate things. I'm happy if you want to merge it.

PR #1251, opened 2024-06-05
Add boardgame provider

jensGiehl:
Data taken from boardgamegeek.com and recommend.games


src/test/java/net/datafaker/providers/base/BoardgameTest.java
```diff
@@ -0,0 +1,23 @@
+package net.datafaker.providers.base;
+
+import java.util.Collection;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
```
kingthorin, line 6: Please use specific imports

kingthorin (commented):
This should probably be an Entertainment provider

src/test/java/net/datafaker/providers/base/BoardgameTest.java
jensGiehl: The import is even completely superfluous. Thanks for the tip!

jensGiehl:
> This should probably be an Entertainment provider

For Entertainment, I wasn't sure if it belonged there, since only movies are listed there so far.

src/main/java/net/datafaker/providers/base/Boardgame.java
```diff
@@ -0,0 +1,40 @@
+package net.datafaker.providers.base;
+
+/**
+ * since 2.2.3
```
snuyanzin, line 4: why is it not an annotation?

jensGiehl: fixed :)

bodiam:
Hi @jensGiehl , thanks for this contribution! Looks great, and thanks for addressing the review comments!

kingthorin:
Geez you were quick. I read the comment about Entertainment and went and look, I almost told you that you merged to quickly and that there was already a Boardgame provider, then I realized the file I was looking at was only 2 minutes old 🤣

PR #1252, opened 2024-06-11
Attempt to publish documentation automatically

bodiam:


kingthorin (approved):
Seems okay to me. I've never seen a run block within a checkout step before but that doesn't mean it's wrong (just means I'm missing knowledge).

asolntsev (approved):

bodiam:
I copied all this from the mkdocs documentation, because for some reason mkdocs is broken locally (some python issue which I haven't been able to resolve)

PR #1256, opened 2024-06-13
fix flaky test: remove leasing "0" from generated phone number

asolntsev:
the leading "0" makes the generated phone number an invalid in very rare cases in Argentina. For example, "011 1579-5896" was generated from AR pattern, but occurred to be invalid.

And it seems that "0" is still needed in Italy.


kingthorin (approved):
Thank you!

PR #1261, opened 2024-06-16
Add Spark SQL dialect

gatear:
###
Add Spark SQL support.
See "INSERT INTO" spec https://spark.apache.org/docs/3.2.1/sql-ref-syntax-dml-insert-into.html

There are some issues with existing design in order to support all Spark types. 
Spark SQL [has 3 complex data types](https://spark.apache.org/docs/latest/sql-ref-datatypes.html):
- Array
- Map
- Struct

Insertions look like this
```sql
-- Inserting an array
INSERT INTO MyTable (array_column) VALUES (ARRAY(1, 2, 3));

-- Inserting a map
INSERT INTO MyTable (map_column) VALUES (MAP('key1', 'value1', 'key2', 'value2'));

-- Inserting a struct
INSERT INTO MyTable (struct_column) VALUES (NAMED_STRUCT('field1', 'value1', 'field2', 'value2'));
```

So notable design changes are
- array start and end markers are now configurable with default "[", "]". Spark requires "(" and ")".
- row prefix is configurable as keyword "ROW" doesn't exist but we have "NAMED_TUPLE" for complex fields in Spark.
- a named tuples requires a prefix for each value, the field name (but root columns don't require field name prefix)
- there's also Map support required. I think it makes sense to correlate `java.util.Map` with Spark SQL Map
- batch insert is not support so I think it's best to throw exception on attempt to generate "batch" statements

I tested on latest Databricks runtime this generated SQL and work well 👍 
```sql
INSERT INTO `MyTable` (`bytes`) VALUES (ARRAY(1, 0));
INSERT INTO `MyTable` (`booleans`) VALUES (ARRAY(true, false));
INSERT INTO `MyTable` (`ints`) VALUES (ARRAY(1, 2, 3));

INSERT INTO `MyTable` (`struct_array`) VALUES (NAMED_STRUCT('name1', '1', 'struct', NAMED_STRUCT('name', ARRAY(1, 2, 3))));
INSERT INTO `MyTable` (`struct_struct`) VALUES (NAMED_STRUCT('name1', '1', 'struct', NAMED_STRUCT('name', '2')));
INSERT INTO `MyTable` (`longs`) VALUES (ARRAY(23, 45));
INSERT INTO `MyTable` (`empty_map`) VALUES (MAP());
INSERT INTO `MyTable` (`maps`) VALUES (MAP('k1', MAP('k1', 'v1'), 'k2', MAP('k1', 'v1')));
```

### DIFF
--- a/src/main/java/net/datafaker/transformations/sql/SqlDialect.java
+++ b/src/main/java/net/datafaker/transformations/sql/SqlDialect.java
@@ -10,7 +10,9 @@ import static net.datafaker.transformations.sql.SqlDialect.AuxiliaryConstants.DE
 import static net.datafaker.transformations.sql.SqlTransformer.SQLKeyWords.INSERT_ALL;
 import static net.datafaker.transformations.sql.SqlTransformer.SQLKeyWords.INSERT_INTO;
 import static net.datafaker.transformations.sql.SqlTransformer.SQLKeyWords.INTO;
+import static net.datafaker.transformations.sql.SqlTransformer.SQLKeyWords.ROW;
 import static net.datafaker.transformations.sql.SqlTransformer.SQLKeyWords.SELECT_1_FROM_DUAL;
+import static net.datafaker.transformations.sql.SqlTransformer.SQLKeyWords.NAMED_STRUCT;
 import static net.datafaker.transformations.sql.SqlTransformer.SQLKeyWords.VALUES;
 
 public enum SqlDialect {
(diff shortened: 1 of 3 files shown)


src/main/java/net/datafaker/transformations/sql/SqlDialect.java
```diff
@@ -54,6 +83,7 @@ public enum SqlDialect {
 
     private static final String DEFAULT_BEFORE_EACH_BATCH_PREFIX = "       ";
 
+
```
snuyanzin, line 86: please delete this line

src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
```diff
@@ -163,6 +176,23 @@ private String handleObjectInCollection(Object value) {
         return result.toString();
     }
 
+    private String handeObjectInMap(Object value) {
```
snuyanzin, line 179:
```suggestion
    private String handleObjectInMap(Object value) {
```
seems like a misprint

src/main/java/net/datafaker/transformations/sql/SqlDialect.java
```diff
@@ -45,7 +47,34 @@ public enum SqlDialect {
     REDSHIFT("\"", Casing.TO_LOWER),
     SNOWFLAKE("\""),
     TERADATA("\""),
-    VERTICA("\"", Casing.UNCHANGED);
+    VERTICA("\"", Casing.UNCHANGED),
+    SPARKSQL("`",
```
snuyanzin, line 51: It's better to keep records here sorted alphabetically

snuyanzin (commented):
Thanks for the contribution 

in general lgtm, I left some comments

gatear:
Thanks @snuyanzin for the review 🙏 

I also have some questions ..
- in order to port this change to faker version 1.x I need to also open a PR to branch `1.x` ?
- I would also invest in some documentation, where would you suggest it's best to make it ?

snuyanzin (approved):

snuyanzin:
Thanks for addressing feedback and for the valuable contribution

snuyanzin:
>    in order to port this change to faker version 1.x I need to also open a PR to branch 1.x ?

yep, need a backport PR for that

>
    I would also invest in some documentation, where would you suggest it's best to make it ?

depending on how large documentation update you want to make, will suit as a subsection of `SQL transformation` ?
https://github.com/datafaker-net/datafaker/blame/f3fa54a4495fd8923e790701fab0570df8bd75b7/docs/documentation/schemas.md#L163

PR #1270, opened 2024-06-26
Document Spark SQL

gatear:
Mention how complex types map from Faker schema to Spark types.


kingthorin (commented):
Seems reasonable to me.

As with the previous PR I haven't used the transformers at all so I can't say if the details are technically correct.

snuyanzin:
thanks for the contribution
I feel like at some point we have to have kind of compatibility matrix: what is supported for what dialect...
However it is a topic for a separate PR/issue

snuyanzin (approved):

PR #1277, opened 2024-06-29
Refactoring: improve code

asolntsev:


src/test/java/net/datafaker/providers/base/AddressTest.java
```diff
@@ -25,8 +25,9 @@ class AddressTest extends BaseFakerTest<BaseFaker> {
     private static final Faker NL_FAKER = new Faker(new Locale("nl", "NL"));
     private static final Faker RU_FAKER = new Faker(new Locale("ru", "RU"));
     private static final Faker AU_FAKER = new Faker(new Locale("en", "AU"));
+    private static final Pattern CYRILLIC_LETTERS = Pattern.compile(".*[а-яА-Я].*");
```
snuyanzin, line 28: This doesn't cover `ёЁ`, may be some other cyrillic letters

asolntsev: That's right, these are SOME cyrillic letters, not ALL cyrillic letters.
I hope it's good enough for tests.

kingthorin: I don't disagree but incase there's a need in the future you can use the character class `\p{IsCyrillic}`

PR #1282, opened 2024-07-01
Unify `TimeAndDate` provider API

valfirst:
`TimeAndDate#past` methods use `minimum` input parameter of `long` type, but `TimeAndDate#future` methods - of `int` type. This change unifies API: `minimum` input parameter is always of `long` type

### DIFF
--- a/src/main/java/net/datafaker/providers/base/TimeAndDate.java
+++ b/src/main/java/net/datafaker/providers/base/TimeAndDate.java
@@ -68,7 +68,7 @@ public class TimeAndDate extends AbstractProvider<BaseProviders> {
      * @param unit    the time unit.
      * @return a future date from now, with a minimum time.
      */
-    public Instant future(long atMost, int minimum, TimeUnit unit) {
+    public Instant future(long atMost, long minimum, TimeUnit unit) {
         Instant minimumDate = Instant.now().plus(minimum, unit.toChronoUnit());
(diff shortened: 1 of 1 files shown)


bodiam (approved):

bodiam:
Good point, thanks for this!

kingthorin:
I know we just made these date/time changes but technically this breaks the public API (people will have to update their code to accommodate). The methods should have been deprecated and new methods implemented (the old could even call the new and cast the params too long).

[For my two cents, I know others on this project don't see it as strictly as I do 😉]

bodiam:
Hi @kingthorin 

> people will have to update their code to accommodate

Why would consumers need to change their code? It's a change from an `int` to a `long`, from the callers perspective nothing should change (hence no unit tests were changed either). 

(and yes, if we'd follow semver to the letter, we'd be on Datafaker 16 by now ;-) )

kingthorin:
Nope you're right, Java treats a number as a number in this case. I had been thinking, ex: 3 ➡️ 3L, but Java doesn't force that.

bodiam:
It's a good point though. I thought there was something, that if you pass in a Long (or Int), and we change the method signature without recompiling, the code cannot find the signature anymore or so. I don't remember the exact case for this. 

I do think one day we should use something like this: https://lvc.github.io/japi-compliance-checker/

kingthorin:
In another project I work on they use `japicmp`.

Issue #1285, opened 2024-07-03
Locale specific generation performance degradation


vitaly-ivanov:
**Describe the bug**
In the latest version, I found a performance hit when the locale is specified.

No locale: Time to process 1000000 values: 2.310795584s

Locale en-GB: Time to process 1000000 values: 40.341209708s

The results are not precise but the difference is abnormal.

**To Reproduce**
Code to reproduce is here:
https://github.com/vitaly-ivanov/datafaker-memory-leak/blob/main/app/src/main/kotlin/org/example/Performance.kt

**Expected behavior**
Similar execution time for both cases, whether locale is set or not.

**Versions:**
 - OS: macOS 14.5
 - JDK 21
 - Faker Version 2.3.0

**Additional context**
I will send a PR soon with a bugfix that improves performance, but the results with the locale specified are still not as good as without.

vitaly-ivanov:
Interim findings.

After the changes from https://github.com/datafaker-net/datafaker/pull/1286 the flame graph looks like this:

With locale:
![flamegraph](https://github.com/datafaker-net/datafaker/assets/3909922/36398287-49d4-4d88-98f9-595655ff4fea)

No locale:
![flamegraph2](https://github.com/datafaker-net/datafaker/assets/3909922/6fc819ab-0bfc-4a0b-a822-625a927c8479)

Issue #1281, opened 2024-07-01
Problems with module-info.java with version 2.3.0


eitan-rosenberg:
After upgrade to version 2.3.0

I am getting: 
Error occurred during initialization of boot layer
java.lang.module.FindException: Error reading module: G:\MyMavenRepo\net\datafaker\datafaker\2.3.0\datafaker-2.3.0.jar
Caused by: java.lang.module.InvalidModuleDescriptorException: Package net.datafaker.formats not found in module

To solve the problem, I download the code.

I move datafaker-main\src\moditect\module-info.java to \src\main\java\module-info.java

After some modifications, I have a working jar.

**Here is the modified module-info.java:**

```
module net.datafaker {

	requires java.sql;
	requires java.management;
	requires java.desktop;

	requires org.yaml.snakeyaml;
	requires libphonenumber;
	requires rgxgen;

	exports net.datafaker;
	exports net.datafaker.annotations;
//	exports net.datafaker.formats;
	exports net.datafaker.idnumbers;
	exports net.datafaker.idnumbers.pt.br;
	exports net.datafaker.providers.base;
	exports net.datafaker.providers.entertainment;
	exports net.datafaker.providers.food;
	exports net.datafaker.providers.healthcare;
	exports net.datafaker.providers.sport;
	exports net.datafaker.providers.videogame;
	exports net.datafaker.sequence;
	exports net.datafaker.service;
	exports net.datafaker.service.files;
	exports net.datafaker.transformations;
	exports net.datafaker.transformations.sql;
//	exports net.datafaker.shaded.snakeyaml;

}
```
**And here is the original module-info.java:**

```
open module net.datafaker {
    requires java.sql;
    requires java.management;

    exports net.datafaker;
    exports net.datafaker.annotations;
    exports net.datafaker.formats;
    exports net.datafaker.idnumbers;
    exports net.datafaker.idnumbers.pt.br;
    exports net.datafaker.providers.base;
    exports net.datafaker.providers.entertainment;
    exports net.datafaker.providers.food;
    exports net.datafaker.providers.healthcare;
    exports net.datafaker.providers.sport;
    exports net.datafaker.providers.videogame;
    exports net.datafaker.sequence;
    exports net.datafaker.service;
    exports net.datafaker.service.files;
    exports net.datafaker.transformations;
    exports net.datafaker.transformations.sql;
    exports net.datafaker.shaded.snakeyaml;
}
```
Thanks.

bodiam:
[...]

PR #1286, opened 2024-07-03
Improve performance when generating with a specific locale

vitaly-ivanov:
## Overview

This is a partial fix for the issue https://github.com/datafaker-net/datafaker/issues/1285.

Result before changes:
```
Time to process 1000000 values: 40.341209708s
```

Result after changes:
```
Time to process 1000000 values: 19.826039958s
```

Result with no locale:
```
Time to process 1000000 values: 2.310795584s
```

Results are based on the code here: https://github.com/vitaly-ivanov/datafaker-memory-leak/blob/main/app/src/main/kotlin/org/example/Performance.kt

## Details

Nulls were interpreted as no attempt to load

If values were missing for the specified locale, the `loadValues` method returned null instead of an empty map, resulting in subsequent attempts to load values into the cache.

### DIFF
--- a/src/main/java/net/datafaker/service/FakeValues.java
+++ b/src/main/java/net/datafaker/service/FakeValues.java
@@ -5,6 +5,7 @@ import org.yaml.snakeyaml.Yaml;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
@@ -113,7 +114,7 @@ public class FakeValues implements FakeValuesInterface {
                 return result;
             }
         }
-        return null;
+        return Collections.emptyMap();
     }
 
     private void enrichMapWithJavaNames(Map<String, Object> result) {


asolntsev (approved):
LGTM

bodiam:
Thanks for this PR. Any chance you add a small unit / performance test to this to prevent regressions?

snuyanzin:
yep, benchmark might be a good idea

bodiam:
> > Thanks for this PR. Any chance you add a small unit / performance test to this to prevent regressions?
> 
> I think adding performance tests as unit tests may not be a good practice due to the execution time. I see that the benchmarks are in the separate repository https://github.com/datafaker-net/datafaker-benchmark. Should I add a new benchmark there?

We have a few lightweight performance tests as part of the build, so it depends a bit on how long the test run is. If it's 10 seconds, it should be fine to add it to this build, if it's more then that, perhaps Datafaker-benchmark might be better.

vitaly-ivanov:
I have added a new banchmark: https://github.com/datafaker-net/datafaker-benchmark/pull/25

Before changes:
```
Benchmark                                        Mode  Cnt     Score     Error   Units
LocalePerformanceBenchmark.en_fullname          thrpt   10  2700.254 ± 161.029  ops/ms
LocalePerformanceBenchmark.en_gb_fullname       thrpt   10    12.079 ±   1.952  ops/ms
LocalePerformanceBenchmark.en_gb_streetAddress  thrpt   10    11.230 ±   1.599  ops/ms
LocalePerformanceBenchmark.en_streetAddress     thrpt   10  2045.465 ±  70.737  ops/ms
```

After changes:
```
Benchmark                                        Mode  Cnt     Score     Error   Units
LocalePerformanceBenchmark.en_fullname          thrpt   10  2645.880 ± 216.553  ops/ms
LocalePerformanceBenchmark.en_gb_fullname       thrpt   10    38.643 ±   3.966  ops/ms
LocalePerformanceBenchmark.en_gb_streetAddress  thrpt   10    51.417 ±  11.488  ops/ms
LocalePerformanceBenchmark.en_streetAddress     thrpt   10  1909.792 ± 213.204  ops/ms
```

It gets a little better, but the difference is still significant for some reason.

snuyanzin (approved):

asolntsev:
@bodiam @snuyanzin Could you **please release DF 2.3.1** with this fix?
I heard complains about performance of DS 2.3.0 (that became obvious after we fixed the memory leaks).

bodiam:
Yes, can do, but I was hoping to release this one too: https://github.com/datafaker-net/datafaker/issues/1281

PR #1289, opened 2024-07-07
Added support for UUID v7.

bodiam:


bodiam:
Hi all, this test is failing (for obvious reasons, since a UUID is time based). How should I fix this? I can imagine 2 ways:

- I'll exclude it from the repeatable tests (there are more providers which aren't repeatable, they just haven't been affected cause their parameter count isn't 0)
- I can make the timestamp more predictable, though I was hoping to keep the concept of calling the uuidv7 method multiple times to generate time sorted IDs.

Suggestions?

kingthorin:
Do both? Then exclude the method that's time based.

That way people that want repeatability have an option and people that want technically correct uuidv7 get sortability.

bodiam:
> Do both? Then exclude the method that's time based.

I don't think it's needed to do both. If it becomes repeatable, there's no reason to exclude it. I'm just not really sure how to fake `Instant.now().toEpochMilli();`

Another alternative is to make `uuidv7` return a `UUID`, which makes sense, but it doesn't really fix the problem: the repeatability test seems to only look for Strings types for repeatable times. I think that's actually a bigger issue, since this means that `faker.timeAndDate().past()` is not checked for repeatability, and, unfortunately, it's not (given the same seed). 

Maybe we need a random `Clock` to deal with this?

kingthorin:
Instead of using time use a random long in an appropriate range?

(I believe epochtomillis returns long right?)

bodiam:
> Instead of using time use a random long with a specific seed within an appropriate range?
> 
> (I believe toEpochMilli returns long right?)

Good idea. I'll have a look at that!

PR #1291, opened 2024-07-09
Fix for module-info.

bodiam:


bodiam:
okay, seems that didn't go well. How do I fix this, anyone?

snuyanzin:
I would suggest to test first what we have after https://github.com/datafaker-net/datafaker/pull/1292

it's very likely it fixes the issue #1281

asolntsev:
@bodiam  I made a "uses ..." commit to fix the issue with ServiceProvider.
But it seems that IdNumberGenerators are still not loaded. I am out of ideas. :(

kingthorin:
For future reference it's better to rebase than merge. When you merge we end up with commits that are already applied to the default branch.

kingthorin:
image

asolntsev:
@kingthorin While we are working on a PR (where every commit can either fix or break something), it's very useful to see all commits separately. I totally don't understand why you can it "mess". 

For me, the mess would be if PR always had only one commit, and you never know what changes this commit contains.

kingthorin:
Don't worry about it. It isn't a blocker.

PR #1299, opened 2024-07-11
2.3.1 Release documentation

bodiam:


asolntsev (approved):
LGTM

docs/releases/2.3.1.md
```diff
@@ -0,0 +1,24 @@
...
+
+## New Contributors
+* @vitaly-ivanov made their first contribution in https://github.com/datafaker-net/datafaker/pull/1286
+
+## What's Changed
+
+* Unify `TimeAndDate` provider API by @valfirst in https://github.com/datafaker-net/datafaker/pull/1282
+* Improve performance when generating with a specific locale by @vitaly-ivanov in https://github.com/datafaker-net/datafaker/pull/1286
+* Use cache for fetchObject by @snuyanzin in https://github.com/datafaker-net/datafaker/pull/1288
+* Undeprecated password generation methods. by @bodiam in https://github.com/datafaker-net/datafaker/pull/1287
```
kingthorin, line 13: I know this is how it actually hit 2.3.1 but were you still planning to re-dep some of the more complicated ones and implement a password(TextConfig) method?

bodiam: @kingthorin I haven't really thought about that. It's not high on my list to be honest, since it doesn't bother me enough, plus if people really want to use the TextConfig approach, there's nothing from stopping them, plus it's sort of documented (though I don't really understand the documentation). For example, where does `EN_UPPERCASE` come from? I'm also not sure what the "other" characters in the password generation example are. Are they picked from the DIGITS and EN_UPPERCASE? And for special symbols, do I need to provide my own String? Too many questions for something like "give me a password".

I mostly wanted the simple password methods back since they're used quite a lot, and the alternative is too complex for me. So, if you're keen, go for it.

PR #1300, opened 2024-07-14
Cleaned up some unused configuration, and added some initial support …

bodiam:
…for Belgian locales.


src/test/java/net/datafaker/providers/base/AddressTest.java
```diff
@@ -23,6 +23,7 @@ class AddressTest extends BaseFakerTest<BaseFaker> {
     private final char decimalSeparator = new DecimalFormatSymbols(getFaker().getContext().getLocale()).getDecimalSeparator();
     private static final Faker US_FAKER = new Faker(new Locale("en", "US"));
     private static final Faker NL_FAKER = new Faker(new Locale("nl", "NL"));
+    private static final Faker BELGIAN_FAKER = new Faker(new Locale("nl", "BE"));
```
kingthorin, line 26: BE_FAKER for consistency?

bodiam: Could do, but we have `be.yml`, which isn't Belgium, but Belarus. If I call `BE_FAKER`, are we talking about a country or a language?

kingthorin: That's fair. Leave it as is.

kingthorin (approved):

Issue #1306, opened 2024-07-15
Unable to resolve expression during build


bodiam:
Hi all,

I have another build where an expression couldn't be resolved:
https://github.com/datafaker-net/datafaker/actions/runs/9940728395/job/27458128814?pr=1305

I was hoping it was resolved in one of the earlier 2.3.x PRs, but it seems it's still happening, I had a few failures earlier today too.

If someone could have a look, that would be great!

asolntsev:
I don't yet know what's the reason, but at least I know how to reproduce it.
Just run this single test in IDEA, then open "Run Configurations" -> "Edit configurations" -> "Modify options" -> "Repeat" -> "Until failure". And run again.

Now test will be re-running endlessly, and sooner or later will fail with the expected error. 

![image](https://github.com/user-attachments/assets/926dac25-33c6-4157-a158-8970f7dedaf5)
![image](https://github.com/user-attachments/assets/21038a50-648d-45f1-b76f-4ae8bd14b415)

snuyanzin:
@RVRhub I remember you also mentioned some findings about this

asolntsev:
@bodiam @snuyanzin I could reproduce the failure, but still don't understand why it happens. 
What I suggest:
I can submit a PR that improves logging and adds debug logs at some places in DF. 
Then we can get more information from debug logs when it happens again.

kingthorin:
I'm good with that 👆

Though I'm not exactly an authority for the project 😁

bodiam:
>  I can submit a PR that improves logging and adds debug logs at some places in DF.

I don't mind that, but perhaps for narrowing down the problem here only? It's probably not something we need to merge right, the debugging can happen on a branch?

asolntsev:
@bodiam It depends on the logging itself. But generally I think it's a good idea to add the logging to main branch. 
DEBUG logs are disabled by default, so it will not affect users. But when someone decides to investigate some specific problem, they can easily enable debug logs for the entire DF or some specific package.

bodiam:
I've done some debugging, and this is what I see:

<img width="1276" alt="image" src="https://github.com/user-attachments/assets/19b5399b-a7b6-4634-9378-34091433a9ec">

It's coming from here:

```
        Map<String[], MethodAndCoercedArgs> accessorMap =
            MAP_OF_METHOD_AND_COERCED_ARGS
                .getOrDefault(clazz, Collections.emptyMap())
[...]

PR #1310, opened 2024-07-21
#1306 add test that reproduces the flaky problem with CopyOnWriteMap

asolntsev:
it's a simplified version of what happens with `FakeValuesService.MAP_OF_METHOD_AND_COERCED_ARGS`.

@snuyanzin Can you help with fixing `CopyOnWriteMap`?

### DIFF
--- a/.run/tests.run.xml
+++ b/.run/tests.run.xml
@@ -1,0 +1,14 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="tests" type="JUnit" factoryName="JUnit">
+    <module name="datafaker" />
+    <useClassPathOnly />
+    <option name="MAIN_CLASS_NAME" value="" />
+    <option name="METHOD_NAME" value="" />
+    <option name="TEST_OBJECT" value="directory" />
+    <option name="VM_PARAMETERS" value="-ea -Duser.language=TR -Duser.country=tr" />
+    <dir value="$PROJECT_DIR$/src/test/java/net/datafaker" />
+    <method v="2">
+      <option name="Make" enabled="true" />
+    </method>
+  </configuration>
+</component>
--- a/pom.xml
+++ b/pom.xml
@@ -442,9 +442,6 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>${maven-surefire-plugin.version}</version>
                 <configuration>
-                    <includes>
-                        <include>**/CopyOnWriteMapTest.*</include>
-                    </includes>
                     <excludes>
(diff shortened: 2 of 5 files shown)


bodiam:
Hi @asolntsev , how about getting rid of the COWMap? I just did a performance test with Spring Boot native, and I did about 300.000 requests, and in my "production" code I'm getting nullpointers. For some reason, I got more NPEs in GraalVM than while using a JVM. If you want, you can easily reproduce it here by running the k6 performance test: https://github.com/datafaker-net/datafaker-native-demo

kingthorin:
Wasn't the COWMap originally introduced to address performance issues?

bodiam:
@kingthorik I believe the idea is that it has some benefits regarding performance, but at the cost that the library no longer works reliably.

asolntsev:
> how about getting rid of the COWMap? 

Yes, I like the idea. 
I even tried to do it in https://github.com/datafaker-net/datafaker/pull/1203

The argument against it was that COWMap is faster. But I never checked this statement by myself. And I don't think this speed is really important.

kingthorin:
Will using a synchronized map alleviate the issue? I know that might cause a slight performance hit but hopefully it still facilitates the original case.

asolntsev:
@kingthorin @bodiam I investigated the problem a bit more. 

Conclusion:
the problem is not in `CopyOnWriteMap`, but it's used. 

When we have map of maps, then we often call `get`, then `if != null` and `put`. This sequence doesn't work well in concurrent execution by multiple threads. Instead, we should use `computeIfAbsent`.

I submitted PR https://github.com/datafaker-net/datafaker/pull/1381 that should fix our flaky problem.

.run/tests.run.xml
```diff
@@ -0,0 +1,14 @@
...
+    <option name="MAIN_CLASS_NAME" value="" />
+    <option name="METHOD_NAME" value="" />
+    <option name="TEST_OBJECT" value="directory" />
+    <option name="VM_PARAMETERS" value="-ea -Duser.language=TR -Duser.country=tr" />
+    <dir value="$PROJECT_DIR$/src/test/java/net/datafaker" />
+    <method v="2">
+      <option name="Make" enabled="true" />
+    </method>
+  </configuration>
+</component>
```
bodiam, line 14: do we need this file?

bodiam (approved):

asolntsev:
The problem was fixed in PR https://github.com/datafaker-net/datafaker/pull/1381.
This PR is not needed anymore.

PR #1311, opened 2024-07-21
improve error handling and logging

asolntsev:
This PR doesn't fix https://github.com/datafaker-net/datafaker/issues/1306, but makes much easier debugging such problems.

### DIFF
--- a/src/main/java/net/datafaker/providers/base/BaseFaker.java
+++ b/src/main/java/net/datafaker/providers/base/BaseFaker.java
@@ -400,4 +400,9 @@ public class BaseFaker implements BaseProviders {
     public static Method getMethod(AbstractProvider<?> ap, String methodName) {
         return ap == null ? null : ObjectMethods.getMethodByName(ap, methodName);
     }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "@" + Integer.toHexString(hashCode());
+    }
 }
--- a/src/main/java/net/datafaker/providers/base/Image.java
+++ b/src/main/java/net/datafaker/providers/base/Image.java
@@ -158,8 +158,7 @@ public class Image extends AbstractProvider<BaseProviders> {
             byte[] imageBytes = baos.toByteArray();
             return "data:" + imageType.mimeType + ";base64," + Base64.getEncoder().encodeToString(imageBytes);
         } catch (IOException e) {
-            e.printStackTrace();
-            return null;
+            throw new RuntimeException("Failed to generate image %s of size %sx%s".formatted(imageType, width, height), e);
         }
     }
 
--- a/src/main/java/net/datafaker/providers/base/Internet.java
+++ b/src/main/java/net/datafaker/providers/base/Internet.java
@@ -246,11 +246,7 @@ public class Internet extends AbstractProvider<BaseProviders> {
      * @return a correctly formatted IPv4 address.
      */
     public String ipV4Address() {
-        try {
-            return getIpV4Address().getHostAddress();
-        } catch (UnknownHostException e) {
-            return "127.0.0.1";
-        }
+        return getIpV4Address().getHostAddress();
     }
 
     /**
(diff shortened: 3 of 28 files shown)


src/main/java/net/datafaker/providers/base/BaseFaker.java
```diff
@@ -400,4 +400,9 @@ public final <B extends ProviderRegistration> B getFaker() {
     public static Method getMethod(AbstractProvider<?> ap, String methodName) {
         return ap == null ? null : ObjectMethods.getMethodByName(ap, methodName);
     }
+
+    @Override
+    public String toString() {
```
bodiam, line 405: isn't this the default toString?

src/main/java/net/datafaker/providers/base/Image.java
```diff
@@ -158,8 +158,7 @@ private String generateBase64RasterImage(ImageType imageType, int width, int hei
             byte[] imageBytes = baos.toByteArray();
             return "data:" + imageType.mimeType + ";base64," + Base64.getEncoder().encodeToString(imageBytes);
         } catch (IOException e) {
-            e.printStackTrace();
-            return null;
+            throw new RuntimeException("Failed to generate image %s of size %sx%s".formatted(imageType, width, height), e);
```
bodiam, line 161: I don't see why you'd want to throw an exception here.

src/main/java/net/datafaker/providers/base/Internet.java
```diff
@@ -246,41 +246,29 @@ public String macAddress() {
      * @return a correctly formatted IPv4 address.
      */
     public String ipV4Address() {
-        try {
-            return getIpV4Address().getHostAddress();
-        } catch (UnknownHostException e) {
-            return "127.0.0.1";
-        }
+        return getIpV4Address().getHostAddress();
```
bodiam, line 249: I don't think this is ideal, I believe the code can throw exceptions now sometimes, and I don't think we should.

src/main/java/net/datafaker/providers/base/Internet.java
```diff
@@ -246,41 +246,29 @@ public String macAddress() {
...
     /**
      * @return a valid private IPV4 address in dot notation
      */
     public String privateIpV4Address() {
-        try {
-            return getPrivateIpV4Address().getHostAddress();
-        } catch (UnknownHostException e) {
-            return "127.0.0.1";
-        }
+        return getPrivateIpV4Address().getHostAddress();
```
bodiam, line 265: again, I don't think this is correct.

src/main/java/net/datafaker/providers/base/Internet.java
```diff
@@ -343,19 +327,15 @@ public String ipV4Cidr() {
      * @return a correctly formatted IPv6 address.
      */
     public String ipV6Address() {
-        try {
-            return getIpV6Address().getHostAddress();
-        } catch (UnknownHostException e) {
-            return "0:0:0:0:0:0:0:1";
-        }
+        return getIpV6Address().getHostAddress();
```
bodiam, line 330: same as above.

src/main/java/net/datafaker/providers/base/BaseFaker.java
asolntsev: Almost. Except that it's shorter due to `getSimpleName` instead of `getName`. 
I would like to make it even more shorter, but it doesn't have any human-readable fields...

src/main/java/net/datafaker/providers/base/Image.java
asolntsev: Because it's much better to clearly show the problem instead of hiding it.
* When you return null, it almost always causes NullPointerException in the next step. And user will not understand what happened. For example, in DF own test: 
```java
        assertThat(faker.image().base64TIFF()).startsWith("data:image/tiff;base64,");
```

* If user wants to protect, he will need to write IF:

```java
String image = faker.image().base64TIFF();
if (image == null) {
  // WHAT SHOULD I DO HERE?..
  throw new RuntimeException("WTF, the image is null, and I have no idea why.");
}
```

src/main/java/net/datafaker/providers/base/Internet.java
asolntsev: I also had a doubt, but no, this code doesn't fail.

This is a test that tries through all possible values:
```java
    @Test
    void allPossibleAddresses() {
        for (int first = 2; first < 256; first++) {
            for (int second = 2; second < 256; second++) {
                for (int third = 2; third < 256; third++) {
                    for (int fourth = 2; fourth < 256; fourth++) {
                        InetAddress inetAddress = inet4Address((byte) first, (byte) second, (byte) third, (byte) fourth);
                        assertThat(inetAddress).isNotNull();
                    }
                }
            }
        }
    }
```

src/main/java/net/datafaker/providers/base/Image.java
bodiam: How is writing that if(null) check different from:

```
try {
   String image = faker.image().base64TIFF();
} catch(Exception e) {
  throw new RuntimeException("WTF, the image generation failed");
}
```

src/main/java/net/datafaker/providers/base/BaseFaker.java
bodiam: but why do you need it?

src/main/java/net/datafaker/providers/base/Image.java
asolntsev: It differs a lot.
* User doesn't need to check if image is null. 
* User doesn't need to pollute his code with unneeded IFs and try/catches. 
* User will see a clear error message explaining WHY the image could not be generated. 

> The implicit contract in DF is that none of the methods will ever throw an exception

It sounds very strange to me. 
Nobody can ever guarantee that his code never throws any exceptions. 

Anyway, in this case, such an exception actually cannot happen. We need to catch `IOException` only because java interfaces like `InputStream` declare `throws IOException`. But in this case we operate with `ByteArrayOutputStream` which **never throws `IOException`**.

src/main/java/net/datafaker/providers/base/Internet.java
asolntsev: same as above. :)

src/main/java/net/datafaker/providers/base/Internet.java
asolntsev: same as above: it's correct :)

asolntsev:
@bodiam Let me summarize: I am pretty sure it's a good PR and should be merged. ;)

bodiam:
Then feel free to merge it! :)

PR #1315, opened 2024-07-29
Add Dota 2 VideoGameProvider

panic08:
All data was taken from open sources (like dota2.fandom.com).

There was already a dota.todo.yml concept, I changed the data and added the data in the afternoon, added a couple more methods to the new provider itself


src/main/java/net/datafaker/providers/videogame/Dota2.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public class Dota2 extends AbstractProvider<VideoGameProviders> {
+
+    protected Dota2(VideoGameProviders faker) {
+        super(faker);
+    }
+
+    public String fraction() {
+        return resolve("games.dota2.fraction");
+    }
```
kingthorin, line 19: Do you mean "faction" ?

src/main/resources/en/dota2.yml
kingthorin, line 1: Nice catch. I hadn't even noticed the todo yaml files yet.

I'll try to get cracking on some of these.

src/main/java/net/datafaker/providers/videogame/Dota2.java
panic08: Yes, fraction/faction

https://dota2.fandom.com/wiki/Factions

kingthorin: Are they called "fractions" in the game (with an R)?

Because I didn't see it on that Fandom page.

I'm happy to leave it if it's topical, I just don't know. I'm not a DOTA player.

panic08: Yes, you're right. I've now manually checked it myself and indeed “faction” is correct

kingthorin (dismissed):
LGTM

kingthorin (approved):
Thanks!

kingthorin:
I was going to suggest something like:
```java
    @Test
    void dota2Quote() {
        // Given / When
        String quote = dota2.heroQuote(dota2.hero());
        // Then
        assertThat(quote).isNotEmpty();
    }
```

 but it seems there aren't quotes for every hero. I guess you could make a list of those that do have quotes and then randomly pick from that Set for the test.

panic08:
Also, in the future, when quotes are added for all heroes from the game itself, we should get rid of “availableHeroes” in Dota2Test itself and change the dota2HeroQuote test method to take heroes from dota2.hero().

kingthorin:
@bodiam how sticky (or not) do you want to be about the codecov stuff?
image

bodiam:
@kingthorin  not very, it was more indicative than anything else. I'm not sure why it breaks the build these days when the coverage goes down a bit, that wasn't intended.

kingthorin:
Thanks. Merging as-is 🙂

PR #1316, opened 2024-08-04
generate ID Number together with birthday and gender

asolntsev:
this commit adds method `IdNumber.valid(IdNumberRequest)` which returns a valid combination of ID number, birthday and gender.

### Usage:
```java
        PersonIdNumber person = faker.idNumber().valid(new IdNumberRequest(18, 96, ANY));
        PersonIdNumber person = faker.idNumber().valid(new IdNumberRequest(25, 45, FEMALE));
```

### Problem to solve:
In many countries, ID Number contains or depends on birthday and/or gender.
We need to generate a valid person containing a valid combination of ID number, birthday and gender matching each other. 

Until now, it was possible to generate them separately, but they didn't always match each other:

```java
        String idNumber = faker.idNumber().valid();  // contains birthday of 80 old man
        LocalDate birthday = faker.timeAndDate().birthday();  // birthday of 16 old woman
```

### DIFF
--- a/src/main/java/net/datafaker/idnumbers/MacedonianIdNumber.java
+++ b/src/main/java/net/datafaker/idnumbers/MacedonianIdNumber.java
@@ -1,9 +1,17 @@
 package net.datafaker.idnumbers;
 
 import net.datafaker.providers.base.BaseProviders;
+import net.datafaker.providers.base.IdNumber.IdNumberRequest;
+import net.datafaker.providers.base.PersonIdNumber;
+import net.datafaker.providers.base.PersonIdNumber.Gender;
 
 import java.time.LocalDate;
 import java.util.List;
+
+import static net.datafaker.idnumbers.Utils.gender;
+import static net.datafaker.idnumbers.Utils.birthday;
+import static net.datafaker.idnumbers.Utils.randomGender;
+import static net.datafaker.providers.base.PersonIdNumber.Gender.FEMALE;
 
 /**
  * The Macedonian Identity Number is a unique personal identification number of 13 digits in a form "DD MM YYY RR BBB K"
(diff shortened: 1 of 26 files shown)


bodiam:
That looks good to me. @kingthorin , any concerns / feedback?

kingthorin:
I'll try to review this in the next few hours.

src/main/java/net/datafaker/idnumbers/MacedonianIdNumber.java
```diff
@@ -64,8 +74,8 @@ private String rr(BaseProviders faker) {
      * - from 000 to 499 for the male, and
      * - from 500 to 999 for the female citizens.
      */
-    private String sss(BaseProviders faker, boolean female) {
-        int ordinal = female ? faker.number().numberBetween(500, 1000) : faker.number().numberBetween(0, 500);
+    private String sss(BaseProviders faker, Gender gender) {
+        int ordinal = gender == FEMALE ? faker.number().numberBetween(500, 1000) : faker.number().numberBetween(0, 500);
```
kingthorin, line 78: I know it was like this before but is 500 really valid for both?

kingthorin (commented):
Only see the one item.

I also predict that we'll have to allow other genders or "prefer not to say" in the future, but I guess that doesn't matter until countries start publishing new specs.

kingthorin:
My point was more that it should probably be
0-499
500-1000
Or
0-500
501-1000

If it's supposed to be gender specific how can 500 be valid in both cases?

asolntsev:
> we'll have to allow other genders

Here we are talking only about "genders used for ID numbers". As far as I know, all countries either don't use gender at all, or use MALE/FEMALE option.

> If it's supposed to be gender specific how can 500 be valid in both cases?

Yes, this is a valid code because `numberBetween` function has `(minInclusive, maxExclusive)` parameters. 
In other words, 
*`numberBetween(500, 1000)` means `500 <= X < 1000`
*`numberBetween(1, 500)` means `1 <= X < 500`.

kingthorin:
Okay sorry, thought it was inclusive

kingthorin (approved):

bodiam:
@asolntsev want to merge this one?

asolntsev:
Sure, yes. :)

PR #1319, opened 2024-08-05
Provides financial terms.

bodiam:
I needed something like this to create events, so now it can be used to generate this:

```
ExportLapseRequested
PolicyCardIssued
ReversalAlertCompleted
ATMCompletionExpired
AlertClaimUpdated
DeclineBreachIssued
AccessDeclinePending
ATMTransactionPending
```


kingthorin:
Add it to the existing Finance provider?

https://github.com/datafaker-net/datafaker/blob/main/src/main/java/net/datafaker/providers/base/Finance.java

kingthorin:
Other than that, it looks good to me.

PR #1332, opened 2024-08-22
simplify EnFile: don't collect a static list that is needed only once

asolntsev:


src/main/java/net/datafaker/service/files/EnFile.java
```diff
@@ -260,7 +260,7 @@ public String getPath() {
         "zodiac.yml").map(EnFile::new).toList();
 
     // files where the search path can't be derived from the filename
-    private static final List<EnFile> FILES_WITH_A_DIFFERENT_PATH = Arrays.asList(
+    private static final List<EnFile> FILES_WITH_A_DIFFERENT_PATH = asList(
```
kingthorin, line 263: This can be immutable can't it? Being a CONSTANT certainly implies that's the case. 
If so why not use List.of and not import anything from the Arrays package?

asolntsev: @kingthorin Good point. Done.

kingthorin (approved):

PR #1333, opened 2024-08-22
convert class Interval to record

asolntsev:


kingthorin (approved):
I haven't really done any work with records yet. This is a nice simplification.

PR #1335, opened 2024-08-23
Refactor FakeValues

asolntsev:


src/main/java/net/datafaker/providers/base/Company.java
```diff
@@ -36,40 +36,27 @@ public String profession() {
...
-                lock.unlock();
-            }
-        }
-        return allBuzzwords.get(faker.random().nextInt(allBuzzwords.size()));
+        return faker.options().nextElement(allBuzzwords.get());
+    }
+
+    private List<String> loadBuzzwords() {
+        List<List<String>> buzzwordLists = faker.fakeValuesService().fetchObject("company.buzzwords", faker.getContext());
+        return buzzwordLists.stream().flatMap(Collection::stream).toList();
```
kingthorin, line 44: Why stream it then map it to a stream then collect it? Why not just:
`buzzwordLists.stream().collect(Collectors.toList())`

asolntsev: no, `buzzwordLists.stream().collect(Collectors.toList())` is the same as just `buzzwordLists`.

The point is that `buzzwordLists` is **list of lists**. And we need to **convert it to a list**.

kingthorin: Okay, devil in the detail 👿

kingthorin (approved):

PR #1342, opened 2024-08-28
Add tier() and neutralItem() methods in Dota2 VideoGameProvider and update data

panic08:
- Added tier() and neutralItem() methods in Dota2.class
- Updated the data in dota2.yml
- Updated the test for dota2HeroQuote()

### DIFF
--- a/src/main/resources/en/dota2.yml
+++ b/src/main/resources/en/dota2.yml
@@ -92,7 +92,7 @@ en:
         - "Morphling"
         - "Muerta"
         - "Naga Siren"
-        - "Natures Prophet"
+        - "Nature`s Prophet"
         - "Necrophos"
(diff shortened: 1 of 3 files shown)


src/main/resources/en/dota2.yml
```diff
@@ -92,7 +92,7 @@ en:
         - "Morphling"
         - "Muerta"
         - "Naga Siren"
-        - "Nature's Prophet"
+        - "Natures Prophet"
```
kingthorin, line 95: I'm pretty sure possessive was correct, not plural.

kingthorin: Oh okay

kingthorin (approved):

Issue #1343, opened 2024-08-29
Faker().locality().allSupportedLocales() returns empty when Datafaker is embedded in Spring Boot Uber JAR


jloisel:
Faker().locality().allSupportedLocales() returns empty when Datafaker is embedded in Spring Boot Uber JAR.

kingthorin:
Please provide way more detail.

jloisel:
There is not much more details needed:
1. Create a Spring Boot App,
2. Add DataFaker as dependency,
3. Package app with spring boot plugin: https://docs.spring.io/spring-boot/maven-plugin/getting-started.html

Given the JAR is within another JAR, within `BOOT-INF/libs`, it's unable to find any locale:
https://docs.spring.io/spring-boot/specification/executable-jar/nested-jars.html

```
example.jar
 |
 +-META-INF
 |  +-MANIFEST.MF
 +-org
 |  +-springframework
 |     +-boot
 |        +-loader
 |           +-<spring boot loader classes>
 +-BOOT-INF
    +-classes
    |  +-mycompany
    |     +-project
    |        +-YourClasses.class
    +-lib
       +-dependency1.jar
       +-dependency2.jar
```

jloisel:
You can create a spring boot app using:
https://start.spring.io/

[demo.zip](https://github.com/user-attachments/files/16816900/demo.zip)

Make sure to add datafaker jar as dependency in pom.xml

bodiam:
Hi @jloisel , thanks for your report. I just had a look at the code, and yes, it's unlikely that this code will work with an uber jar indeed. What do you need this method for? If you have a suggestion or a PR on how to fix it, happy to accept it.

kingthorin:
Why not try whatever locales you need and have a reasonable fallback or error handling. Instead of trying to enumerate on either side?

jloisel:
It's not a feasable approach to have a fallback / error handler for every external library called from your code. We cannot suppose for every method being call that "it may not work". Every class / method should have a contract we can trust, otherwise it's of no use.

It's fine by me if you don't want to fix this issue. But, from now, you are aware that listing locales from inside any Spring Boot / Uber Jar application doesn't work at all. At least, I would suggest you document this issue / limitation so future users are aware of it.

bodiam:
[...]

PR #1350, opened 2024-09-12
#1343 simplify code for returning all available locales

asolntsev:
Now we don't scan classpath at all, but just return a hardcoded list of locales. Instead, we now have a unit-test that verifies that this list is up-to-date.

Related to:
- datafaker-net/datafaker#1343 
- datafaker-net/datafaker#1349

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Locality.java
+++ b/src/main/java/net/datafaker/providers/base/Locality.java
@@ -15,7 +15,7 @@ import java.util.Set;
  * @since 1.7.0
  */
 public class Locality extends AbstractProvider<BaseProviders> {
-    private static final List<String> locales = List.of(
+    private static final List<String> LOCALES = List.of(
         "_al", "_bg", "_by", "_ca", "_ch", "_cn", "_cz", "_ee", "_ge", "_md", "_mk", "_ru", "_us",
         "ar", "be", "bg", "by", "ca", "ca-cat", "cs", "cs-cz",
         "da-dk", "de", "de-at", "de-ch",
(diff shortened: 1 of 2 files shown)


src/main/java/net/datafaker/providers/base/Locality.java
```diff
@@ -2,39 +2,32 @@
...
 
 /**
  * Generates random locales in different forms.
  *
  * @since 1.7.0
  */
 public class Locality extends AbstractProvider<BaseProviders> {
-    private static final Logger log = Logger.getLogger(Locality.class.getName());
-    private static final Predicate<String> SCAN_ALL_JARS = name -> true;
+    private static final List<String> locales = List.of(
```
kingthorin, line 18: Should be "LOCALES"

kingthorin: Was the list generated programmatically? (If so it might be worth commenting about that for future maint)

asolntsev: 1. Renamed to `LOCALES`.
2. I added a hint to error message, it should help.

For example, if I add a file `src/main/resources/_FOO.yml`, but forget to add `_foo` to constant `Locality.LOCALES`, then one test will fail with a very clear instruction:

```java
java.lang.AssertionError: [Apparently, somebody forgot to add the new locale to Locality.LOCALES] 
Expecting actual:
  ["_al", "_bg",    .....]
to contain exactly in any order:
  ["ro-md", ... ]
but could not find the following elements:
  ["_foo"]
```

kingthorin: Great, okay!

PR #1375, opened 2024-10-12
Fix for K-pop artist lists: H.O.T., Fin.K.L, Sunmi, Stray Kids

badoken:
H.O.T., Fin.K.L - were missed in the generations list
Sunmi, Stray Kids - were missed in all the lists


kingthorin (approved):
Thanks

PR #1376, opened 2024-10-13
Added initial support for GraalVM Native images.

bodiam:


asolntsev (approved):
I have no idea what's happening here, but looks good. :)

bodiam:
GraalVM likes statically compiled things, and doesn't like reflection and resource loading much, since it's too dynamic. So when you use Graal with a library which depends on dynamic resource loading, you have to tell Graal which resources will be used. Since that's quite tedious, you can use some tool to generate these mappings for you, and bundle then with the library, so Graalvm doesn't have to be configured anymore.

It could be that I missed a few files though, so no promises that it works 100%!

PR #1379, opened 2024-10-15
move India country-specific settings to file "_IN.yml"

asolntsev:
these are valid in all languages, not only in English.

NB! Rename "IND" to "IN" because this is the valid 2-letter India country code.


kingthorin (approved):

kingthorin (approved):

kingthorin:
Thanks!

PR #1381, opened 2024-10-16
#1306 fix non-thread-safe usages of CopyOnWriteMap

asolntsev:
Instead of two-three separate calls like `containsKey`, `get` and `put`, now we use a single `computeIfAbsent`.

P.S. I would like to also deny method `CopyOnWriteMap.put()`, but it's still used in one place (that was harder to refactor).


kingthorin (approved):
This looks good to me, thanks for digging into the problem.

Issue #1386, opened 2024-10-17
🐞 Random service has a bug with boundaring values


ylazakovich:
**Describe the bug**

After bumping to `2.4.0` started reproducing an issue with using the following method 
`random().nextInt(1, Integer.MAX_VALUE)`

2.3.1 - works fine
2.4.0 - has a bug

**To Reproduce**

❌ FAILED
```java
    @Test
    void testFaker() {
        assertThat(faker.random().nextInt(1, Integer.MAX_VALUE)).isNotNull();
    }
```

✅ PASSED
```java
    @Test
    void testFaker() {
        assertThat(faker.random().nextInt(1, Integer.MAX_VALUE - 1)).isNotNull();
    }
```

**Versions:**
 - JDK `21-coretto`
 - Faker Version `2.4.0`

**Details**

```log
java.lang.RuntimeException: java.lang.IllegalArgumentException: bound must be greater than origin

	at org.testng.internal.invokers.MethodInvocationHelper.invokeMethodNoCheckedException(MethodInvocationHelper.java:57)
	at org.testng.internal.invokers.MethodInvocationHelper.invokeDataProvider(MethodInvocationHelper.java:157)
	at org.testng.internal.Parameters.handleParameters(Parameters.java:803)
	at org.testng.internal.Parameters.handleParameters(Parameters.java:728)
	at org.testng.internal.invokers.ParameterHandler.handleParameters(ParameterHandler.java:72)
	at org.testng.internal.invokers.ParameterHandler.createParameters(ParameterHandler.java:51)
	at org.testng.internal.invokers.TestInvoker$MethodInvocationAgent.invoke(TestInvoker.java:911)
	at org.testng.internal.invokers.TestInvoker.invokeTestMethods(TestInvoker.java:200)
	at org.testng.internal.invokers.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:148)
	at org.testng.internal.invokers.TestMethodWorker.run(TestMethodWorker.java:128)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.IllegalArgumentException: bound must be greater than origin
[...]

PR #1388, opened 2024-10-17
#1386 fix nextInt and nextLong with extreme boundaries

asolntsev:
e.g.
* `randomService.nextInt(Integer.MIN_VALUE, Integer.MAX_VALUE)`
* `randomService.nextLong(Long.MIN_VALUE, Long.MAX_VALUE)`

### DIFF
--- a/src/main/java/net/datafaker/providers/base/DateAndTime.java
+++ b/src/main/java/net/datafaker/providers/base/DateAndTime.java
@@ -322,8 +322,8 @@ public class DateAndTime extends AbstractProvider<BaseProviders> {
     /**
      * Generates a random Duration between min and max.
      *
-     * @param min  the maximum value
-     * @param max  the minimal value
+     * @param min  the minimum value
+     * @param max  the maximum value
      * @param unit the temporal unit (day or shorter than a day)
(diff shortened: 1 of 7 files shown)


kingthorin:
Drop the unrelated workflow change?

kingthorin:
To be clear I do think that's probably a handy change if you want to submit it separately.

(I'll leave that approval/merge up to bodiam but for my two cents it's good)

src/main/java/net/datafaker/providers/base/DateAndTime.java
```diff
@@ -322,8 +322,8 @@ public Duration duration(long max, ChronoUnit unit) {
     /**
      * Generates a random Duration between min and max.
      *
-     * @param min  the maximum value
-     * @param max  the minimal value
+     * @param min  the minimum value
+     * @param max  the maximum value
```
kingthorin, line 326: Ha good catch, one of obviously didn't notice the mismatch.

kingthorin (approved):

asolntsev:
> that's probably a handy change if you want to submit it separately

I am not sure if it's useful. I wanted looked for a github action that would generate a readable test report.
Because usually it's hard to find which test failed (when a build fails in Github actions).
But this specific action didn't help a lot.

PR #1392, opened 2024-10-19
Added numberBetween for doubles.

bodiam:


kingthorin:
https://github.com/datafaker-net/datafaker/blob/12bdd2af5d3e0deae8287b916d82769f07e190ed/src/test/java/net/datafaker/FakerTest.java#L211-L215

bodiam:
> numberBetweenRepeated

Interesting. What should we do with this? I think what I'm basically asking is:
`"#{number.number_between '1','10'}"`, should/could that return `1.5`?

kingthorin:
Yes in this case it could be.

bodiam:
> Yes in this case it could be.

I agree. I'm just not sure if that's going to break anything. Probably not likely, but not sure. Any suggestions on what to do here? Reject the PR, fix the test, something else?

kingthorin:
I see two options.
1) Adjust the regex
2) Make the double method specific instead of `numberBetween` maybe `doubleBetween`.

src/main/java/net/datafaker/providers/base/Number.java
```diff
@@ -59,6 +59,16 @@ public int numberBetween(final int min, final int max) {
...
     }
 
+    /**
+     * @param min the lower bound (include min)
+     * @param max the upper bound (not include max)
+     * @return a random number on faker.number() between min and max
+     * if min = max, return min
+     */
+    public double numberBetween(double min, double max) {
+        return faker.random().nextDouble(min, max);
```
snuyanzin, line 69: it looks like nothing prevents from calling it like
```java
faker.number().numberBetween(123.456, 1.0);
```
what is expected output of it?
based on name the result should be somewhere between the boudaries

bodiam: The same as what the other methods return.

```
faker.number().numberBetween(10, 0) 
```

Also returns a number between 0 and 10.

bodiam:
@kingthorin 

> Adjust the regex

I did.

src/main/java/net/datafaker/providers/base/Number.java
snuyanzin: would be great to have a test for this case as well

bodiam: Fair enough, I added it.

PR #1402, opened 2024-10-24
fix generate sql will be truncate when quote char in word

shanjuren:
Some thesaurus`s word will be truncate when I use datafaker to generate SQL, like Address,Book... That reason is when SqlTransformer calling the handledObjectToString method will be one by one add char to new Object ,but if miss quote char will be stop.

### DIFF
--- a/src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
+++ b/src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
@@ -219,7 +219,7 @@ public class SqlTransformer<IN> implements Transformer<IN, CharSequence> {
     }
 
     private String handledObjectToString(int length, String strValue) {
-        StringBuilder builder = new StringBuilder(32);
+        StringBuilder builder = new StringBuilder(length);
 
         for (int k = 0; k < length; k++) {
             if (strValue.charAt(k) == quote) {
--- a/src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
+++ b/src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
@@ -219,19 +219,16 @@ public class SqlTransformer<IN> implements Transformer<IN, CharSequence> {
     }
 
     private String handledObjectToString(int length, String strValue) {
-        StringJoiner joiner = null;
+        StringBuilder builder = new StringBuilder();
 
-        int j = 0;
         for (int k = 0; k < length; k++) {
             if (strValue.charAt(k) == quote) {
-                if (joiner == null) {
-                    joiner = new StringJoiner("" + quote);
-                }
-                joiner.add(strValue.substring(j, k + 1));
-                j = k + 1;
+                builder.append('\\').append(quote);
+            }else {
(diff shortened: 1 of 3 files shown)


asolntsev (approved):

asolntsev (commented):
@shanjuren Could you please add a test? Probably in `SqlTest.java`

src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
```diff
@@ -219,19 +219,16 @@ private String handleObject(Object value) {
     }
 
     private String handledObjectToString(int length, String strValue) {
-        StringJoiner joiner = null;
+        StringBuilder builder = new StringBuilder();
```
kingthorin, line 222: May as well set the builder size to length to start with, that will probably mean less resizing

shanjuren:
> @shanjuren Could you please add a test? Probably in `SqlTest.java`

OK，i will try do it

src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
shanjuren: yes ,It seems more reasonable to set the length

shanjuren: is done

shanjuren:
is done

src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
```diff
@@ -219,7 +219,7 @@ private String handleObject(Object value) {
     }
 
     private String handledObjectToString(int length, String strValue) {
-        StringBuilder builder = new StringBuilder();
+        StringBuilder builder = new StringBuilder(32);
```
kingthorin, line 222: I meant literally the length being passed in.

src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
```diff
@@ -219,19 +219,16 @@ private String handleObject(Object value) {
     }
 
     private String handledObjectToString(int length, String strValue) {
-        StringJoiner joiner = null;
+        StringBuilder builder = new StringBuilder(32);
```
bodiam, line 222:
```suggestion
        StringBuilder builder = new StringBuilder(length);
```

src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
bodiam: I've made the change, thank you!

src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
bodiam: done.

bodiam:
it's failing spotless, can you please run the maven command to fix it?

bodiam:
I fixed up the PR, I'll merge it in.

Issue #1405, opened 2024-10-25
Junit 5 extension for injecting Faker object at test methods


agomezlucena:
Hi I used for some months this project and I saw that sometimes we need to have some boilerplate code for use it.
Right now we must to create always an instace and setup manually the injection of the Faker object in our test cases like this

```java
class ExampleTestClass {
    private final Faker faker = new Faker();
    //some tests
}
```
or 

```java
class ExampleTestClass {
    @Test
    void exampleTest(){
        var faker = new Faker();
        var givenIsbn = faker.code().isbn13(false);
        //test logic
    }
}
```
But this could be repetitive and make that the logic of the test a litte bit messy.

The Idea is to create a Junit5 extension that allows you to inject a shared faker instance in the test with something like this:

```java
@ExtendWith(DataFakerExtension.class)
class ExampleTestClass {
    @Test
    void exampleTest(Faker faker){
        var givenIsbn = faker.code().isbn13(false);
        //test logic
   }
}
```

I think that this kind of extension would improve the legibility of the tests and the usability of the library.

asolntsev:
@agomezlucena Why not just put this `Faker` instance to a static variable?

bodiam:
I like the idea in theory, it's actually a long time on my list to have better Junit 5 integration. However, at the same time, I'm not sure if Junit should be included in Datafaker, since we like to keep the core as free from dependencies as we can, and it probably wouldn't be ideal to tie it a specific test engine. 

Perhaps a module outside of the project would be more suitable at this moment? (datafaker-junit5 module or so, so that perhaps we can have a datafaker-testng or similar module in the future?)

kingthorin:
Couldn't you just create a class that extends Faker and has a constant or singleton you can re-use instead of numerous instantiations?

bodiam:
I don't think there's any reason why you couldn't just create a class level instance of Datafaker, and reuse that over the tests. It's a one line thing, I wouldn't use a faker per test case, and I'm not a fan of using Faker directly in a test class either, I use object mothers in 99% of the time.

But on the other hand, I also wouldn't mind having something like "DatafakerSource" similar to CsvSource etc for parametized tests. It can be done by creating a MethodSource, but I always have to look up the correct syntax. So I was hoping that maybe a junit5 project could be the start of some native integration, and we'll see where it takes us?

agomezlucena:
> @agomezlucena Why not just put this `Faker` instance to a static variable?

[...]

PR #1409, opened 2024-10-26
Created Junit5 extension for datafaker

agomezlucena:
https://github.com/datafaker-net/datafaker/issues/1405


datafaker-junit-jupiter-extension/src/main/java/net/datafaker/extensions/junit/JunitJupiterDataFakerExtension.java
```diff
@@ -0,0 +1,42 @@
...
+ *         var isbn = faker.code().isbn13();
+ *         //insert test logic.
+ *     }
+ * }
+ * </code></pre>
+ *
+ * @author Alejandro Gómez Lucena
+ */
+public class JunitJupiterDataFakerExtension implements ParameterResolver {
+    private final Faker faker = new Faker();
```
snuyanzin, line 25: how about seed and locale?
what if user wants japanese names?

PR #1411, opened 2024-10-27
fix: University tests

kingthorin:
Extended the expected regex to include a few characters with accents.

Addresses https://github.com/datafaker-net/datafaker/pull/1404#issuecomment-2436479028

~~Add recently introduced test artifact `test.txt` to `.gitignore`~~

FYI if anyone needs to run single UnitTest classes or methods locally you can use:
` ./mvnw test -Dtest=UniversityTest`
` ./mvnw test -Dtest=UniversityTest#testName`
syntax.

### DIFF
--- a/src/test/java/net/datafaker/providers/base/UniversityTest.java
+++ b/src/test/java/net/datafaker/providers/base/UniversityTest.java
@@ -9,11 +9,12 @@ import java.util.Collection;
 
 class UniversityTest extends BaseFakerTest<BaseFaker> {
 
+    private static final String UNIVERSITY_MATCHER = "[A-Za-z'() öèü-]+";
     private final University university = faker.university();
 
     @Test
     void testName() {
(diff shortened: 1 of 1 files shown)


asolntsev (changes requested):
1. Why it's needed to ignore file "test.txt"? I don't see any tests that would create such a file.
2. This regex is not enough. UniversityTest still can fail:
```java
java.lang.AssertionError: 
Expecting actual:
  "North Leffler Academy of Aix-Marseille"
to match pattern:
  "[A-Za-z'() öèü]+"
```

kingthorin:
Oops thanks, I didn’t account for hyphen or other chars, I had searched non-ASCII.

I’ll dig into the test.txt thing, maybe it was something I’d polluted the repo with somehow.

kingthorin:
@asolntsev fixed.

asolntsev (approved):

Issue #1314, opened 2024-07-25
Add weighted selection 


gatear:
Support weighted faking in custom providers.

Inspired from https://fakerjs.dev/api/helpers#weightedarrayelement
```javascript
faker.helpers.weightedArrayElement([
{ weight: 5, value: 'sunny' }, 
{ weight: 4, value: 'rainy' },
{ weight: 1, value: 'snowy' }
]) // 'sunny', 50% of the time, 'rainy' 40% of the time, 'snowy' 10% of the time
```

With custom providers, we could make a YML provider with weights associated to each value
```yml
en:
  faker:
    insectsfromfile:
      ants: [5, 'Driver ant', 4, 'Fire ant', 1, 'Harvester ant']

```
Questions
- Is there a way to achieve weighted faking with current codebase ?

bodiam:
Hi @gatear , I think there is a way, but it's a very poor way: I think by repeating the element a few times, you could achieve the same, though perhaps the fakerjs way is a bit more elegant?

gatear:
Agreed, it's a good workaround. Actually used something like this today.
```yaml
en:
  faker:
    insectsfromfile:
      ants: ['Driver ant', 'Harvester ant', 'Harvester ant', 'Harvester ant'] 
```
So in theory we would have 75% chances of getting 'Harvester ant'.
That being said I think it's valuable to offer an explicit way of configuring this.

SamuelNeceda:
Hi @bodiam , I would like to take care of this issue. Can it be assigned to me, please?

bodiam:
@SamuelNeceda  thanks for you offer, it would be great to have this feature in. If you need help or want to share some thoughts, feel free to reach out!

kingthorin:
I believe you'd add your docs to: https://github.com/datafaker-net/datafaker/blob/main/docs/documentation/custom-providers.md

snuyanzin:
thanks for driving this
I still see some issues/open questions here 
1. how to limit the total weight or how to cope with overflow?
2. how can we be sure that `weight` is not used for other purpose by custom provider? Ans if it is used what could/should be done? Especially if a user is not an owner/maintainer of this custom provider
3. Any new property and extra reading/boxing/unboxing means more processing resources
4. If it is related to only some special providers then I would vote not to add it as a functionality for all, instead extract it in a separate interface like WeightedProviders and add functionality only for them.
Otherwise the question would be: why do we need a generic feature which is failing for 99% of providers?

[...]

PR #1414, opened 2024-10-30
Addition of weighted selection to custom providers

SamuelNeceda:
As for the [Issue](https://github.com/datafaker-net/datafaker/issues/1314), I have implemented the capability for weighted selection with custom providers.

datafaker-net/datafaker#1314


kingthorin (commented):
What happens in the case of an empty or non-double weight?

(I haven't pulled the code and tried).

I'd also suggest creating a test YAML file and testing as you see users actually using it. (Which can likely be re-used for documentation later)

SamuelNeceda:
@kingthorin , great catch on the edge cases for weights! I’ve implemented a solution that should address the necessary scenarios.

Additionally, in [this file](https://github.com/datafaker-net/datafaker/pull/1414/files#diff-6b867d96f22fbd6485e9d317a9dffd8a7870afd9c37656cdf53b19ea5a5f8d58), I demonstrated the weighted selection with a custom hardcoded provider.

Currently, I’m working on implementing weighted selection using a custom provider that loads data from a YAML file. I’ve created the file, but I’m facing challenges loading the data to apply the weighted selection. Following the logic from the ant() method, it looks like the key changes might be needed in AbstractProvider and FakeValueService classes, but I'm unsure how to proceed. Could you please provide any recommendations?

kingthorin (dismissed):

src/main/java/net/datafaker/service/RandomService.java
```diff
@@ -141,6 +145,84 @@ public Random getRandomInternal() {
...
+        if (weight <= 0 || Double.isNaN(weight) || Double.isInfinite(weight)) {
+            throw new IllegalArgumentException("Weight must be a positive number and cannot be NaN or infinite");
+        }
+    }
+
+    private double calculateTotalWeight(List<Map<String, Object>> items) {
+        return items.stream()
+            .mapToDouble(item -> (Double) item.get(WEIGHT_KEY))
+            .sum();
+    }
```
snuyanzin, line 210: How will it work if 
1. there are N weights each of them is Double.MAX_VALUE - 1 or something like that, are we protected against overflow?
2. what if there are custom fields with same name however with different meaning in same provider and even number type? Will there at least any warning appear  or how should user be warned about this (without hours of debugging)

src/main/java/net/datafaker/service/RandomService.java
```diff
@@ -141,6 +145,84 @@ public Random getRandomInternal() {
...
+
+    private <T> T selectWeightedElement(List<Map<String, Object>> items, double randomValue) {
+        double currentWeight = 0.0;
+
+        for (Map<String, Object> item : items) {
+            currentWeight += (Double) item.get(WEIGHT_KEY);
+            if (randomValue < currentWeight) {
+                return (T) item.get(VALUE_KEY);
+            }
+        }
```
snuyanzin, line 220: is there a reason to do this on each iteration?
IIUC data in file is persistent then we could sort them once including all unboxing things (not sure if need them...) only once and have something like array of doubles
then just use something like adopted binary search

or did I miss anything here?

src/test/java/net/datafaker/utils/TestConstants.java
kingthorin, line 1: Is there a reason these can't be public in the appropriate test class?

SamuelNeceda: I moved the constants to the appropriate test class and removed TestConstants.java. Initially, they were placed in a separate utils package to make them accessible across multiple classes, but since they are only used here, keeping them in the test class simplifies the structure.

src/main/java/net/datafaker/service/RandomService.java
SamuelNeceda: 1. I appreciate your insight regarding potential overflow. While weights are typically not expected to be extremely large, I've implemented an iterative approach to ensure that adding each weight is checked against Double.MAX_VALUE. This helps prevent overflow in cases where unusually large weights might occur.
2. I believe that values should be unique when performing weighted selection. Therefore, while weights can be the same, duplicates in the value parameter trigger an exception that informs the user of the specific duplicate value. This approach seems appropriate to ensure the integrity of the selection process.

src/main/java/net/datafaker/service/RandomService.java
SamuelNeceda: Thank you for the suggestion. I implemented the proposed optimization of sorting and preprocessing the cumulative weights once, which improved the efficiency.

src/main/java/net/datafaker/service/RandomService.java
```diff
@@ -1,14 +1,24 @@
...
+    private static final String WEIGHT_KEY = "weight";
+    private static final String VALUE_KEY = "value";
     private static final char[] HEX_UP = "0123456789ABCDEF".toCharArray();
     private static final char[] HEX_LOWER = "0123456789abcdef".toCharArray();
     private static final Random SHARED_RANDOM = new Random();
     private final Random random;
 
+    private double[] cumulativeWeights;
+    private Object[] values;
+
```
snuyanzin, line 21: Definitively it should be outside of this class
Otherwise it will slow down all other not weight related cases

There should be a weight provider dedicated interface/class instead this very generic

src/main/java/net/datafaker/service/RandomService.java
```diff
@@ -141,6 +151,101 @@ public Random getRandomInternal() {
...
+        if (!values.add(value)) {
+            throw new IllegalArgumentException("Duplicate value found: " + value + ". Values must be unique.");
+        }
+    }
+
+    private void validateItem(Map<String, Object> item) {
+        if (item == null || item.isEmpty()) {
+            throw new IllegalArgumentException("Item cannot be null or empty");
+        }
+        if (!item.containsKey(WEIGHT_KEY) || !item.containsKey(VALUE_KEY)) {
```
snuyanzin, line 199: same here about equals/hashcode

src/main/java/net/datafaker/service/RandomService.java
snuyanzin: > While weights are typically not expected to be extremely large,

if we are talking about project in open source which is used then any input should be expected if it passes all existing validation

SamuelNeceda: Sure, I got your point. I provided the guard for overflow.

src/main/java/net/datafaker/service/RandomService.java
```diff
@@ -141,6 +151,101 @@ public Random getRandomInternal() {
...
+
+        for (var item : items) {
+            validateItem(item);
+            assertUniqueValues(item, uniqueValues);
+        }
+    }
+
+    private void assertUniqueValues(Map<String, Object> item, Set<Object> values) {
+        Object value = item.get(VALUE_KEY);
+        if (!values.add(value)) {
```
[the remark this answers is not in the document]
SamuelNeceda: Do you have some suggestions on how to refactor it?

src/main/java/net/datafaker/service/RandomService.java
SamuelNeceda: Sorry, but I might not get this. I am checking if each item contains weight and value keys defined as strings. If for some reason they do not, I am directly throwing an exception.

src/main/java/net/datafaker/service/RandomService.java
SamuelNeceda: I've moved all the logic for the weight selector to the WeightedRandomSelector class. Currently, I’ve placed it in the service package, as it seems to fit best there, though I believe you initially suggested putting it in the providers package?

Additionally, based on the issue description, WeightedRandomSelector appears intended mainly for custom providers. At this point, no provider has data in the format required for this functionality. Should I register it only to be used with custom providers, or do you have a different solution in mind?

SamuelNeceda:
@kingthorin, could you please review the PR? I'd like to wrap it up or address any potential comments. Thanks!

kingthorin (approved):

src/main/java/net/datafaker/service/WeightedRandomSelector.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Set;
+
+/**
+ * A utility class for selecting a random element from a list based on assigned weights.
+ **/
+public class WeightedRandomSelector {
+    private static final String WEIGHT_KEY = "weight";
+    private static final String VALUE_KEY = "value";
+    private double[] cumulativeWeights;
+    private Object[] values;
```
snuyanzin, line 17: How about several (at least 2 threads) working with this
i guess they will concurrently change same class fields and inpact each other

snuyanzin (changes requested):
concurrency issue shouldn't be introduced in base classes

src/main/java/net/datafaker/service/WeightedRandomSelector.java
SamuelNeceda: @snuyanzin , with my last changes, I made the class stateless. Please, review.

src/main/java/net/datafaker/providers/base/ProviderRegistration.java
```diff
@@ -150,6 +151,8 @@ <PR extends ProviderRegistration, AP extends AbstractProvider<PR>> AP getProvide
 
     RandomService random();
 
+    WeightedRandomSelector weightedRandomSelector();
```
snuyanzin, line 154: what is the reason to have it here?
I would rather keep it completely separate

src/main/java/net/datafaker/service/WeightedRandomSelector.java
snuyanzin: i see, however it looks like now again refilling of arrays is happening on each iteration or did I miss anything?

src/main/java/net/datafaker/service/WeightedRandomSelector.java
```diff
@@ -0,0 +1,126 @@
...
+     * @return A randomly selected element based on its weight.
+     * @throws IllegalArgumentException if:
+     *                                  - the list is null or empty,
+     *                                  - any item in the list is null or empty,
+     *                                  - the item does not contain 'weight' or 'value' keys,
+     *                                  - any weight is null, non-positive, NaN or infinite,
+     *                                  - any values in the list are not unique or null,
+     *                                  - the sum of weights exceeds Double.MAX_VALUE.
+     */
+    public <T> T select(List<Map<String, Object>> items) {
```
snuyanzin, line 43: looks like most of the methods here do not work with object state => could be turned to `static`

src/test/weightedAnts.yml
```diff
@@ -0,0 +1,10 @@
+en:
+  faker:
+    insectsfromfile:
+      weightedAnts:
+        - value: "Driver ant"
+          weight: 6.0
+        - value: "Fire ant"
+          weight: 3.0
+        - value: "Harvester ant"
+          weight: 1.0
```
snuyanzin, line 10: With simple example it works
however what it we have more complex structure like 
https://github.com/datafaker-net/datafaker/blob/73ae1267e053dd9cb93039b1c721fdf0d282996b/src/main/resources/en/planet.yml#L14-L71

src/main/java/net/datafaker/service/WeightedRandomSelector.java
```diff
@@ -0,0 +1,126 @@
...
+    public WeightedRandomSelector(Random random) {
+        this.random = random != null ? random : new Random();
+    }
+
+    public Random getRandom() {
+        return random;
+    }
+
+    /**
+     * Returns a weighted random element from the given list, where each element is represented as a Map
```
snuyanzin, line 27: Am I right that the only way to change weight is change file and rebuild everything?

src/main/java/net/datafaker/service/WeightedRandomSelector.java
```diff
@@ -0,0 +1,126 @@
...
+    private void validateWeight(Object weightObj) {
+        if (!(weightObj instanceof Double weight)) {
+            throw new IllegalArgumentException("Weight must be a non-null Double");
+        }
+        if (weight <= 0 || Double.isNaN(weight) || Double.isInfinite(weight)) {
+            throw new IllegalArgumentException("Weight must be a positive number and cannot be NaN or infinite");
+        }
+    }
+
+    public double[] preprocessItems(List<Map<String, Object>> items, Object[] values) {
```
snuyanzin, line 99: I wonder why it is public?

src/main/java/net/datafaker/service/WeightedRandomSelector.java
```diff
@@ -0,0 +1,126 @@
...
+        return cumulativeWeights;
+    }
+
+    public <T> T selectWeightedElement(double randomValue, double[] cumulativeWeights, Object[] values) {
+        int index = Arrays.binarySearch(cumulativeWeights, randomValue);
+        index = (index < 0) ? -index - 1 : index;
+
+        if (index >= cumulativeWeights.length) {
+            index = cumulativeWeights.length - 1;
+        }
```
snuyanzin, line 122:
```suggestion
        index = Math.max(index, cumulativeWeights.length - 1);
```
i think it would be shorter

src/main/java/net/datafaker/service/WeightedRandomSelector.java
```diff
@@ -0,0 +1,126 @@
...
+        if (!values.add(value)) {
+            throw new IllegalArgumentException("Duplicate value found: " + value + ". Values must be unique.");
+        }
+    }
+
+    private void validateItem(Map<String, Object> item) {
+        if (item == null || item.isEmpty()) {
+            throw new IllegalArgumentException("Item cannot be null or empty");
+        }
+        if (!item.containsKey(WEIGHT_KEY) || !item.containsKey(VALUE_KEY)) {
```
snuyanzin, line 77: are we checking separately existing of keys and non null values only for the sake of different error messages>

SamuelNeceda:
Thank you for the feedback @snuyanzin
At this point, I’m not in a position to fully address the comments. Initially, I expected this issue to involve only minor changes, but it has become more complex than anticipated, and I’m unable to foresee all edge cases. Would it be possible for someone else to take it over, or should we consider creating a separate issue to track this?

snuyanzin:
with current architecture this topic is very complex that's why there are so many questions.
My proposal is to not touch base classes at all and have something as a proof of concept (POC) which could be accessible only on demand and not used by default

SamuelNeceda:
So the recommendation is to keep this as a POC only. What’s the next step from here? I’d like to ensure the work I've done isn't lost, even though I'm unable to complete the entire issue as intended. Any suggestions @kingthorin @snuyanzin?

kingthorin:
I don't mind "finishing" it. However I believe all that might be left to bring it in for POC purposes is to disconnect it from the faker or value service. Right @snuyanzin? 

Just let WeightedRandomSelector be instantiated and do all the work for this functionality while people test it and improve it.

snuyanzin:
yes, kind of
while all major issues are not solve we shouldn't introduce any change in classes like `BaseFaker.java`, `ProviderRegistration.java`, `FakeValuesService.java`

src/main/java/net/datafaker/service/WeightedRandomSelector.java
SamuelNeceda: Proposed change will always set index to `cumulativeWeights.length - 1` if index is less than `cumulativeWeights.length - 1`, which is not the intended behavior. The intent is to clamp the value only when index is greater than or equal to `cumulativeWeights.length`

src/main/java/net/datafaker/providers/base/ProviderRegistration.java
SamuelNeceda: as proposed, it was removed from `ProviderRegistration.java`

src/main/java/net/datafaker/service/WeightedRandomSelector.java
SamuelNeceda: Relevant methods were turned to static and private

src/test/weightedAnts.yml
SamuelNeceda: Since it was agreed that no changes should be introduced to `FakeValuesService.java`, I removed the `fetchWeightedList` method. Consequently, interacting with YAML files is no longer supported so I have also removed the example YAML file.

src/main/java/net/datafaker/service/WeightedRandomSelector.java
SamuelNeceda: The weights are provided as part of the input items list and are processed at runtime. Changing weights does not require rebuilding the application but simply involves modifying the input list or the file (if applicable) and rerunning the program.

src/main/java/net/datafaker/service/WeightedRandomSelector.java
SamuelNeceda: changed

src/main/java/net/datafaker/service/WeightedRandomSelector.java
SamuelNeceda: I split the condition into separate checks with distinct error messages.

SamuelNeceda:
@kingthorin @snuyanzin 
I have removed all changes from the mentioned classes, ensuring that the WeightedRandomSelector can now be instantiated independently.

Additionally, I have updated the documentation and examples to emphasize that weighted random selection is still in the Proof of Concept (POC) stage.

I believe these changes address your concerns, and the PR is now ready for merging. Any future suggestions for improvements will be considered and addressed in subsequent Issues, ...

docs/documentation/custom-providers.md
```diff
@@ -31,10 +31,21 @@ Create a custom provider of data:
...
         }
-    }
-    ```
 
+        public String weightedInsectName() {
+            List<Map<String, Object>> insects = List.of(
+                Map.of("value", "Driver ant", "weight", 6.0),
+                Map.of("value", "Fire ant", "weight", 3.0),
+                Map.of("value", "Harvester ant", "weight", 1.0)
+            );
```
snuyanzin, line 40: It would be better to follow same way as existing example and place all the example data in constant like above rather than generating it for each method invocation

docs/documentation/custom-providers.md
```diff
@@ -31,10 +31,21 @@ Create a custom provider of data:
...
+
+            WeightedRandomSelector selector = new WeightedRandomSelector(new Random());
```
snuyanzin, line 42: Also I guess we don't need to create a new selector per getter invocation and can reuse it

src/main/java/net/datafaker/service/WeightedRandomSelector.java
snuyanzin: how about this method
```java
<T> T select(List<Map<String, Object>> items) {
```

why can not it be switched to `static`?

src/main/java/net/datafaker/service/WeightedRandomSelector.java
```diff
@@ -0,0 +1,128 @@
...
+        if (valueObj == null) {
+            throw new IllegalArgumentException("Value cannot be null");
+        }
+    }
+
+    private static void validateWeight(Object weightObj) {
+        if (!(weightObj instanceof Double weight)) {
+            throw new IllegalArgumentException("Weight must be a non-null Double");
+        }
+        if (weight <= 0 || Double.isNaN(weight) || Double.isInfinite(weight)) {
```
snuyanzin, line 96: out of curiosity: why do we consider `weight = 0` as an error?
IMHO it might be a valid way to turn off this value while the total must be positive

src/test/java/net/datafaker/service/WeightedRandomSelectorTest.java
```diff
@@ -0,0 +1,326 @@
...
+    void testWeightedArrayElement_withInfiniteWeight(WeightedRandomSelector weightedRandomSelector) {
+        List<Map<String, Object>> items = List.of(
+            Map.of(VALUE_KEY, ELEMENT_1, WEIGHT_KEY, Double.POSITIVE_INFINITY),
+            Map.of(VALUE_KEY, ELEMENT_2, WEIGHT_KEY, ELEMENT_2_WEIGHT)
+        );
+
+        assertThatThrownBy(() -> weightedRandomSelector.select(items))
+            .isInstanceOf(IllegalArgumentException.class)
+            .hasMessage("Weight must be a positive number and cannot be NaN or infinite");
+    }
```
snuyanzin, line 168: It looks like all these tests (my be some more) could be replaced with one parameterized test
since we have input of key/value sequence, exception and error message

docs/documentation/custom-providers.md
SamuelNeceda: done

src/main/java/net/datafaker/service/WeightedRandomSelector.java
SamuelNeceda: My thought was that a weight of 0 implies that the corresponding item cannot be selected. Since WeightedRandomSelector is designed to assign probabilities proportional to weights, including a weight of 0 would lead to unnecessary processing for an element that cannot be selected. But I refactored the code to enable also 0 weight, allowing weights of 0 to be a legitimate way to "turn off" certain elements while still keeping them in the configuration

src/test/java/net/datafaker/service/WeightedRandomSelectorTest.java
SamuelNeceda: Good point, I refactored a whole test class to use a parameterized tests.

SamuelNeceda:
Thanks for the comments @snuyanzin , I have already addressed them.

src/test/java/net/datafaker/service/WeightedRandomSelectorTest.java
```diff
@@ -0,0 +1,273 @@
...
+    public static final double ELEMENT_2_WEIGHT = 2.0;
+    public static final double ELEMENT_3_WEIGHT = 3.0;
+    public static final double NEGATIVE_WEIGHT = -1.0;
+
+    /**
+     * Tests for exception scenarios using parameterized tests.
+     */
+    @ParameterizedTest
+    @MethodSource("exceptionCasesProvider")
+    void testWeightedArrayElement_exceptions(List<Map<String, Object>> items, String expectedMessage) {
```
snuyanzin, line 34: Can you rearrange a bit methods within this class
in a way: methods with different test annotations first
all other methods after that

snuyanzin, line 34:
```suggestion
    void testWeightedArrayElementExceptions(List<Map<String, Object>> items, String expectedMessage) {
```
lets's do not mix different notations

SamuelNeceda: refactored

src/main/java/net/datafaker/service/WeightedRandomSelector.java
```diff
@@ -0,0 +1,136 @@
...
+     * @return A randomly selected element based on its weight.
+     * @throws IllegalArgumentException if:
+     *                                  - the list is null or empty,
+     *                                  - any item in the list is null or empty,
+     *                                  - the item does not contain 'weight' or 'value' keys,
+     *                                  - any weight is null, non-positive, NaN or infinite,
+     *                                  - any values in the list are not unique or null,
+     *                                  - the sum of weights exceeds Double.MAX_VALUE.
+     */
+    public <T> T select(List<Map<String, Object>> items) {
```
snuyanzin, line 39:
```suggestion
    public static <T> T select(List<Map<String, Object>> items) {
```
this seems not addressed yet
not sure if we need it public

snuyanzin (approved):

kingthorin (approved):

PR #1418, opened 2024-10-31
Refactor few tests

asolntsev:


src/test/java/net/datafaker/providers/base/DateAndTimeTest.java
```diff
@@ -15,18 +15,32 @@
...
+import static java.time.temporal.ChronoUnit.MINUTES;
+import static java.time.temporal.ChronoUnit.NANOS;
+import static java.time.temporal.ChronoUnit.SECONDS;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 /**
  * @author pmiklos
  */
+@SuppressWarnings("removal")
```
kingthorin, line 38: If something here is deprecated for removal why not address the issue now instead of waiting for it to fail in the future?

asolntsev: The deprecated class is DataFaker's own class `DateAndTime.java`.
And this `DateAndTimeTest.java` is the test for this deprecated class.  :)
We still need this test until we finally delete both files.

kingthorin: Oh okay, I thought it was something in the Java base. Thanks for clarifying.

Issue #1412, opened 2024-10-29
NPE at net.datafaker.service.FakeValuesService.fetchObject():268


axelfontaine:
**Describe the bug**
```
java.lang.NullPointerException: Cannot invoke "java.util.Map.get(Object)" because the return value of "net.datafaker.internal.helper.CopyOnWriteMap.getOrDefault(Object, Object)" is null
      at net.datafaker.service.FakeValuesService.fetchObject(FakeValuesService.java:268)
      at net.datafaker.service.FakeValuesService.safeFetch(FakeValuesService.java:207)
      at net.datafaker.service.FakeValuesService.resolve(FakeValuesService.java:507)
      at net.datafaker.service.FakeValuesService.resolve(FakeValuesService.java:489)
      at net.datafaker.providers.base.AbstractProvider.resolve(AbstractProvider.java:20)
      at net.datafaker.providers.base.Animal.name(Animal.java:15)
```

Similar, but not identical to #1403 which was resolved in 2.4.1.

**To Reproduce**
Failure is intermittent and probably due to a concurrency issue when running multiple tests in parallel which all try to initialize datafaker at the same time

**Expected behavior**
No NPE

**Versions:**
 - OS: Win 11 x64
 - JDK: 22
 - Faker Version: 2.4.1

bodiam:
Hi @axelfontaine, 

Thanks for reporting. We are aware of this, it's maybe related to the CopyOnWrite map, or some odd initialisation or cleanup. I know @asolntsev gave it a good attempt, but so far we've been unable to get to the root cause of this. If you want, you might want to give it a try by replacing the COW map by a HashMap, but I believe @asolntsev mentioned the issue wasn't there.

asolntsev:
Hope it's fixed by https://github.com/datafaker-net/datafaker/pull/1419

bodiam:
@axelfontaine could you please verify using the snapshot version if it's fixed? If yes, we'll release another version.

axelfontaine:
@bodiam Do you have a snapshot repo I can reference in my build so I can pull it?

kingthorin:
https://www.datafaker.net/documentation/getting-started/#snapshot-versions

```
repositories {
    mavenCentral()
    maven(url = "https://s01.oss.sonatype.org/content/repositories/snapshots")
}

dependencies {
    implementation("net.datafaker:datafaker:2.4-2-SNAPSHOT")
}
```

axelfontaine:
[...]

PR #1419, opened 2024-10-31
#1412 fix concurrency issue in `CopyOnWriteMap`

asolntsev:
The default method `Map.getOrDefault` is not thread-safe: the condition `if (v == null && containsKey(key))` leaves `v = null` if the `key` was added by another thread exactly at the same moment. Then both conditions are true: `v==null` and `containsKey(key)`.

Now `CopyOnWriteMap.getOrDefault` does NOT check `containsKey(key)`. 
P.S. As a side-effect, `CopyOnWriteMap` doesn't allow putting `null` values to the map. But we don't need it anyway.

### DIFF
--- a/src/test/java/net/datafaker/Issue759Test.java
+++ b/src/test/java/net/datafaker/Issue759Test.java
@@ -1,13 +1,21 @@
 package net.datafaker;
 
 import org.junit.jupiter.api.RepeatedTest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.CountDownLatch;
 
+import static java.util.Collections.synchronizedList;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.assertj.core.api.Assertions.assertThat;
 
 class Issue759Test {
+    private static final List<Throwable> errors = synchronizedList(new ArrayList<>());
+    private static final Logger log = LoggerFactory.getLogger(Issue759Test.class);
+
     private static class WorkerThread extends Thread {
         private final Faker _faker;
         private final int _maxIterations;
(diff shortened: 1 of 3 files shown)


kingthorin (approved):
Seems reasonable to me.

src/test/java/net/datafaker/Issue759Test.java
```diff
@@ -29,13 +37,16 @@ public void run() {
...
-        String zipCode = faker.address().zipCodeByState(state);
         try {
+            String state = faker.address().stateAbbr();
+            String zipCode = faker.address().zipCodeByState(state);
             String county = faker.address().countyByZipCode(zipCode);
             assertThat(county).isNotEqualTo(zipCode);
-        } catch (RuntimeException expected) {
-            assertThat(expected).hasMessageStartingWith("County is not configured for postcode " + zipCode);
+        } catch (RuntimeException | Error e) {
+            if (!e.getMessage().startsWith("County is not configured for postcode ")) {
```
snuyanzin, line 46: Are we sure there couldn't be exception/error with null message?

asolntsev: Good point. Yes, theoretically it can happen.

asolntsev: @snuyanzin Fixed.

snuyanzin (approved):

asolntsev:
@kingthorin I think we should release DF 2.4.2
People sometimes complain about this concurrency issue.

bodiam:
@asolntsev I think we should perhaps first have the documentation for DF 2.4.1 before releasing 2.4.2?

bodiam:
@asolntsev I know, I know. I'll organize a training session soon ;)

But maybe I should make a quick document with steps involved, I know it's not the most exciting thing to work on, but I think it's kinda nice when the documentation is a little up to date.

kingthorin:
Ya if we can get that roll PR in first that'd be handy

bodiam:
Perhaps it would be nice to get this one in (https://github.com/datafaker-net/datafaker/pull/1414), or we keep this for 2.5?

kingthorin:
Oops sorry I missed your comment about 1414 earlier somehow.

I just came in here to say I'd done the 2.4.2 release.

bodiam:
Thanks for that. It doesn't look the like version in the pom.xml got updated though?

bodiam:
Oh, it creates a PR. Oki, that works!

asolntsev:
> Oops sorry I missed your comment about 1414 earlier somehow.
> 
> I just came in here to say I'd done the 2.4.2 release.

Sorry, but README is still not updated. :)

image

kingthorin:
Yup that hasn't been automated yet.

PR #1429, opened 2024-11-04
Update snapshot version in pom.xml

kingthorin:
@bodiam right?


SamuelNeceda (approved):

bodiam:
@kingthorin Nice. Bit odd, I thought the github action would do this, but maybe it needs another input parameter or so? (release version + next version) ?

kingthorin:
I'll have a look.

PR #1434, opened 2024-11-06
swedish ssn cannot have format like XXXXXX-000X

TrueJacobG:
swedish ssn cannot have format like XXXXXX-000X, the end number part has to be between 001 and 999

explanation: https://www4.skatteverket.se/rattsligvagledning/edition/2024.5/330242.html#update_20240802105842

Information obtained with help of @Jakub-Pazio and @Johannestegner

### DIFF
--- a/src/main/java/net/datafaker/idnumbers/SwedenIdNumber.java
+++ b/src/main/java/net/datafaker/idnumbers/SwedenIdNumber.java
@@ -45,13 +45,8 @@ public class SwedenIdNumber implements IdNumberGenerator {
         return new PersonIdNumber(idNumber, birthday, gender(f, request));
     }
 
-    private String generateEndPart(BaseProviders f) {
-        String end;
-        do {
-            end = f.numerify("###");
-        } while (end.equals("000"));
-
-        return end;
+    public static String generateEndPart(BaseProviders f) {
+        return "%03d".formatted(f.number().numberBetween(1, 1000));
     }
 
     @Deprecated
--- a/src/main/java/net/datafaker/idnumbers/SwedenIdNumber.java
+++ b/src/main/java/net/datafaker/idnumbers/SwedenIdNumber.java
@@ -37,12 +37,21 @@ public class SwedenIdNumber implements IdNumberGenerator {
     @Override
     public PersonIdNumber generateValid(BaseProviders f, IdNumber.IdNumberRequest request) {
         LocalDate birthday = birthday(f, request);
-        String end = f.numerify("###");
+        String end = generateEndPart(f);
         String basePart = DATE_TIME_FORMATTER.format(birthday)
             + f.options().option(PLUS_MINUS)
             + end;
         String idNumber = basePart + calculateChecksum(basePart);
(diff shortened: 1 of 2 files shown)


src/test/java/net/datafaker/idnumbers/SwedishIdNumberTest.java
```diff
@@ -14,6 +14,7 @@ void validSwedishSsn() {
 
     @Test
     void invalidSwedishSsn() {
+        assertThat(SwedenIdNumber.isValidSwedishSsn("020914-0001")).isFalse();
```
kingthorin, line 17: I don't get this, you said 001 to 999 are fine for last 3. So why should 001 be isvalid false?

Or did you mean it should be 3 digits (excluding triple zero) following the hyphen not 4?

TrueJacobG: I mean last 3 number without checksum digit so 020914-0001 is invalid.

kingthorin (approved):
Thanks

src/main/java/net/datafaker/idnumbers/SwedenIdNumber.java
```diff
@@ -37,14 +37,23 @@ public String getValidSsn(BaseProviders f) {
...
             + end;
         String idNumber = basePart + calculateChecksum(basePart);
         return new PersonIdNumber(idNumber, birthday, gender(f, request));
     }
 
+    private String generateEndPart(BaseProviders f) {
+        String end;
+        do {
+            end = f.numerify("###");
+        } while (!end.equals("000"));
```
asolntsev, line 52: Isn't it a bug?
Shouldn't it be the opposite: `while (end.equals("000"));`?

kingthorin: Ha, good catch. I think you're right

TrueJacobG: Yeah, you're right 😅. Already fixed

src/main/java/net/datafaker/idnumbers/SwedenIdNumber.java
```diff
@@ -78,6 +87,10 @@ public static boolean isValidSwedishSsn(String ssn) {
             return false;
         }
 
+        if(ssn.startsWith("000", 7)){
```
kingthorin, line 90:
```suggestion
        if (ssn.startsWith("000", 7)) {
```

kingthorin (approved):

asolntsev (changes requested):
@TrueJacobG There is a way to generate non-"000" number without a loop:

```java
    private String generateEndPart(BaseProviders f) {
        return "%03d".formatted(f.number().numberBetween(1, 1000));
    }
```

asolntsev (approved):

bodiam:
Thanks @asolntsev , @kingthorin and of course @TrueJacobG for working on this together, nice improvement!

PR #1438, opened 2024-11-11
Automate rolling of SNAPSHOT version in pom

kingthorin:
Follow-up to discussion in: https://github.com/datafaker-net/datafaker/issues/1412

### DIFF
--- a/.github/workflows/pom_update.yml
+++ b/.github/workflows/pom_update.yml
@@ -1,0 +1,29 @@
+on:
+  workflow_call:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  create_pr:
+    name: Create Pull Request
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+    - name: Set up Git
+      run: |
+        git config --global user.email "github-actions[bot]@users.noreply.github.com"
+        git config --global user.name "github-actions[bot]"
+        echo "BRANCH=pom-snapshot-updt" >> $GITHUB_ENV
+    - name: Build PR
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        git checkout -b $BRANCH
+        ./mvnw build-helper:parse-version versions:set -DnewVersion=${parsedVersion.majorVersion}.${parsedVersion.minorVersion}.${parsedVersion.nextIncrementalVersion}-SNAPSHOT
+        git add .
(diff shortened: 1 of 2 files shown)


.github/workflows/pom_update.yml
```diff
@@ -0,0 +1,29 @@
...
+      run: |
+        git config --global user.email "github-actions[bot]@users.noreply.github.com"
+        git config --global user.name "github-actions[bot]"
+        echo "BRANCH=pom-snapshot-updt" >> $GITHUB_ENV
+    - name: Build PR
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        git checkout -b $BRANCH
+        ./mvnw build-helper:parse-version versions:set -DnewVersion=\${parsedVersion.majorVersion}.\${parsedVersion.minorVersion}.\${parsedVersion.nextIncrementalVersion}-SNAPSHOT
```
kingthorin, line 25: Assuming x.y.z it'll always be z that's incremented.

We can always edit the PR or change it manually at any time.

If we want minor (y) incremented on most releases I can likely revise it for that as well.

bodiam: Any chance we can make this an input? I think we don't often make a Z change (I hope not, since they are bugfixes), but sometimes it's an Y, sometimes Z, so it would be nicer if I can just type in : next version is 2.5.0 (or 2.4.2), bit more explicit.

kingthorin: If you want to kick it off manually then yes you could have an input.

bodiam: Oh right, this is created as part of the releases? In that case, can't we use the release version here?

kingthorin: I could probably rewrite to third to do that but it wouldn't really make the situation any better. Having the existing version doesn't give any hint what element to increment.

I think the best plan is to just switch to incrementing the minor version (y) which as you described above should be correct in the majority of circumstances.

bodiam: Ah, we're talking about the next/snapshot version, my bad. In that case, I think it doesn't matter if we increment y or z right? (I mean, if 2.5.2-SNAPSHOT results in a 2.6.0 release, I don't think anyone would mind a huge deal)

kingthorin: Agreed, and we can always adjust it manually at any time between releases if necessary/wanted.

kingthorin:
@bodiam good to go for now?

PR #1442, opened 2024-11-18
feature: add movie name to movie provider

hardikSinghBehl:
This PR adds the below method to the existing movie provider:

- `new Faker().movie().name();`

Functionality has been validated and tested locally.


asolntsev (approved):

bodiam:
Nice one, thanks!

PR #1450, opened 2024-11-21
feat: add italian id number generator

cbxp:


asolntsev (approved):

src/main/java/net/datafaker/idnumbers/LatinLetters.java
```diff
@@ -0,0 +1,21 @@
...
+
+    private static final char[] VOWELS = {'A', 'E', 'I', 'O', 'U', 'Y'};
+
+    static boolean isConsonant(int c) {
+        return isConsonant((char) c);
+    }
+
+    static boolean isConsonant(char c) {
+        return Arrays.binarySearch(VOWELS, c) < 0;
+    }
```
kingthorin, line 15: Is this actually faster than using a `List` (`List.of`) and `.contains()`?

cbxp: @kingthorin Thanks, fixed. Now using `Set`.

asolntsev (approved):

src/main/resources/META-INF/services/net.datafaker.idnumbers.IdNumberGenerator
```diff
@@ -5,6 +5,7 @@ net.datafaker.idnumbers.ChineseIdNumber
 net.datafaker.idnumbers.EstonianIdNumber
 net.datafaker.idnumbers.FrenchIdNumber
 net.datafaker.idnumbers.GeorgianIdNumber
+net.datafaker.idnumbers.ItalianIdNumber
```
kingthorin, line 8: ?

kingthorin (approved):

PR #1452, opened 2024-11-24
Add argentina zipcodes

Jonathan-Zollinger:
# Overview
replaces the current 4-digit generator with an exhaustive list of argentine zip codes. I also noted that `es-AR` wasn't included in the list of locales in the README, so I added it there. 

# Points to consider
Argentine zip codes use preface 4 digits with an iso code, or a state abbreviation. This change includes that letter with the zip code such that each zip code provided is an actual zip code used in argentina. These can be found at https://postalcodes.azinfoportal.com/argentina/

# Test coverage
I added a new test which iterates through all 4025 new zip codes and validates that they're all formatted with an expected letter followed by 4 digits. I'm not sure how much more to test than that. 

calling `./mvnw clean test` from root runs w/ no errors. `./mvnw clean install` fails when it attempts to sign 4 files, which makes sense since I'm not expecting to sign any files. please correct me if that's wrong.


Jonathan-Zollinger:
closing this since i was missing some commits and this is getting sloppy. will resubmit a new PR

PR #1457, opened 2024-11-29
add new Canadian area code 942

asolntsev:
942 is a new area code to be introduced in 2025, see https://www.newswire.ca/news-releases/new-area-code-942-to-be-introduced-in-ontario-in-2025-817201718.html


kingthorin:
#Thanks!

PR #1458, opened 2024-12-03
swedish ssn symbol depends on birthday date of person

TrueJacobG:
If a person is older than 100 years, their SSN should have a "+" as the separator.

I found that when I was using datafaker and it sometimes creates SSNs that look like "000229+XXXX," which is an invalid SSN because February 29 did not exist in the year 1900.

more info: https://swedish.identityinfo.net/personalidentitynumber

### DIFF
--- a/src/main/java/net/datafaker/idnumbers/SwedenIdNumber.java
+++ b/src/main/java/net/datafaker/idnumbers/SwedenIdNumber.java
@@ -46,7 +46,7 @@ public class SwedenIdNumber implements IdNumberGenerator {
     }
 
     private String generateSymbol(LocalDate date){
-        return isYearOver100YearsAgo(date.toString().substring(0, 4)) ? "+" : "-";
+        return isYearOver100YearsAgo(date.toString().substring(0, 4), LocalDate.now()) ? "+" : "-";
     }
 
     public static String generateEndPart(BaseProviders f) {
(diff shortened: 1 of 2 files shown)


src/main/java/net/datafaker/idnumbers/SwedenIdNumber.java
```diff
@@ -39,12 +39,16 @@ public PersonIdNumber generateValid(BaseProviders f, IdNumber.IdNumberRequest re
         LocalDate birthday = birthday(f, request);
         String end = generateEndPart(f);
         String basePart = DATE_TIME_FORMATTER.format(birthday)
-            + f.options().option(PLUS_MINUS)
+            + generateSymbol(birthday)
```
asolntsev, line 42: Is the initial problem really solved here?
Now you generate the right symbol (`-` or `+`), but the date still may be `**0229+*****`.

asolntsev (changes requested):

src/main/java/net/datafaker/idnumbers/SwedenIdNumber.java
TrueJacobG: It's nothing incorrect with **0229+****, because 29th of February happend in few years. The incorrect value was 000229+****, because there was no 29th of February in 1900.

src/main/java/net/datafaker/idnumbers/SwedenIdNumber.java
```diff
@@ -91,19 +95,48 @@ public static boolean isValidSwedishSsn(String ssn) {
...
+        }
+
+        if (!ChronoField.DAY_OF_MONTH.range().isValidIntValue(Integer.parseInt(dateString.substring(4)))) {
+            return true;
+        }
+
+        char symbol = ssn.charAt(6);
+
+        if (symbol == '+') {
+            dateString = "19" + dateString;
```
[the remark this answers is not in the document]
TrueJacobG: Yeah, you're right that's way better!

but, in fact this is a problem of swedish SSN itself. The format is not ready for 22 century so it will eventually break in the future (or rather ssn format will got change in the future).

src/main/java/net/datafaker/idnumbers/SwedenIdNumber.java
asolntsev: 1. Method `isYearOver100YearsAgo` expects year number in the first argument. But it's named `date` (not `year`), and you pass `dateString` in line 120 (`isYearOver100YearsAgo(dateString, ...`).

2. 
This code:
```java
    @RepeatedTest(1000)
    void swedishSsn_containsPlus_forPersonsOlderThan100Years() {
        PersonIdNumber person = new SwedenIdNumber().generateValid(new Faker(), new IdNumber.IdNumberRequest(123, 125, ANY));
        assertThat(person.idNumber()).contains("+");
        assertThat(SwedenIdNumber.isValidSwedishSsn(person.idNumber())).as(person.idNumber()).isTrue();
    }
```

still can generate invalid IDs, e.g. `991207+4219`, `991224+0851` or `991207+5430`.

TrueJacobG: Fixed it!

also 991207+4219, 991224+0851 and 991207+5430 they are valid SSNs. For ssn validation logic I recommend using https://swedish.identityinfo.net/personalidentitynumber/validate?number=991207%2B4219

asolntsev (approved):

asolntsev:
@TrueJacobG Thank you, merged!

PR #1460, opened 2024-12-06
simplify few tests

asolntsev:
I see that `@ParametrizedTest` may be over-used. It causes "IFs" and logic in tests trying to handle all given values.

Sometimes it's better to just write plain old test without any IFs.


kingthorin (approved):

kingthorin:
Thanks!

PR #1465, opened 2024-12-13
Respect arg types while looking for method for expressions

### LINKED ISSUE #1463: `number_between` for integers returns doubles from time to time
**Describe the bug**
This is a regression issue. It's not reproduced on `2.4.11`, but reproduced from time to time on `2.4.2`. Most likely it is caused by this commit: https://github.com/datafaker-net/datafaker/commit/f073184da3b743e4cc0faa05d540196baa99ac89.

**To Reproduce**
```java
import net.datafaker.Faker;

public class Main
{
    {
        Faker faker = new Faker();
        String result = faker.expression("#{number.number_between '1000','1000'}");
        System.out.printf(result);
    }
}
```

**Expected behavior**
The result should be '1000', but it is '1000.0' from time to time.

**Versions:**
 - OS: any
 - JDK: no correlation detected
 - Faker Version: `2.4.2`

**Additional context**
The methods order is preserved when they are returned from class: https://github.com/datafaker-net/datafaker/blob/abdecbc048b2c26c766d91edbf32ff954073f2d4/src/main/java/net/datafaker/service/FakeValuesService.java#L964
But then they are stored in HashMap with unpredictable order: https://github.com/datafaker-net/datafaker/blob/abdecbc048b2c26c766d91edbf32ff954073f2d4/src/main/java/net/datafaker/service/FakeValuesService.java#L966
I see 2 possible solutions:
- rely on method order in the class and use ordered collections to keep sequence - fragile
- improve the algorithm to distinguish double-s vs long-s vs int-s, and search for the method in priority starting from best matching type: e.g. 10 -> int, 10000000000 -> long, 10.0 -> double

snuyanzin:
This will fix #1463

### DIFF
--- a/src/test/java/net/datafaker/FakerTest.java
+++ b/src/test/java/net/datafaker/FakerTest.java
@@ -162,7 +162,9 @@ class FakerTest extends AbstractFakerTest {
         assertThat(faker.expression("#{letterify '????','true'}")).matches("[A-Z]{4}");
         assertThat(faker.expression("#{templatify '????','?','1','2','q','r'}")).matches("([12qr]){4}");
         assertThat(faker.expression("#{Name.first_name} #{Name.first_name} #{Name.last_name}")).matches("[a-zA-Z']+ [a-zA-Z']+ [a-zA-Z']+");
-        assertThat(faker.expression("#{number.number_between '1','10'}")).matches("[1-9](\\.[0-9]+)?");
+        assertThat(faker.expression("#{number.number_between '1','10'}")).matches("[1-9]([0-9]+)?");
+        assertThat(faker.expression("#{number.number_between '1','10.0'}")).matches("[1-9](\\.[0-9]+)?");
+        assertThat(faker.expression("#{number.number_between '1.0','10'}")).matches("[1-9](\\.[0-9]+)?");
         assertThat(faker.expression("#{color.name}")).matches("[a-z\\s]+");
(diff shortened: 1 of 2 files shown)


src/test/java/net/datafaker/FakerTest.java
```diff
@@ -162,7 +162,9 @@ void expression() {
         assertThat(faker.expression("#{letterify '????','true'}")).matches("[A-Z]{4}");
         assertThat(faker.expression("#{templatify '????','?','1','2','q','r'}")).matches("([12qr]){4}");
         assertThat(faker.expression("#{Name.first_name} #{Name.first_name} #{Name.last_name}")).matches("[a-zA-Z']+ [a-zA-Z']+ [a-zA-Z']+");
-        assertThat(faker.expression("#{number.number_between '1','10'}")).matches("[1-9](\\.[0-9]+)?");
+        assertThat(faker.expression("#{number.number_between '1','10'}")).matches("[1-9]([0-9]+)?");
```
asolntsev, line 165: I think this regex can be simplified: `[1-9]\\d*`  @snuyanzin

asolntsev (approved):

src/test/java/net/datafaker/FakerTest.java
snuyanzin: thanks, done

valfirst (approved):

Issue #1176, opened 2024-04-26
Provide Stream output for transformers


snuyanzin:
The problem with current method like `net.datafaker.transformations.JsonTransformer#generate(net.datafaker.transformations.Schema<IN,?>, int)`
it generates the whole `String` and then returns it. As a result for bigger numbers it consumes larger amount of memory and e.g.
such test fails with OutOfMemory
```java
@Test
    void test2() {
        BaseFaker faker = new BaseFaker(new Random(10L));
        Schema<Object, ?> schema = Schema.of(
            field("Text", () -> faker.name().firstName()),
            field("Bool", () -> faker.name().lastName())
        );

        JsonTransformer<Object> transformer = JsonTransformer.builder().build();
        String json = transformer.generate(schema, 50_000_000);
        System.out.println(json);
    }
```

There is not so much we can do about this method since anyway with such approach we need somehow to store that giant string value.

Another approach is instead of generation the final string value we could generate a stream of values and return it.

snuyanzin:
partially covered with https://github.com/datafaker-net/datafaker/pull/1177
however not all formats are supported yet

RVRhub:
Do I understand correctly that this proposal suggests implementing a solution similar to the one used for JsonTransformer, but also for:
 - XmlTransformer
 - JavaObjectTransformer
 - SqlTransformer

snuyanzin:
yes you are right

gatear:
My PR covers the SQL transformer support
https://github.com/datafaker-net/datafaker/pull/1264

gatear:
Also a PR for JavaObjectTransformer https://github.com/datafaker-net/datafaker/pull/1313

PR #1468, opened 2024-12-15
Add write to output stream

snuyanzin:
This is the next step after https://github.com/datafaker-net/datafaker/issues/1176 (some formats are already supporting streaming).

Now it would make sense to add ability to write data to output streams (in tests there is an example with files).

Currently supported for json, sql, csv


src/main/java/net/datafaker/transformations/CsvTransformer.java
```diff
@@ -111,13 +113,13 @@ public String generate(Schema<IN, ?> schema, int limit) {
     @Override
     public String getStartStream(Schema<IN, ?> schema) {
         StringBuilder sb = new StringBuilder();
-        generateHeader(null, sb);
+        generateHeader(schema, sb, false);
```
snuyanzin, line 116: suddenly it was an uncovered bug leading to NPE 🙈

src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
```diff
@@ -138,6 +139,11 @@ public Stream<Object> generateStream(final Schema<Object, ?> schema, long limit)
                 .limit(limit);
     }
 
+    @Override
+    public void writeToOutputStream(OutputStream outputStream, Schema<Object, ?> schema, long limit) {
+        throw new UnsupportedOperationException("Not implemented yet");
```
snuyanzin, line 144: not sure how it should look like in this case

kingthorin: I think this is fine for now.

Maybe add some JavaDoc as well.

snuyanzin: added a comment for that

src/main/java/net/datafaker/transformations/Transformer.java
```diff
@@ -29,18 +32,24 @@ default String getElementSeparator() {
...
-
     default Stream<OUT> generateStream(final Schema<IN, ?> schema, long limit) {
         Item item = new Item(0);
         return Stream.generate(() -> {
             StringBuilder res = new StringBuilder();
             if (item.current == 0) {
-                res.append(getStartStream(schema)).append(getLineSeparator());
+                final String startStream = getStartStream(schema);
+                if (startStream != null) {
+                    res.append(getStartStream(schema)).append(getLineSeparator());
```
kingthorin, line 42: Why bother with the intermediate store? Why not just call the method in the condition? (Esp since it's called again to append anyway)

kingthorin: Or do it like the end one and use it for the append 😉

src/main/java/net/datafaker/transformations/XmlTransformer.java
```diff
@@ -55,12 +55,12 @@ public CharSequence generate(Schema<IN, ?> schema, int limit) {
...
     @Override
     public String getStartStream(Schema<IN, ?> schema) {
-        throw new UnsupportedOperationException();
+        return null;
     }
 
     @Override
     public String getEndStream() {
-        throw new UnsupportedOperationException();
+        return null;
```
kingthorin, line 63: Wouldn't it be better to return empty string?

src/main/java/net/datafaker/transformations/Transformer.java
snuyanzin: seems forgot about this, thanks

src/main/java/net/datafaker/transformations/XmlTransformer.java
snuyanzin: that's where I'm still not sure, empty string also might be ok here

kingthorin: In both cases

L58 too

kingthorin: Sorry for some reason I was only seeing one changed. (Maybe I looked at a single commit vs the whole change set 🤷‍♂️)

kingthorin (approved):

src/main/java/net/datafaker/transformations/XmlTransformer.java
snuyanzin: no problem, thanks for the feedback

PR #1473, opened 2024-12-30
Fix 'since' version for Dota2 provider

valfirst:
Commit for reference: https://github.com/datafaker-net/datafaker/commit/26e75e390d1d35dcad82822e1aa61b19c0122a93


kingthorin:
Doesn't this actually suggest 2.4.2? (Since it seems to show between 2.4.1 and 2.4.2)
image

valfirst:
@kingthorin 'since' define the first version the provider is available from
Screenshot 2024-12-31 at 12 57 23

kingthorin:
Thanks!

kingthorin (approved):

PR #1489, opened 2025-01-25
Refactor country-specific properties

asolntsev:
Move country-specific properties from `lang.yml` to `_COUNTRY.yml` files.

These are settings that depend on country, not language (e.g. postal code, building numbers, passport numbers, internet domains).

The primary test showing the difference is `AddressTest.usingOnlyCountryCodeWithoutLanguage()`. 

For example,

```java
assertThat(new Faker(new Locale("et", "AL")).address().countryCode()).isEqualTo("AL");
  // was: EE
  // now: AL
```

And:
```java
assertThat(new Faker(new Locale("et", "BY")).address().postcode()).matches("\\d{6}");
  // was: <5 digits> (Estonian postal code)
  // now: <6 digits> (Belorussian postal code)
```

In this PR, I had to introduce concept of "language's default country". 
In the case when only the language is given, we need to detect the country:
```java
assertThat(new Faker(new Locale("et")).address().countryCode()).isEqualTo("EE"); // for Estonian language, we use Estonia settings by default
assertThat(new Faker(new Locale("hy")).address().countryCode()).isEqualTo("AM"); // for Armenian language, we use Armenia settings by default
```

The list of "language's default countries" is located in `FakerContext.LANGUAGE_DEFAULT_COUNTRY`.


src/test/java/net/datafaker/service/FakerContextTest.java
```diff
@@ -0,0 +1,86 @@
...
+    @ParameterizedTest
+    @MethodSource("languagesMatchingCountryCode")
+    void appendsDefaultCountryForGivenLanguage(String language) {
+        String countryCode = language.toUpperCase(Locale.ROOT);
+        assertThat(chain(new Locale(language))).containsExactly(new Locale(language), new Locale("", countryCode), ENGLISH);
+    }
+
+    @ParameterizedTest
+    @MethodSource("languagesDifferentFromCountryCode")
+    void appendsDefaultCountryForGivenLanguage2(String language, String country) {
```
snuyanzin, line 37: nit: can we rename it to have better match with what we have in `MethodSource` since suffix `2` doesn't really bring lots of benefit

README.md
```diff
@@ -475,20 +475,20 @@ Supported Locales
 * ar (Armenian)
```
snuyanzin, line 475: how about making this section collapsible? 
Not sure that everyone what to scroll all the locales mentioned here to see what else is in readme

snuyanzin (approved):

PR #1490, opened 2025-01-27
Added support for generating Latvian id number

cbxp:
* We support two formats: old (before 01.07.2017) and new (since 01.07.2017)
* For generating invalid ID numbers, we use only the old format (for simplicity)

### DIFF
--- a/src/main/java/net/datafaker/idnumbers/LatvianIdNumber.java
+++ b/src/main/java/net/datafaker/idnumbers/LatvianIdNumber.java
@@ -58,7 +58,7 @@ public class LatvianIdNumber implements IdNumberGenerator {
             case 18 -> 0;
             case 19 -> 1;
             case 20 -> 2;
-            default -> throw new IllegalStateException("Too far in future: " + birthYear);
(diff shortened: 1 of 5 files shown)


src/main/java/net/datafaker/idnumbers/LatvianIdNumber.java
```diff
@@ -0,0 +1,77 @@
...
+            centuryDigit(birthday.getYear()) +
+            faker.number().digits(3);
+    }
+
+    static int centuryDigit(int birthYear) {
+        return switch (birthYear / 100) {
+            case 18 -> 0;
+            case 19 -> 1;
+            case 20 -> 2;
+            default -> throw new IllegalStateException("Too far in future: " + birthYear);
```
snuyanzin, line 61: nit: i guess in case of year 1700 it will fail with `Too far in future: 1700` which doesn't sound like true

cbxp: Thanks, error message improved.

asolntsev (approved):

snuyanzin (approved):

PR #1492, opened 2025-01-28
fix flaky CustomFakerTest

asolntsev:


kingthorin (commented):
Why do we want 3million operations vs 100?

asolntsev:
> ~Why~ Do we really want 3million operations vs 100?

:)

No-no, I added `@RepeatedTest(10_000)` temporary - just to be sure that my change stably fixes the problem.
The actual fix is to change "100" to "300".

kingthorin (commented):
Thanks

PR #1504, opened 2025-02-26
Added Uzbek locale support with uz.yml and _UZ.yml

jahong1r-t:
I’ve added the uz.yml and _UZ.yml files with localized data for Uzbek (names, addresses, phone numbers, etc.) and updated README.md. Tests confirmed that the data loads correctly (e.g., names like "Ali", addresses like "Toshkent"). Let me know if you need any further adjustments!

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,6 +539,7 @@ In the example above,
 * tr (Turkish)
 * th_TH (language: Thai, country: Thailand)
 * uk_UA (language: Ukrainian, country: Ukraine)
+* uz (language: Uzbek, country: Uzbekistan)
 * vi_VN (language: Vietnamese, country: Vietnam)
 * zh_CN (language: Chinese, country: China)
 * zh-TW (language: Chinese, country: Taiwan)
--- a/src/main/java/net/datafaker/providers/base/Locality.java
+++ b/src/main/java/net/datafaker/providers/base/Locality.java
@@ -22,7 +22,7 @@ public class Locality extends AbstractProvider<BaseProviders> {
         "_jp",
         "_il", "_in",
         "_kr",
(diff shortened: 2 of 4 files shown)


asolntsev (approved):

README.md
```diff
@@ -539,6 +539,7 @@ In the example above,
 * tr (Turkish)
 * th_TH (language: Thai, country: Thailand)
 * uk_UA (language: Ukrainian, country: Ukraine)
+* uz (language: Uzbek, country: Uzbekistan)
```
snuyanzin, line 542: This line seems to be incorrect
if you look at others then letters before underscore mean language and after country name.
In your case there is only a part before underscore, then only language should be in parenthesis (similar to `tr (Turkish)` above)

jahong1r-t: Ok, I'll fix that.

jahong1r-t: I updated the readme file. I changed uz (language: Uzbek, country: Uzbekistan) to uz (Uzbek)

README.md
```diff
@@ -95,53 +98,80 @@ $5 ==> "Vernie Schmidt"
...
+expression("#{regexify '[a-z]{4,10}'}"); // wbevoa
+faker.
+
+expression("#{options.option '23','2','5','$','%','*'}"); // *
+faker.
+
+expression("#{date.birthday 'yy DDD hh:mm:ss'}"); // 61 327 08:11:45
+faker.
+
+expression("#{csv '1','name_column','#{Name.first_name}','last_name_column','#{Name.last_name}'}");
```
snuyanzin, line 124: Hmm...
I'm pretty sure all these and most of other changes have nothing to do with the actual PR description. Please stick only to those changes which are solving the issue mentioned in PR description. If some other things should be changed we can discussed it in separate PRs

snuyanzin (approved):

PR #1510, opened 2025-03-02
Added ARM support to the build to see if that has performance benefits.

bodiam:
Inspired by [this post on LinkedIn](https://www.linkedin.com/posts/laurentbroudoux_arm-activity-7295081454189891586-t6yy?utm_source=social_share_send&utm_medium=member_desktop_web&rcm=ACoAAAC299sBAspEXK7yi4HueOvVFeTyVEHeFts,) I tried moving to ARM as well (See [ARM support for Github runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)) to see if this speeds things up. It seems it doesn't?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       matrix:
         java-version: [ 17, 21, 23 ]
-        runs-on: [ ubuntu-latest, macos-latest, windows-latest ]
+        runs-on: [ ubuntu-latest, ubuntu-24.04-arm, macos-latest, windows-latest ]
     name: Jdk ${{ matrix.java-version }}, os ${{ matrix.runs-on }}
     runs-on: ${{ matrix.runs-on }}
     steps:


bodiam:
Well, that's odd:

```
 Jdk 17, os ubuntu-24.04-arm
succeeded Mar 2, 2025 in 1m 49s 
```

https://github.com/datafaker-net/datafaker/actions/runs/13614876518/job/38056657086?pr=1510

```
Jdk 17, os ubuntu-latest
succeeded Mar 2, 2025 in 1m 49s 
```

https://github.com/datafaker-net/datafaker/actions/runs/13614876518/job/38056657271?pr=1510#logs

Am I missing something?

snuyanzin:
I guess in your case you need to pay attention only to `Build with Maven` step
in case of `ubuntu-latest` - 1m 10s
in case of `ubuntu-24.04-arm` - 1m 28s

it is worth to note that in case of `ubuntu-latest` we run tests for one dummy downstream project to be able to catch issues with jars which is impossible to catch within datafaker ci. There is no need to run it for every os, so only one was selected

bodiam:
The ARM wasn't supposed to be slower. Perhaps the performance gain was more applicable to them than to us. I'll run it a few more times, see if it happens, or merge it in, so we can see if it does anything.

PR #1520, opened 2025-03-24
Use `allowZeroInvocations` junit feature instead of a WA

snuyanzin:
At some point of time there was introduced a work around for the case when parameterized test input is empty.
Since junit 5.12.x there is a special annotation flag for that, so just replacing old workaround with this flag usage.

Edit: Original intro was here: https://github.com/datafaker-net/datafaker/pull/677


kingthorin (approved):
Thanks!

PR #1530, opened 2025-04-06
[refactor] extract picking a random enum value to method `RandomService.nextEnum()`

asolntsev:


89snake89:
Hi Everyone
When 2.4.3 release planned ?

src/main/java/net/datafaker/service/RandomService.java
```diff
@@ -116,6 +116,11 @@ public byte[] nextRandomBytes(int numberOfBytes) {
         return randomBytes;
     }
 
+    public <T extends Enum<T>> T nextEnum(Class<T> klass) {
```
bodiam, line 119: Since this is now a public method, would it be possible to add a unit test for this?

src/main/java/net/datafaker/service/RandomService.java
```diff
@@ -116,6 +116,11 @@ public byte[] nextRandomBytes(int numberOfBytes) {
...
+        T[] enumConstants = klass.getEnumConstants();
```
snuyanzin, line 120: there is already existing method with similar code https://github.com/datafaker-net/datafaker/blob/646a3f91892c8d7d1bc073a6d105c5c51358c429/src/main/java/net/datafaker/providers/base/Options.java#L146-L149

probably it doesn't have obvious name since it is not the first time there is a PR with another method like that..
How about making one of them reusing another?

src/main/java/net/datafaker/service/RandomService.java
asolntsev: @bodiam Thanks, added the tests.

src/test/java/net/datafaker/service/RandomServiceTest.java
```diff
@@ -145,6 +153,36 @@ void testDefaultHex(RandomService randomService) {
...
+    })
+    void predictableEnum(int seed, Ring expected1, Ring expected2, Ring expected3, Ring expected4, Ring expected5) {
+        RandomService randomService = new RandomService(new Random(seed));
+        assertThat(List.of(
+            randomService.nextEnum(Ring.class),
+            randomService.nextEnum(Ring.class),
+            randomService.nextEnum(Ring.class),
+            randomService.nextEnum(Ring.class),
+            randomService.nextEnum(Ring.class)
+        )).isEqualTo(List.of(expected1, expected2, expected3, expected4, expected5));
```
snuyanzin, line 181:
```suggestion
        )).containsExactly(expected1, expected2, expected3, expected4, expected5);
```

src/test/java/net/datafaker/service/RandomServiceTest.java
```diff
@@ -145,6 +153,36 @@ void testDefaultHex(RandomService randomService) {
         assertThat(randomService.hex()).matches("^[0-9A-F]{8}$");
     }
 
+    @Test
+    void nextEnum() {
+        RandomService randomService = new RandomService();
+        Set<Ring> all = new HashSet<>();
```
snuyanzin, line 159:
```suggestion
        Set<Ring> all = new EnumSet<>();
```

src/test/java/net/datafaker/service/RandomServiceTest.java
asolntsev: thanks, done.

src/test/java/net/datafaker/service/RandomServiceTest.java
asolntsev: thanks, done.

PR #1533, opened 2025-04-07
Cleanup release process.

bodiam:
There's still an error in the create-new-PR part, somewhere here:

```
        git commit -m "Update pom SNAPSHOT version" -m "A release just completed, this PR increments the SNAPSHOT version in pom.xml" --signoff
        git push --set-upstream origin $BRANCH --force
        gh pr create --fill
```

See https://github.com/datafaker-net/datafaker/actions/runs/14307537882/job/40094866379

But it seems the 2.4.3 release has been created. Also, I thought this might be nice to share:

image

1M is within reach!!!


snuyanzin (approved):

kingthorin:
If we're comfortable that it works properly it could just push directly instead of raising a PR.

I think a PR is safer but 🤷‍♂️

I guess the readme also needs its snapshot reference updated each time

bodiam:
@kingthorin I'm not sure. I think there's something wrong with the `--fill` part, maybe putting an explicit `--title` and `--body` would be more safe? I couldn't figure out the root cause why the fill didn't work. I like sending the PR though, though pushing directly is fine as well with me.

kingthorin:
Need to be careful if deciding to change on the fly, things like @since annotations or comments could end up being wrong

PR #1545, opened 2025-04-13
Descrease amount of calls to nextInt for numerify

snuyanzin:
Since `nextInt` is one of the top consumers while executing `numerify` the idea is to reduce amount of calls to `nextInt`. 
For instance if we need to execute `numerify` for `####` then 4 digit number should be generated.
Current approach is generation a digit for every `#`. At the same time if `#` are together we can first count them and generate one number and then pick digits for every `#`. Then only one call to `nextInt` will be required. 
For sure if amount of sharps is large then several calls will be required (because of limited amount of numbers for int)
I also tried `nextLong` however it works significantly slower than `nextInt`.

some measurements for 6 `#` like
```java
@Benchmark
public void numerify(Blackhole blackhole) {
    blackhole.consume(DATA_FAKER.numerify("######"));
}
```
before PR
```
Benchmark                           Mode  Cnt      Score     Error   Units
DatafakerTemplateStrings.numerify  thrpt   10  39593.364 ± 237.243  ops/ms
```
after PR
```
Benchmark                           Mode  Cnt      Score     Error   Units
DatafakerTemplateStrings.numerify  thrpt   10  48787.181 ± 255.048  ops/ms
```

For cases with less than 5 `#` it is almost same behavior (no significant diff)

As a result it should also speed up a number of providers for id and address generation where such `numerify` calls are used, probably something else. Perhaps in similar way `letterify` might be improved... however not in this PR

### DIFF
--- a/src/main/java/net/datafaker/service/FakeValuesService.java
+++ b/src/main/java/net/datafaker/service/FakeValuesService.java
@@ -373,22 +373,27 @@ public class FakeValuesService {
 
     private String bothify(String input, FakerContext context, boolean isUpper, boolean numerify, boolean letterify) {
         final int baseChar = isUpper ? 'A' : 'a';
+        boolean changed = false;
         final char[] res = input.toCharArray();
+        RandomService randomService = context.getRandomService();
         for (int i = 0; i < res.length; i++) {
             switch (res[i]) {
                 case '#' -> {
                     if (numerify) {
-                        res[i] = DIGITS[context.getRandomService().nextInt(10)];
+                        changed = true;
+                        i += GenerationUtils.generateAndSetNumber(i, res, res[i], randomService) - 1;
                     }
                 }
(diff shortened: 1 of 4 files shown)


kingthorin (approved):

asolntsev:
@snuyanzin @kingthorin If you ask me, I would say this PR brings quite a complex code, and doesn't give remarkable benefits. The improvement in speed is extremely insignificant. Sorry :)

kingthorin:
It's already been merged. I'm not rolling it back at this point. You guys can great debate it and let us know.

asolntsev:
@kingthorin No-no, I don't want battles. All right. Just shared my impression. :)

kingthorin:
😁 battle was a ridiculous word to use, I dunno why I said it that way.

More like debate....

PR #1549, opened 2025-05-07
feat: add MartialArt provider with names

paulofranklins2:
- Created MartialArt provider in `net.datafaker.providers.sport`
- Added `martialart()` accessor in SportProviders interface
- Registered YAML resource file `martial_art.yml` with martial art names
- Updated EnFile to include "martial_art.yml"
- Implemented unit tests in `MartialArtTest` for name


kingthorin:
To address the build failure:
`01:54:10.734 [ERROR] Run 'mvn spotless:apply' to fix these violations.`

asolntsev (approved):

src/main/java/net/datafaker/providers/sport/MartialArt.java
```diff
@@ -0,0 +1,27 @@
...
+import net.datafaker.providers.base.AbstractProvider;
+
+/**
+ * Custom Faker provider for martial arts.
+ * Supports resolving values from "martial_art.name" and "martial_art.origin" in the locale YAML file.
+ *
+ * <p>Contributed by Paulo Cunha</p>
+ *
+ * @see <a href="https://github.com/paulofranklins2/">https://github.com/paulofranklins2/</a>
+ * @see <a href="https://www.linkedin.com/in/paulofranklins2/">https://www.linkedin.com/in/paulofranklins2/</a>
```
bodiam, line 12: You have a typo here, your LinkedIn is https://www.linkedin.com/in/paulofranklins/, without the `2` :)

src/main/resources/en/martial_art.yml
```diff
@@ -0,0 +1,109 @@
...
+        - "Istunka"
+        - "Mixed Martial Arts"
+        - "Krav Maga"
+        - "Systema"
+        - "Sambo"
+        - "Defendo"
+        - "Kajukenbo"
+        - "Shootfighting"
+        - "LINE"
+      origin:
```
bodiam, line 79: I'm not sure if we really need an origin, it's hardly different than the `Country.name()` right? For example, there's also `Dutch kickboxing`, which could work well with Country.name(). 

Note: I don't have too many concerns about having this list here, I'm just thinking out loud a little.

paulofranklins2: removed.

src/main/java/net/datafaker/providers/sport/MartialArt.java
paulofranklins2: thank you good catch.

kingthorin:
Still needs the aforementioned spotless apply

kingthorin (approved):

PR #1555, opened 2025-05-16
add Hungarian id number generator

cbxp:


src/main/java/net/datafaker/idnumbers/HungarianIdNumber.java
```diff
@@ -0,0 +1,81 @@
...
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+import static java.lang.Character.getNumericValue;
+import static net.datafaker.idnumbers.Utils.randomGender;
+
+/**
+ * Hungarian personal identification number
+ * <p>
+ * The structure of such number is GYYMMDDXXC, whereas G indicates century of birth and gender), YYMMDD is the
```
kingthorin, line 16: unmatched bracket after gender

src/test/java/net/datafaker/helpers/IdNumberPatterns.java
```diff
@@ -10,5 +10,6 @@ public class IdNumberPatterns {
     public static final Pattern FRENCH = Pattern.compile("[12]\\d{2}(1[0-2]|0[1-9])\\d[0-9a-zA-Z]\\d{8}");
     public static final Pattern ITALIAN = Pattern.compile("[A-Z]{6}\\d{2}[ABCDEHLMPRST]\\d{2}[\\dA-Z]{5}");
     public static final Pattern POLISH = Pattern.compile("\\d{11}");
+    public static final Pattern HUNGARIAN = Pattern.compile("\\d{10}");
```
kingthorin, line 13: Should we specify that the first digit should be in the set 1-4?

src/test/java/net/datafaker/idnumbers/HungarianIdNumberTest.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Test
+    void checkDigit() {
+        assertThat(HungarianIdNumber.getCheckDigit("807159215")).isEqualTo(3);
+        assertThat(HungarianIdNumber.getCheckDigit("100000000")).isEqualTo(1);
+        assertThat(HungarianIdNumber.getCheckDigit("010000000")).isEqualTo(2);
+        assertThat(HungarianIdNumber.getCheckDigit("001000000")).isEqualTo(3);
+        assertThat(HungarianIdNumber.getCheckDigit("000000001")).isEqualTo(9);
+        assertThat(HungarianIdNumber.getCheckDigit("170010103")).isEqualTo(10);
```
kingthorin, line 32: If it's 10 wouldn't that make the whole value 11 char?

kingthorin (commented):
Sorry I probably should have noticed these two items earlier.

src/test/java/net/datafaker/idnumbers/HungarianIdNumberTest.java
asolntsev: Yes, that's why we need this loop in `HungarianIdNumber.generateValid`:

```java
        do {
            ...
            checkDigit = getCheckDigit(...);
        } while (checkDigit == 10);
```

src/test/java/net/datafaker/helpers/IdNumberPatterns.java
asolntsev: Good point. Maybe.
Other numbers are theoretically possible, but we don't use them because it's legacy.

src/test/java/net/datafaker/idnumbers/HungarianIdNumberTest.java
kingthorin: Thanks, sorry I have a bad habit of overlooking conditions in do whiles (that's on me).

src/test/java/net/datafaker/helpers/IdNumberPatterns.java
cbxp: done.

asolntsev (approved):

kingthorin (approved):

PR #1559, opened 2025-05-26
Irish PPSN ID Number

89snake89:
Irish PPSN ID Number


src/test/java/net/datafaker/idnumbers/IrishIdNumberTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.Locale;
+import java.util.logging.Logger;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class IrishIdNumberTest {
+    private static final Logger LOG = Logger.getLogger(IrishIdNumberTest.class.getName());
+    private final IrishIdNumber irishIdNumber = new IrishIdNumber();
+    private final Faker faker = new Faker(new Locale("ie", "IE"));
```
asolntsev, line 15: @89snake89 Wait, isn't "ga" the language code of Ireland (Irish Gaelic)?
While "ie" is the code of international auxiliary language "Interlingue".

src/main/java/net/datafaker/idnumbers/IrishIdNumber.java
```diff
@@ -0,0 +1,132 @@
...
+        StringBuilder ppsn = new StringBuilder();
+        for (int d : digits) {
+            ppsn.append(d);
+        }
+        ppsn.append(checkChar).append(suffix);
+
+        return ppsn.toString();
+    }
+
+    public boolean validateAndCheckModulo23(String ppsn) {
```
asolntsev, line 97: I would make this method non-public (so called "package private").
Or wait, why this method is needed at all?

src/main/java/net/datafaker/idnumbers/IrishIdNumber.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Override
+    public String countryCode() {
+        return "IE";
+    }
+
+    @Override
+    public String generateInvalid(final BaseProviders faker) {
+        // Generate 7 digits
+        Random random = new Random();
```
asolntsev, line 47: Usually we don't use `new Random()` in DataFaker.
Instead, please use `faker` methods, e.g.

```java
String digits = faker.number().digits(7);
```

It's good for reproducible result.

src/main/java/net/datafaker/idnumbers/IrishIdNumber.java
```diff
@@ -0,0 +1,132 @@
...
+        for (int i = 0; i < 7; i++) {
+            digits.append(random.nextInt(10));
+        }
+        // Append always invalid character (es: 'Z')
+        return digits.append('Z').toString();
+    }
+
+    @Override
+    public String generateValid(final BaseProviders faker) {
+        Random random = new Random();
```
asolntsev, line 58: Instead of `new Random()`, please use `faker` methods, e.g.

```java
String digits = faker.number().digits(7);
boolean addSuffix = faker.bool().bool();
String suffix = addSuffix ? faker.options().option(allowedSuffixes) : "";
```

etc.

asolntsev (changes requested):

src/main/java/net/datafaker/idnumbers/IrishIdNumber.java
89snake89: I was thinking to have a separate method to divide and test the validity of the PPSN

src/test/java/net/datafaker/idnumbers/IrishIdNumberTest.java
kingthorin: Gemini's summary: 
> The ISO 639-1 language code for Irish (Ireland) is ga. The corresponding locale code, which combines the language and country, is ga-IE. This indicates the Irish language as spoken in Ireland.

src/main/java/net/datafaker/idnumbers/IrishIdNumber.java
asolntsev: It's a good idea to split code into smaller methods and test them separately. 
But I assume you don't expect your users to call this method, right? @89snake89 

But then remove keyword "public" from such methods. 
1. Then it becomes "package-private", and you can still can call this method from tests.
2. But end-users cannot call this method.

src/test/java/net/datafaker/idnumbers/IrishIdNumberTest.java
89snake89: You are correct, my mistake it must be new Locale("en", "IE")

src/main/java/net/datafaker/idnumbers/IrishIdNumber.java
```diff
@@ -0,0 +1,127 @@
...
+    public String countryCode() {
+        return "IE";
+    }
+
+    @Override
+    public String generateInvalid(final BaseProviders faker) {
+        // Generate 7 digits
+        String digits = faker.number().digits(7);
+        // Append always invalid character (es: 'Z')
+        return digits + "Z";
```
kingthorin, line 49: You could combine the two and return directly avoiding the intermediate store of digits

src/main/java/net/datafaker/idnumbers/IrishIdNumber.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @Override
+    public String generateValid(final BaseProviders faker) {
+        Random random = new Random();
+        int[] weights = {8, 7, 6, 5, 4, 3, 2};
+        // Generate 7 digits
+        int[] digits = faker.number().digits(7).chars().map(c -> c - '0').toArray();
+        String[] allowedSuffixes = {"A", "B", "H", "W"};
+        boolean addSuffix = random.nextBoolean(); // 50% chance
+        String suffix = addSuffix ? allowedSuffixes[random.nextInt(allowedSuffixes.length)] : "";
```
kingthorin, line 60: Here too, you could inline the random Boolean and skip the intermediate assignment

src/main/java/net/datafaker/idnumbers/IrishIdNumber.java
```diff
@@ -0,0 +1,127 @@
...
+                case 'A', 'B', 'H' -> extraChar - 'A' + 1;
+                case 'W' -> 0;
+                default -> 0;
+            };
+            sum += extraValue * 9;
+        }
+
+        // Calculate checksum character for modulus 23
+        int remainder = sum % 23;
+        char checkChar = (remainder == 0) ? 'W' : (char) ('A' + remainder - 1);
```
kingthorin, line 80: Here too

src/main/java/net/datafaker/idnumbers/IrishIdNumber.java
```diff
@@ -0,0 +1,127 @@
...
+
+        // Build the PPSN
+        StringBuilder ppsn = new StringBuilder();
+        for (int d : digits) {
+            ppsn.append(d);
+        }
```
kingthorin, line 86: Digits is a string right? Just append it directly.

src/main/java/net/datafaker/idnumbers/IrishIdNumber.java
```diff
@@ -0,0 +1,121 @@
...
+        return faker.number().digits(7) + "Z";
+    }
+
+    @Override
+    public String generateValid(final BaseProviders faker) {
+        Random random = new Random();
+        int[] weights = {8, 7, 6, 5, 4, 3, 2};
+        // Generate 7 digits
+        String digitsPpsn = faker.number().digits(7);
+        int[] digits = digitsPpsn.chars().map(c -> c - '0').toArray();
```
kingthorin, line 56:
```suggestion
        int[] digits = digitsPpsn.chars().map(c -> Character.getNumericValue(c)).toArray();
```

src/main/java/net/datafaker/idnumbers/IrishIdNumber.java
```diff
@@ -0,0 +1,121 @@
...
+           return false;
+       }
+        int sum = 0;
+        int[] weights = {8, 7, 6, 5, 4, 3, 2};
+        for (int i = 0; i < 7; i++) {
+            char c = ppsn.charAt(i);
+            if (!Character.isDigit(c)) {
+                return false;
+            }
+            sum += (c - '0') * weights[i];
```
kingthorin, line 97:
```suggestion
            sum += Character.getNumericValue(c) * weights[i];
```

src/main/java/net/datafaker/idnumbers/IrishIdNumber.java
```diff
@@ -0,0 +1,121 @@
...
+                case 'W' -> 0;
+                default -> -1; // Suffisso non valido
+            };
+            if (extraValue == -1) return false;
+            sum += extraValue * 9;
+        }
+        char checkChar = ppsn.charAt(7);
+        int remainder = sum % 23;
+        char expectedCheckChar = (remainder == 0) ? 'W' : (char) ('A' + remainder - 1);
+        return checkChar == expectedCheckChar;
```
kingthorin, line 113:
```suggestion
        int remainder = sum % 23;
        return ppsn.charAt(7) == (remainder == 0) ? 'W' : (char) ('A' + remainder - 1);
```

src/test/java/net/datafaker/idnumbers/IrishIdNumberTest.java
```diff
@@ -0,0 +1,30 @@
...
+import net.datafaker.helpers.IdNumberPatterns;
+import org.junit.jupiter.api.RepeatedTest;
+
+import java.util.Locale;
+import java.util.logging.Logger;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class IrishIdNumberTest {
+    private static final Logger LOG = Logger.getLogger(IrishIdNumberTest.class.getName());
```
kingthorin, line 13:
```suggestion
```

kingthorin:
Seems to need a spotless apply

kingthorin (approved):
LGTM

asolntsev (approved):

PR #1562, opened 2025-05-30
Support custom Random Generators

zyclonite:
Add support for using custom `RandomGenerators` like `SplittableRandom` for improved speed and less overhead

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Locality.java
+++ b/src/main/java/net/datafaker/providers/base/Locality.java
@@ -3,11 +3,10 @@ package net.datafaker.providers.base;
 import net.datafaker.annotations.Deterministic;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
-import java.util.Random;
 import java.util.Set;
+import java.util.random.RandomGenerator;
 
 /**
  * Generates random locales in different forms.
@@ -97,7 +96,7 @@ public class Locality extends AbstractProvider<BaseProviders> {
      * @param random random number generator (can utilize seed for deterministic random selection)
      * @return String of a randomly selected locale (e.g. "es", "es-MX")
      */
-    public String localeStringWithRandom(Random random) {
+    public String localeStringWithRandom(RandomGenerator random) {
 
         // Randomly select a locale from list of all locales supported
         int randomIndex = random.nextInt(LOCALES.size());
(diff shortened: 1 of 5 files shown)


src/main/java/net/datafaker/providers/base/Locality.java
```diff
@@ -97,7 +98,7 @@ public String localeString() {
      * @param random random number generator (can utilize seed for deterministic random selection)
      * @return String of a randomly selected locale (e.g. "es", "es-MX")
      */
-    public String localeStringWithRandom(Random random) {
+    public String localeStringWithRandom(RandomGenerator random) {
```
kingthorin, line 101: Will this break existing code? Does Random implement RandomGenerator?

kingthorin (commented):
Could you revert the formatting changes? 

Not that any of them are wrong but just to keep this focused on the relevant changes.

zyclonite:
> Could you revert the formatting changes?
> 
> Not that any of them are wrong but just to keep this focused on the relevant changes.

this was by mistake, removed it now

src/main/java/net/datafaker/providers/base/Locality.java
```diff
@@ -118,13 +119,13 @@ public String localeStringWithoutReplacement() {
...
      */
-    public synchronized String localeStringWithoutReplacement(Random random) {
+    public synchronized String localeStringWithoutReplacement(RandomGenerator random) {
         if (shuffledLocales.isEmpty() || shuffledLocaleIndex >= shuffledLocales.size() - 1) {
             // copy list of locales supported into shuffledLocales
             shuffledLocales.clear();
             shuffledLocales.addAll(LOCALES);
             shuffledLocaleIndex = 0;
-            Collections.shuffle(shuffledLocales, random);
+            Collections.shuffle(shuffledLocales, (Random) random);
```
asolntsev: Maybe I missed something, but what is the point of declaring parameter as `RandomGenerator` if we anyway cast it to `Random`? Then no any other types would work here...

zyclonite: because this new method does start to exist in jdk21... this would break the compatibility for jdk17

zyclonite: but good point - seems there is not test for that...

zyclonite: added a way to fix it now... but the same is true for RgxGen, so this would need to be fixed first

src/test/java/net/datafaker/providers/base/LocalityTest.java
```diff
@@ -82,6 +84,19 @@ void localeStringWithoutReplacement() {
...
     }
 
+    @Test
+    void localeStringWithoutReplacementWithCustomRandomGenerator() {
+        RandomGenerator random = new SplittableRandom();
+        // loop through all supported locales
+        for (int i = 0; i < 2; i++) {
+            Set<String> returnedLocales = IntStream.range(0, locality.allSupportedLocales().size())
+                .mapToObj(j -> locality.localeStringWithoutReplacement(random))
+                .collect(Collectors.toSet());
```
kingthorin, line 94: I think in Java 17+ you can just `.toSet()`

src/test/java/net/datafaker/providers/base/LocalityTest.java
```diff
@@ -82,6 +84,19 @@ void localeStringWithoutReplacement() {
         }
     }
 
+    @Test
+    void localeStringWithoutReplacementWithCustomRandomGenerator() {
+        RandomGenerator random = new SplittableRandom();
+        // loop through all supported locales
+        for (int i = 0; i < 2; i++) {
```
kingthorin, line 91: Why twice?

zyclonite: this is just a copy with a different rng

src/test/java/net/datafaker/providers/base/LocalityTest.java
zyclonite: trying to keep the same style - i would either change all or none

zyclonite:
depends on https://github.com/curious-odd-man/RgxGen/pull/108

src/test/java/net/datafaker/providers/base/LocalityTest.java
kingthorin: Fair enough

zyclonite:
the new version 3.0 of RgxGen supports RandomGenerator now
upgraded the version and fixed the implementation

asolntsev (approved):

PR #1572, opened 2025-06-22
TomlTransformer implementation

badoken:
Adding to the collection of Transformers (JsonTransformer, YamlTransformer etc.)
**TOML support**,  to generate data in TOML format ([specs](https://toml.io/en/v1.0.0#objectives)).

The format is often used in Python and Rust.

The implementation is based on YamlTransformer and its tests structure

### DIFF
--- a/src/main/java/net/datafaker/transformations/TomlTransformer.java
+++ b/src/main/java/net/datafaker/transformations/TomlTransformer.java
@@ -1,0 +1,131 @@
+package net.datafaker.transformations;
+
+import net.datafaker.sequence.FakeSequence;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class TomlTransformer<IN> implements Transformer<IN, String> {
+    @Override
+    public String apply(IN input, Schema<IN, ?> schema) {
+        if (schema == null) {
+            return "";
+        }
+        StringBuilder sb = new StringBuilder();
+        writeSchema(sb, input, schema, "");
+        int len = sb.length();
+        int sepLen = LINE_SEPARATOR.length();
+        if (len >= sepLen && sb.lastIndexOf(LINE_SEPARATOR, len - sepLen) >= 0) {
+            sb.setLength(len - sepLen);
+        }
+        return sb.toString();
+    }
+
+    @Override
+    public String generate(Iterable<IN> input, Schema<IN, ?> schema) {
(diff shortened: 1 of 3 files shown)


src/main/java/net/datafaker/transformations/TomlTransformer.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    public String apply(IN input, Schema<IN, ?> schema) {
+        if (schema == null) {
+            return "";
+        }
+        StringBuilder sb = new StringBuilder();
+        writeSchema(sb, input, schema, "");
+        int len = sb.length();
+        int sepLen = LINE_SEPARATOR.length();
+        if (len >= sepLen && sb.substring(len - sepLen).equals(LINE_SEPARATOR)) {
```
snuyanzin:
```suggestion
        if (sb.endsWith(LINE_SEPARATOR)) {
```
could it be simplified like that?

snuyanzin:
great to see new formats appearing
can you please also 
1 update doc (https://github.com/datafaker-net/datafaker/blob/main/docs/documentation/schemas.md)
2. rebase the PR (in order to eliminate merge commit from the PR)

src/main/java/net/datafaker/transformations/TomlTransformer.java
badoken: It's a `StringBuilder`, that doesn't have `endsWith`, unfortunately.
It can be first built, then checked with nicer `endsWith` and then the string can be cut again, but I assume that would be a bit less performant due to the string rebuild twice.
Let me know if you think it still makes sense to change

badoken:
Thanks @snuyanzin for the review, I've rebased the PR and extended the doc

src/main/java/net/datafaker/transformations/TomlTransformer.java
snuyanzin: ah, that's true

then if it is StringBuilder I think it is possible to use indexOf/lastIndexOf, something like
```java
if (len >= sepLen && sb.lastIndexOf(LINE_SEPARATOR, len - sepLen) >= 0)
```
then no need to create new strings
WDYT?

badoken: Sure, updated :+1:

snuyanzin:
Thank you for your contribution @badoken

PR #1582, opened 2025-07-06
Attempt to fix maven deploy using Maven Central

bodiam:


kingthorin (approved):
I have no idea what else might be needed but this seems reasonable

bodiam:
@kingthorin I have really no idea what I'm doing here myself, just experimenting to see how far I get.

kingthorin:
Are you going to remove 2.4.4 and try again now that this is in place?

Issue #1593, opened 2025-07-22
Singapore NricNumber generation algorithm produces invalid value


hieule3004:
**Describe the bug**

`net.datafaker.idnumbers.NricNumber` class `CODE` field has length 8 but is used to compute pair-wise weighted sum with random digits of length 7. The correct algorithm should have first value in array i.e. `0` removed.

**To Reproduce**
Code to reproduce

**Expected behavior**

The value of `new Faker(new Locale("en", "SG")).idNumber().singaporeanFin()` should be valid with any online NRIC validator.

**Versions:**
Both Faker v1 and v2

**Additional context**
I have tested by patching the array and it works as expected.

asolntsev:
@hieule3004 Thank you for reporting the issue! Can you please review https://github.com/datafaker-net/datafaker/pull/1594 ?

hieule3004:
Hi @asolntsev, thanks for the quick fix. The PR looks good to me.

PR #1594, opened 2025-07-22
#1593 fix Singapore ID number generation

asolntsev:
`SingaporeIdNumber.CODE` array has length 8 but is used to compute pair-wise weighted sum with random digits of length 7. This commit removes the first element (0).

I've tested results with online NRIC validator https://nric.biz/

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Number.java
+++ b/src/main/java/net/datafaker/providers/base/Number.java
@@ -24,6 +24,17 @@ public class Number extends AbstractProvider<BaseProviders> {
     }
 
     /**
+     * Returns an array of random numbers from 0 to 9 (both inclusive) of given length
+     */
+    public int[] randomDigits(int length) {
+        int[] result = new int[length];
+        for (int i = 0; i < length; i++) {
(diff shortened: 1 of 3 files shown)


src/main/java/net/datafaker/providers/base/Number.java
```diff
@@ -23,6 +23,17 @@ public int randomDigit() {
         return faker.random().nextInt(0, 9);
     }
 
+    /**
+     * Returns an array of random numbers from 0 to 9 (both inclusive) of given length
+     */
+    public int[] randomDigits(int length) {
```
kingthorin, line 29: Throw an exception if length is negative?

asolntsev: Yes, the exception will be thrown out-of-the-box:

```java
java.lang.NegativeArraySizeException: -42

	at net.datafaker.providers.base.Number.randomDigits(Number.java:30)
	at net.datafaker.providers.base.NumberTest.randomDigits(NumberTest.java:49)
```

kingthorin (approved):

PR #1600, opened 2025-08-02
Removing parent, not sure if still needed.

bodiam:


bodiam:
I did the same with Claude and got similar results. I think it's no longer needed with the move to Maven Central, and it was giving issues with the Snapshot publishing, so I removed it, and snapshots seem to work again.

PR #1606, opened 2025-08-12
Refactor tests extending `BaseFakerTest`

asolntsev:
1. Remove `extends BaseFakerTest` from all test that don't need to run `testProviderList` and `testNoDuplications`
2. Simplify few tests
3. Remove `setFaker()`

### DIFF
--- a/src/test/java/net/datafaker/providers/base/InternetTest.java
+++ b/src/test/java/net/datafaker/providers/base/InternetTest.java
@@ -5,8 +5,6 @@ import org.apache.commons.validator.routines.EmailValidator;
 import org.assertj.core.api.Condition;
 import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
-import org.mockito.Mockito;
-import org.mockito.Spy;
 
 import java.net.URL;
 import java.util.ArrayList;
@@ -21,12 +19,12 @@ import static org.assertj.core.api.Assertions.anyOf;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
 
 class InternetTest {
 
     public static final Pattern IPV6_HOST_ADDRESS = Pattern.compile("[0-9a-fA-F]{1,4}(:([0-9a-fA-F]{1,4})){1,7}");
-    @Spy
-    private BaseFaker mockedFaker;
+    private final Faker faker = new Faker();
 
     @RepeatedTest(100)
     void testUsername() {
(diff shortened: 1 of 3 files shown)


kingthorin:
Thanks!!

src/test/java/net/datafaker/providers/base/InternetTest.java
```diff
@@ -39,10 +38,20 @@ void emailSubject() {
...
+        BaseFaker mockedFaker = new BaseFaker() {
+            @Override
+            public Name name() {
+                return name;
+            }
+        };
+        assertThat(mockedFaker.internet().username())
+            .doesNotContain(" ", "'")
+            .matches("^(\\w+)\\.(\\w+)$")
+            .matches("^\\p{javaLowerCase}+\\.\\p{javaLowerCase}+$");
```
snuyanzin, line 54: ideally this test should be rewritten

currently it doesn't check much
If I replace implementation of `username` with `return a.a` it will pass

snuyanzin: even more: if you compare output of 
`firstName() + "." + lastName()` vs `javaLowerCase` for `internet().username()` it will not match because tests are running in TR locale
first will give
`jin.quan`
second
`jın.quan` (with turkish dotless i)

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
```diff
@@ -379,7 +369,7 @@ public String lastName() {
         }
 
         public String hello() {
-            return "Hello";
+            return greetings.get(counter.getAndIncrement());
```
snuyanzin, line 372: I would vote for either circular approach or hard fail with a nicer exception

otherwise the one who will use something like that
```yml
sameResolution: "#{hello} #{hello} #{hello} #{hello}"
```
will spend lots of time for debugging

src/test/java/net/datafaker/providers/base/InternetTest.java
asolntsev: Emmmm... 

> currently it doesn't check much
Yes, this is true for most DF tests. It's because they call random generator and cannot fully verify the random output.

Anyway, now this test checks more than before.

> it will not match because tests are running in TR locale

Sorry, I don't understand how to make this test fail. For me, it works with both English and Turkish locale.

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
asolntsev: Maybe I don't understand, but current implementation do exactly that: fails with a clear exception:

```java
java.lang.RuntimeException: Failed to invoke MethodAndCoercedArgs[method=DummyService.hello()...]
...
Caused by: java.lang.IndexOutOfBoundsException: Index: 1 Size: 1
```

src/test/java/net/datafaker/providers/base/InternetTest.java
kingthorin: It isn't lower casing the value(s), it's comparing against a character set of all Java lower case characters. (From my understanding)

asolntsev: @kingthorin Yes. `username` returns lower case, and `.matches("...\p{javaLowerCase}...")` verifies that this is lower case.

snuyanzin: @asolntsev 
>Sorry, I don't understand how to make this test fail.

more details are at

https://github.com/datafaker-net/datafaker/pull/1669

PR #1608, opened 2025-08-13
Remove duplicates in "*.yml" files

asolntsev:
Add test for verifying that "*.yml" files don't contain duplicates

It appears that current *.yml files do contain quite a lot of duplicates. :)
Had to fix them all (now we don't need tests `testNoDuplications` in `ProviderListTest` and its inheritors).


kingthorin:
For my two cents the duplicates should be removed. Assuming they aren't just dupes in file but dupes in "section" or whatever you want to call it. (Haven't looked at code yet).

Removing dupes ensures two things:
- No favouritism or fake randomness.
- DF gets slightly smaller.

src/main/resources/hy.yml
```diff
@@ -162,7 +162,7 @@ hy:
         NU: Նիուե
         NF: Նորֆոլկ Կղզի
         MP: Հյուսիսային Մարիանյան Կղզիներ
-        NO: Նորվեգիա
+        'NO': Նորվեգիա
```
kingthorin, line 165: Why is this necessary?

(I don't disagree, just wondering if it's something I should watch for)

kingthorin: Is NO a reserved word?

asolntsev: @kingthorin I am glad you asked it :)
The problem is that Yaml parser reads `NO: Նորվեգիա` as `false: Նորվեգիա` :) :) :) Hilarious!

By the way, is `country_by_code` property in yml files needed at all?
Afaik it's not used anywhere...

src/test/java/net/datafaker/NoDuplicatesInYmlTest.java
```diff
@@ -0,0 +1,73 @@
...
+    private Map<String, Set<Object>> findDuplicates(Object node) {
+        Map<String, Set<Object>> errors = new HashMap<>();
+        findDuplicatesRecursively(node, "", errors);
+        return errors;
+    }
+
+    private void findDuplicatesRecursively(Object node, String path, Map<String, Set<Object>> errors) {
+        if (node instanceof Map<?, ?> map) {
+            for (var entry : map.entrySet()) {
+                String key = (String) entry.getKey();
```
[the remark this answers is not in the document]
asolntsev: Yes, it may cause ClassCastException - but it's exactly what I want. This is test, and I want to check that all keys are Strings.

asolntsev:
> Removing dupes ensures two things:
> 
> * No favouritism or fake randomness.

I agree. I also vote for removing duplicates.

BUT

I see few places where duplicates probably are intentionally used:
image

**Is it possible that DF creators intentionally wanted to prefer one format over the other?** @kingthorin @bodiam

src/main/resources/hy.yml
kingthorin: I'm have no idea if it's used or not. You could trim it and run the tests and see if anything fails. Or check the blame for its addition and see if it was just added from the start and never used.

kingthorin:
Assuming that's Canada (I'm Canadian) I don't see a reason why it would need to be three name preferred vs having a prefix.

Also a contributor recently added some functionality for weighted generation which would seem to fit that use case better (if there needs to be an intentional preference). In my opinion.

I've only been working on DF for a year or two, I don't have much insight into the original choices or early history.

asolntsev:
> the original choices or early history

I've asked [here](https://github.com/orgs/faker-ruby/discussions/3112) and [here](https://github.com/faker-ruby/faker/pull/491#issuecomment-3188318566) - no answers by far :(

kingthorin:
Im happy with this. I haven't had a detailed review of the YAML files but am trusting that de-duplicating is somewhat straight forward.

I'd really like @bodiam to add his thoughts before merging in case there's background I'm missing.

bodiam:
Hi all! I think some of the duplication was actually done on purpose. I know I've done a few myself on purpose. For example, something like this:

```
    - "#{male_first_name} #{male_last_name}"
      - "#{male_first_name} #{male_last_name}"
      - "#{male_first_name} #{male_last_name}"
      - "#{male_first_name} #{male_last_name}"
      - "#{female_first_name} #{female_last_name}"
      - "#{female_first_name} #{female_last_name}"
      - "#{female_first_name} #{female_last_name}"
```

The reason why it's like this is to change the distribution of the data. Some patterns are much more unlikely than just having a first and last name, and since we don't have an easy way to change the distribution in the config, duplicating some patterns is the current way of effecting this.

However, for other changes, like this:

```
      country_code: [CH, CH, CH, DE, AT, US, LI, US, HK, VN]
```

to 

```
      country_code: [CH, DE, AT, US, LI, HK, VN]
```

I have no problem with, but something like this:

```
      domain_suffix: [com, com, com, net, org, it, it, it]
      domain_suffix: [com, net, org, it]
```

This was done on purpose, since .com and .it addresses are much more common in Italy than .net and .org.

So, I think this PR needs some more attention perhaps?

bodiam:
> **Is it possible that DF creators intentionally wanted to prefer one format over the other?** @kingthorin @bodiam

yep, exactly. It's to change the distribution of the data to favour more  common patterns. I'm not aware of every situation where this this is done on purpose or by accident (I can image that currencies should be de-duped, but some name patterns shouldn't). 

I'd like to help, and while I can see some good value in the PR (including fixing the typos, thanks!!), the PR is a little too big to review effectively right now.

kingthorin:
@asolntsev maybe cherry pick each commit into its own PR? (Add the tests last?)

asolntsev:
No problems, i cam temove one-two commits from this PR. But tomorrow.

asolntsev:
> yep, exactly. It's to change the distribution of the data to favour more common patterns.

@bodiam Let me ask first. Why do you think it's a good idea to favour more common patterns? What is the purpose?
At least in tests, it's rather a good idea to generate rare cases, not-so-usual names etc. It increases chances to reproduce some bugs etc.

bodiam:
> et me ask first. Why do you think it's a good idea to favour more common patterns? What is the purpose?
> At least in tests, it's rather a good idea to generate rare cases, not-so-usual names etc. It increases chances to reproduce some bugs etc.

Because people, like me, use this library to generate real looking data. In my country it's possible to have a name like:

```
firstname middle_name middle_name lastname
Karel van der Spoel
```

It's uncommon, but it's there. More common is just a `first_name last_name` pattern. So, if you generate 100 names, and every 3rd name is `van der`, that's a very unlikely distribution of names. It's the same with generating domain names for Italy: I can imagine a ".it" domain is more likely than a ".org" one. 

It's not a perfect distribution or a perfect solution, but it's the way we've done it so far, and it works reasonably well.

If you want to test rare cases, increase the test duration. If I had more time I'd change some other methods as well, for example to generate date of birth dates, since right now is an even distribution, which means people have the same chance to be 100 years old as being 50 years old, which seems odd to me, or at least, unrealistic.

asolntsev:
> to generate real looking data

Thank you, I've reverted the changes of names. Now duplicates are allowed for few data types (names, company names and job titles) - so they will be more realistic.


@bodiam @kingthorin As an idea, what if we declare the desired distributions in percentages instead of duplicating the values? Like, instead of
```yml
- name:
  - "#{Name.last_name} #{suffix}"
  - "#{Name.last_name} #{suffix}"
  - "#{Name.man_last_name} a #{Name.man_last_name} #{suffix}"
```

we would write:
```yml
- name:
 - 60%: "#{Name.last_name} #{suffix}"
 - 40%: "#{Name.man_last_name} a #{Name.man_last_name} #{suffix}"
```

or even with lists:
```yml
- domain:
  - 90%: ".it"
  - 10%: [".ee", ".lv", ".lt", ".ca", ...]
```

PR #1618, opened 2025-08-20
New method: toLocalPart(String name)

rcriosbr:
Created a new toLocalPart(String name) method which creates usernames from a provided name. Refactored the username() method to use the new method.

Maybe it's just my obsessive-compulsive disorder, but I needed to generate an email address that was made up of the same names that were generated by faker.name(). I couldn't have a name like "Han Solo" and an email like "minnie.mouse". It had to be consistent.

Of course, I could generate a localPart locally and call the emailAddress(String localPart), but it seemed like a good idea to include this in the library.

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Internet.java
+++ b/src/main/java/net/datafaker/providers/base/Internet.java
@@ -10,6 +10,7 @@ import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 import java.text.Normalizer;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 import java.util.regex.Pattern;
@@ -19,20 +20,21 @@ import java.util.regex.Pattern;
  */
 public class Internet extends AbstractProvider<BaseProviders> {
     private static final Pattern COLON = Pattern.compile(":");
+    private static final Pattern LOCALPART_PATTERN = Pattern.compile("[^a-z0-9]");
     private static final List<String> HTTP_SCHEMES = List.of("http://", "https://");
 
     protected Internet(BaseProviders faker) {
         super(faker);
     }
 
-
     /**
-     * A lowercase username composed of the first_name and last_name joined with a '.'. Some examples are:
+     * A lowercase username composed of the first_name and last_name joined with a
+     * '.'. Some examples are:
      * <ul>
-     *     <li>(template) {@link Name#firstName()}.{@link Name#lastName()}</li>
-     *     <li>jim.jones</li>
-     *     <li>jason.leigh</li>
-     *     <li>tracy.jordan</li>
+     * <li>(template) {@link Name#firstName()}.{@link Name#lastName()}</li>
+     * <li>jim.jones</li>
+     * <li>jason.leigh</li>
+     * <li>tracy.jordan</li>
      * </ul>
      *
      * @return a random two part username.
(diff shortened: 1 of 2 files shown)


bodiam:
I basically agree with everything @asolntsev mentions above. It's also not clear to me what this method does, and when I read the documentation, I see this:

```
     * @param name the name to convert
     * @return the local part of the email address
```

So I pass in a name, and I get back the local part of the email address. Then I wonder: which email address? I understand now that you can use it as a local part of the email address, but that's not really what the method does, it's more like: 

`formatStringToLowerCaseAndDots`. Yes, that's not a great name, I get that, but it covers better what it does. But wouldn't something like `toUsername(String)`or `formatUsername(String)` be more clear? We already have a username() method, and this seems very closely aligned, and I think it would be clear to have `username()` and `toUsername(String)` as a pair.

It would be nice indeed to not reformat the code as part of this PR, I don't think we're too religious on it, but even with 'hide whitespace changes' in GitHub there's a lot of noise, so it would be preferable if you could undo that.

It also seems like your change broke the behaviour of the `username()` method, perhaps something to look into as well?

PS: thanks for your contribution btw. We absolutely appreciate it, and are more than happy to merge this into the code base if we can address the above items. We're here to help!

kingthorin:
Could you rollback all the whitespace changes so this is focused on actual changes?
Also it seems you need to fix some tests if you really expect this to land.

rcriosbr:
@asolntsev I was creating data to test a DTO (NaturalPerson) when I came across with DF library. So I followed the documentation and did:
```
Faker faker = new Faker(Locale.of("es", "ES"));
String name = faker.name().nameWithMiddle();
String email = faker.internet().emailAddress();
```
The problem is that I ended up with two pieces of inconsistent data: the person's name remained as `Daniela Núñez Lovato` and the email address as `victor.guerra@hotmail.com`.

So, looking at the documentation, I saw that there was a method `Internet#emailAddress(String localPart)` and I ended up creating a helper in my test class to create a localPart from a name. And this method is what I sent in the PR.

Regarding the name "local part", I used it because, in addition to being the official name used in the RFC, there are already other references in the `Internet` class, so it seemed natural to name the method this way.

I will fix items 2 and 3.

rcriosbr:
@bodiam The goal is to have consistency/coherence between names and emails, as I commented in my previous answer.

In my mind, when I read “username,” the term reminds me of security (such as credentials) rather than email. And in RFC 5322 (Internet Message Format), that's how they refer to it (`addr-spec = local-part "@" domain`). And by the way, I was thinking of proposing another change and creating a new class called `Credential ` and putting the `username ` and `password ` methods in it, which strictly speaking (at least in my mind) are not functions related only to the internet...

So, since there was a method that receives a string with the name `localPart` as a parameter, it seemed natural to me to call the method this way: a method that returns the parameter of another method. Like:

```
Faker faker = new Faker(Locale.of("es", "ES"));
String name = faker.name().nameWithMiddle(); // Daniela Núñez Lovato
String localPart = faker.internet().toLocalPart(name); // daniela.lovato
String email = faker.internet().emailAddress(localPart); // daniela.lovato@yahoo.es
```
I will fix the tests

asolntsev:
> The goal is to have consistency/coherence between names and emails, as I commented in my previous answer.

Now I understand your logic. Yes, it seems reasonable to keep using "local part" term if it was already used.

@rcriosbr 
But I would suggest to make the API even simpler without using "local part" at all.
Let's change `emailAddress` method so that it accepts the name parameter and converts it to a "local name":

```java
Faker faker = new Faker(Locale.of("es", "ES"));
String name = faker.name().nameWithMiddle(); // Daniela Núñez Lovato
String email = faker.internet().emailAddress(name); // daniela.lovato@yahoo.es
```

bodiam:
> @bodiam The goal is to have consistency/coherence between names and emails, as I commented in my previous answer.
> 
> In my mind, when I read “username,” the term reminds me of security (such as credentials) rather than email. And in RFC 5322 (Internet Message Format), that's how they refer to it (`addr-spec = local-part "@" domain`). And by the way, I was thinking of proposing another change and creating a new class called `Credential ` and putting the `username ` and `password ` methods in it, which strictly speaking (at least in my mind) are not functions related only to the internet...
> 
> So, since there was a method that receives a string with the name `localPart` as a parameter, it seemed natural to me to call the method this way: a method that returns the parameter of another method. Like:
> 
> ```
> Faker faker = new Faker(Locale.of("es", "ES"));
> String name = faker.name().nameWithMiddle(); // Daniela Núñez Lovato
> String localPart = faker.internet().toLocalPart(name); // daniela.lovato
> String email = faker.internet().emailAddress(localPart); // daniela.lovato@yahoo.es
> ```
> 
> I will fix the tests

Makes sense! I have no issue with making a Credential class either, and putting the username in there (we can delegate the current username() to the Credential username, and make it deprecated, that's fine.

Issue #1617, opened 2025-08-19
` java.lang.NoSuchMethodException` is logged on `Text.text` expression with length parameter


valfirst:
**To Reproduce**
```java
import net.datafaker.Faker;

public class Test
{
    public static void main(String[] args)
    {
        Faker faker = new Faker();

        // Passed
        System.out.println(faker.text().text(7));

        // Failed
        System.out.println(faker.expression("#{Text.text '7'}"));
    }
}
```

**Actual behavior**
The error is logged (the output is created using `2.4.1` version):
```java
twrdeyi
Aug 19, 2025 10:41:45 PM net.datafaker.service.FakeValuesService coerceArguments
SEVERE: Unable to coerce 7 to TextRuleConfig via TextRuleConfig(String) constructor
net.datafaker.service.FakeValuesService$NoSuchMethodRuntimeException: java.lang.NoSuchMethodException: net.datafaker.providers.base.Text$TextRuleConfig.<init>(java.lang.String)
	at net.datafaker.service.FakeValuesService.getConstructorWithString(FakeValuesService.java:1202)
	at net.datafaker.service.FakeValuesService.coerceArguments(FakeValuesService.java:1085)
	at net.datafaker.service.FakeValuesService.accessor(FakeValuesService.java:984)
	at net.datafaker.service.FakeValuesService.lambda$retrieveMethodAccessor$25(FakeValuesService.java:949)
	at java.base/java.util.Map.computeIfAbsent(Map.java:1067)
	at net.datafaker.service.FakeValuesService.retrieveMethodAccessor(FakeValuesService.java:949)
	at net.datafaker.service.FakeValuesService.resolveFakerObjectAndMethod(FakeValuesService.java:931)
	at net.datafaker.service.FakeValuesService.resolveExpression(FakeValuesService.java:808)
	at net.datafaker.service.FakeValuesService.resExp(FakeValuesService.java:722)
	at net.datafaker.service.FakeValuesService.resolveExpression(FakeValuesService.java:646)
	at net.datafaker.service.FakeValuesService.expression(FakeValuesService.java:521)
	at net.datafaker.providers.base.BaseFaker.expression(BaseFaker.java:411)
	at Test.main(Test.java:13)
Caused by: java.lang.NoSuchMethodException: net.datafaker.providers.base.Text$TextRuleConfig.<init>(java.lang.String)
	at java.base/java.lang.Class.getConstructor0(Class.java:3833)
[...]

PR #1620, opened 2025-08-20
#1617 avoid useless `NoSuchMethodRuntimeException` in logs

asolntsev:


bodiam:
LGTM, feel free to merge this. Thanks for fixing this so quickly!

asolntsev:
Seems we should release [2.4.5](https://github.com/datafaker-net/datafaker/milestone/3) soon?

bodiam:
> Seems we should release [2.4.5](https://github.com/datafaker-net/datafaker/milestone/3) soon?

I'm happy if we release a new version. You should be able to make the release, so if you want to, feel free, I have no issue with that, but perhaps we can get those 2 open PRs in?

Issue #1623, opened 2025-08-20
Test `testWeightedArrayElementWithMultipleElements` fails time-to-time


asolntsev:
We have a flaky test in DataFaker:

```java
Error:  Failures: 
Error:    WeightedRandomSelectorTest.testWeightedArrayElementWithMultipleElements:124 
Expecting actual:
  132
to be close to:
  166
by less than 20% but difference was 20.481927710843372%.
(a difference of exactly 20% being considered valid)
```

For example, it failed here: https://github.com/datafaker-net/datafaker/actions/runs/17111857693/job/48534827890

kingthorin:
Can we allow 20% plus/minus 1% ?

I don't think it is terribly surprising that sometimes it isn't going to be exact.

(I have no idea how the test is setup.)

asolntsev:
> Can we allow 20% plus/minus 1% ?

Error tolerance in this test is 20%. Yes, we can change it to 21%. :)
But I guess sooner or later it can fail again even with 21%. 


> (I have no idea how the test is setup.)

The test is essentially:

#### Given
- One (weight: 1)
- Two (weight: 2)
- Three (weight: 3)

#### When
Generate 600 random values

#### Then
Assert that "One" was ~100 times, "Two" was ~200 times, and "Three" was ~300 times (plus/minus 20%).

kingthorin:
Oh the 20% is the uh error range.

I'm not sure then, we'd have to look at the implementation and see if it can be made more specific.

asolntsev:
@kingthorin  I've played with plain Java `Random` class, and from my experiments, 40% is enough for a stable test. 

```java
/**
 * 15% - fails 126 or 1000
 * 20% - fails 36 or 1000
 * 25% - fails 12 or 1000
 * 30% - fails 4 or 1000
 * 35% - fails 1 or 1000
 * 40% - fails 0 or 1000
 */
public class JavaRandomServiceTest {

    private final Percentage tolerance = withinPercentage(40);

    @RepeatedTest(1000)
    void todo() {
        Random random = new Random();
        Map<Integer, AtomicInteger> results = new HashMap<>();
        for (int i = 0; i < 100; i++) {
            int value = random.nextInt(2);
            results.computeIfAbsent(value, k -> new AtomicInteger(0)).incrementAndGet();
        }
        System.out.println(results);
        assertThat(results.get(0).get()).isCloseTo(50, tolerance);
[...]

PR #1626, opened 2025-08-22
#1623 fix flaky test WeightedRandomSelectorTest

asolntsev:
instead of using `Random` which can sometimes generate too "un-random" values, we now user custom `EvenlyDistributedRandomGenerator` which returns values in random order, but guarantees that all possible values are covered.


kingthorin (approved):
Thanks!

PR #1637, opened 2025-09-04
Update CreditCard formats in Finance provider

Fishbowler:
Fix random data generation of credit cards

- Removes deprecated card types (Switch, Solo)
- Removes an invalid Mastercard range
- Updates JCB and UnionPay card formats

I've put the citation of sources in the commits for anyone who's interested.


asolntsev:
@Fishbowler Thank you for the PR, looks good!
Now need to fix few tests... 

```java
19:42:04.733 [ERROR]   FinanceTest.unionpayCard:104 
Expecting actual:
  "6441241451881464"
to start with:
  "62"
```

src/test/java/net/datafaker/providers/base/FinanceTest.java
```diff
@@ -101,7 +101,8 @@ void discoverCard() {
     @RepeatedTest(100)
     void unionpayCard() {
         String creditCard = finance.creditCard(CreditCardType.UNIONPAY).replace("-", "");
-        assertThat(creditCard).startsWith("62").hasSize(16);
+        assertThat(creditCard).hasSize(16);
+        assertThat(creditCard.startsWith("62") || creditCard.startsWith("64") || creditCard.startsWith("65") || creditCard.startsWith("81")).isTrue();
```
asolntsev, line 105: Such checks have one problem: when this test fails, you cannot see the actual credit card number in the error message. You will only see "Test failed. Expected: true, actual: false".

(But the previous code also had the same problem)

asolntsev (approved):

kingthorin:
Thanks!

src/test/java/net/datafaker/providers/base/FinanceTest.java
Fishbowler: Fixed in #1643

Fishbowler:
The invalid Mastercard range is causing me some grief in a project. Any chance that v2.4.5 is imminent?

PR #1640, opened 2025-09-04
Add missing javadoc for the Finance provider

Fishbowler:
I needed better docs from this page recently, and hated digging in the source for the answer. Fixed!

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Finance.java
+++ b/src/main/java/net/datafaker/providers/base/Finance.java
@@ -69,6 +69,13 @@ public class Finance extends AbstractProvider<BaseProviders> {
         return countryCodeToBasicBankAccountNumberPattern.keySet();
     }
 
+    /**
+     * Generates a random credit card number of the specified type.
+     * The number will pass the LUHN check.
+     *
+     * @param creditCardType the type of credit card to generate (see {@link CreditCardType} for options)
+     * @return a valid credit card number of the specified type
+     */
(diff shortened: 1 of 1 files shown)


src/main/java/net/datafaker/providers/base/Finance.java
```diff
@@ -115,18 +126,34 @@ public String bic() {
...
         List<String> countryCodes = new ArrayList<>(countryCodeToBasicBankAccountNumberPattern.keySet());
         String randomCountryCode = countryCodes.get(faker.random().nextInt(countryCodes.size()));
         return iban(randomCountryCode);
     }
 
+    /** Generates an IBAN for the given country code.
+     * The generated IBAN will pass the checksum test.
+     * @param countryCode the 2-letter country code for the IBAN.
+     * @return a valid IBAN
+     */
```
asolntsev, line 148: Though it's minor, usually the first line contains only "/**" characters. And the text starts from the second line:

```
    /**
     * Generates an IBAN for the given country code.
     * The generated IBAN will pass the checksum test.
     * @param countryCode the 2-letter country code for the IBAN.
     * @return a valid IBAN
     */
```

asolntsev (approved):

src/main/java/net/datafaker/providers/base/Finance.java
Fishbowler: Good spot. Fixed!

PR #1643, opened 2025-09-04
Make Finance provider's UnionPay test more readable, especially on failure

Fishbowler:
Makes a test in the Finance provider easier to read, and gives actionable feedback to the developer when it fails.

Addresses [the comment](https://github.com/datafaker-net/datafaker/pull/1637#discussion_r2323372486) left on the previous PR.

### DIFF
--- a/src/test/java/net/datafaker/providers/base/FinanceTest.java
+++ b/src/test/java/net/datafaker/providers/base/FinanceTest.java
@@ -10,7 +10,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import java.util.List;
 import java.util.Collection;
 import java.util.Set;
-import java.util.stream.Stream;
 
 class FinanceTest extends BaseFakerTest {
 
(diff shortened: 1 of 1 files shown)


src/test/java/net/datafaker/providers/base/FinanceTest.java
```diff
@@ -102,7 +102,9 @@ void discoverCard() {
     void unionpayCard() {
         String creditCard = finance.creditCard(CreditCardType.UNIONPAY).replace("-", "");
         assertThat(creditCard).hasSize(16);
-        assertThat(creditCard.startsWith("62") || creditCard.startsWith("64") || creditCard.startsWith("65") || creditCard.startsWith("81")).isTrue();
+        assertThat(creditCard.startsWith("62") || creditCard.startsWith("64") || creditCard.startsWith("65") || creditCard.startsWith("81"))
+            .as("UnionPay credit card number begins with the correct digits")
```
kingthorin: Could you concatenation credit card here so the actual generated value is visible?

asolntsev: @Fishbowler Something like this:

```java
assertThat(creditCard.startsWith("62") || creditCard.startsWith("64") || creditCard.startsWith("65") || creditCard.startsWith("81"))
            .as(() -> "UnionPay credit card number begins with the correct digits, but received: " + creditCard)
            .isTrue();
```

Fishbowler: Oooh, that's even better! ❤️

Fishbowler: Fixed

> org.opentest4j.AssertionFailedError: [Expected UnionPay credit card number to begin with the correct digits (62, 64, 65, 81), but received: 1221262098276253] 
Expecting value to be true but was false
Expected :true
Actual   :false

kingthorin: I'd like to propose:
```java
    @RepeatedTest(100)
    void unionpayCard() {
        List<String> startingDigits = List.of("62", "64", "65", "81");
        String creditCard = finance.creditCard(CreditCardType.UNIONPAY).replace("-", "");
        assertThat(creditCard).hasSize(16);
        assertThat(startingDigits.contains(creditCard.subString(0,2))
            .as(() -> "Expected UnionPay credit card number to begin with the correct digits " + startingDigits + ", but received: " + creditCard)
            .isTrue();
    }
```

PR #1654, opened 2025-09-12
Updated references from 2.4.5 to 2.5.0, since that most likely will b…

bodiam:
…ecome the release.


kingthorin:
https://github.com/datafaker-net/datafaker/blob/main/pom.xml#L7

bodiam:
I thought about it, but I didn't think it was worth it right now. This will be addressed during the release, we just won't have 2.5.0-snapshot jars nor doc, I think we can survive.

Issue #1659, opened 2025-09-16
Faker().text().text(1, 64, true, false, true) returns empty String when length is 1


axelfontaine:
**Describe the bug**
`Faker().text().text(1, 64, true, false, true)` returns empty String when length is 1.

**To Reproduce**
```groovy
(1..1000).forEach { _ -> check(Faker().text().text(1, 64, true, false, true).isNotEmpty()) }
```

**Expected behaviour**
A fake text according to the specifications.

**Versions:**
 - JDK: 24
 - Faker Version: 2.5.0

**Additional context**
It happens here: https://github.com/datafaker-net/datafaker/blob/main/src/main/java/net/datafaker/providers/base/Text.java#L206

asolntsev:
@axelfontaine From your side, the mistake is parameter "1". The minimum length cannot be 1 because you expect your password to contain at least two symbols (upper case and digit). So this code would work correctly:

```java
Faker().text().text(2, 64, true, false, true)
```

From DF side, we can make a fix so that expression `Faker().text().text(1, 64, true, false, true)` always generates a string of length 2..64

axelfontaine:
@asolntsev The contract as it stands now is really ambiguous. It is easily interpreted as a random string between this and that length, drawing from a pool of both lowercase letters and digits. From your comment and the code, I understand you meant something else. You may want to clarify that in JavaDoc to avoid this kind of confusion in the future.

erikpragt-connectid:
> From DF side, we can make a fix so that expression Faker().text().text(1, 64, true, false, true) always generates a string of length 2..64

We could do that, but then someone will report a bug that the generated length would never be 1, even though they explicitly said it should be 1...

asolntsev:
> someone will report a bug that the generated length would never be 1, even though they explicitly said it should be 1...

Now it's a question of definition. 
For me, it seems reasonable to read `faker.text().text(1, 64, true, false, true)` as "generate a random string NO SHORTER than 1 and NO LONGER than 64". For me, it's totally OK if it never returns 1-length string.
**Just describing it explicitly in javadoc is good enough.**

Alternative is to throw IllegalArgumentException (because 1 is not possible with (true, false, true) parameters). But it's a breaking change, and it may cause headache for many users.

axelfontaine:
Once the contract is made explicit in JavaDoc, rejecting invalid input and not producing invalid output seem perfectly fine to me.

kingthorin:
[...]

PR #1660, opened 2025-09-16
#1659 fix method `faker.text().text(1, 64, true, false, true)`

asolntsev:
... to return a valid string of length 2..64 instead of empty string.

NB! This PR contains a breaking change:

```java
    lorem.characters(-1) // before: returned "", now: throws exception
    lorem.characters(-1, true) // before: returned "", now: throws exception
    lorem.characters(-1, true, true, true) // before: returned "", now: throws exception
```

### DIFF
--- a/src/test/java/net/datafaker/idnumbers/ItalianIdNumberTest.java
+++ b/src/test/java/net/datafaker/idnumbers/ItalianIdNumberTest.java
@@ -53,7 +53,7 @@ class ItalianIdNumberTest {
             .isEqualTo('R');
     }
 
-    @RepeatedTest(1000)
+    @RepeatedTest(100)
     void checksumShouldMatchForValidCodes() {
         PersonIdNumber personIdNumber = impl.generateValid(new Faker(LOCALE), new IdNumberRequest(1, 200, ANY));
         String idNumber = personIdNumber.idNumber();
--- a/src/test/java/net/datafaker/providers/base/TextTest.java
+++ b/src/test/java/net/datafaker/providers/base/TextTest.java
@@ -100,19 +100,19 @@ class TextTest {
         }
     }
 
-    @Test
+    @RepeatedTest(10)
     void testCharacter() {
         Character character = faker.text().character();
         assertThat(character.toString()).matches(characterPattern);
     }
 
-    @RepeatedTest((100))
+    @RepeatedTest(10)
     void testUppercaseCharacter() {
         Character character = faker.text().uppercaseCharacter();
         assertThat(character).isUpperCase();
     }
 
-    @RepeatedTest((100))
+    @RepeatedTest(10)
     void testLowercaseCharacter() {
         Character character = faker.text().lowercaseCharacter();
         assertThat(character).isLowerCase();
@@ -120,7 +120,7 @@ class TextTest {
 
     @Test
     void testFixedLengthText() {
-        for (int i = 0; i < 100; i++) {
+        for (int i = 0; i < 10; i++) {
             String text = faker.text().text(i);
             assertThat(text).hasSize(i).matches("[a-z]*");
         }


src/main/java/net/datafaker/providers/base/Text.java
```diff
@@ -66,25 +66,38 @@ public String text(int minimumLength, int maximumLength, boolean includeUppercas
         return text(minimumLength, maximumLength, includeUppercase, includeSpecial, false);
     }
 
-    public String text(int minimumLength, int maximumLength, boolean includeUppercase, boolean includeSpecial, boolean includeDigit) {
+    public String text(int requestedMinimumLength, int maximumLength, boolean includeUppercase, boolean includeSpecial, boolean includeDigit) {
+        if (requestedMinimumLength > maximumLength) {
+            throw new IllegalArgumentException("Min length (%s) should be not greater than max length (%s)".formatted(
```
kingthorin: "...should not be..."

src/main/java/net/datafaker/providers/base/Text.java
```diff
@@ -66,25 +66,38 @@ public String text(int minimumLength, int maximumLength, boolean includeUppercas
...
+                requestedMinimumLength, maximumLength));
+        }
+
+        final int minimumLength = Math.max(requestedMinimumLength, min(includeUppercase) + min(includeSpecial) + min(includeDigit));
+        if (minimumLength > maximumLength) {
+            throw new IllegalArgumentException("Minimum number of required characters (%s) should be not greater than max length (%s)".formatted(
```
kingthorin: "...should not be..."

src/main/java/net/datafaker/providers/base/Text.java
asolntsev: thanks, fixed.

src/main/java/net/datafaker/providers/base/Text.java
asolntsev: fixed.

erikpragt-connectid:
I'm not sure about this change. If we implement this, suddenly uppercase/lowercase becomes more important than the String length, since when you say: `1..64`, it will never be one. 

Also, this change would be a breaking change without people knowing it. I'm not a huge fan.

Would it be possible to have some validation logic on it? Like if you do:

```
faker.text().text(1, 64, true, false, true)
```

It will throw an exception, since this configuration can never work? (note, this would also be a breaking runtime change without clear communication to our users), some code might suddenly start failing. 

Perhaps we should keep it as in, and instead make a (validating) text builder or so? `faker.text().minLength(1).maxLength(60).withUppercase().withLowercase()`

asolntsev:
> some code might suddenly start failing

@erikpragt-connectid 
No, nothing will fail. Let me clearly emphasize: the line `faker.text().text(1, 64, true, false, true)` did not fail, and will not fail.
What actually will change is:
1. BEFORE: sometimes it returned an empty string - which clearly breaks the requirement "from 1 to 64".
2. AFTER: it will always return a valid string of length 2...64 - which is totally valid response to the requirement "from 1 to 64".

asolntsev:
@erikpragt-connectid Added javadoc:
```java
     * @param minimumLength The generated text will not be shorter than this length
     * @param maximumLength The generated text will not be longer than this length
```

bodiam:
> No, nothing will fail.

How so? The behaviour of the method changed, and consumers might be depending on this behaviour in their code. Why do you think they won't fail?

In my understanding:

```
faker.text().text(1, 64, true, false, false)
faker.text().text(1, 64, true, true, false)
faker.text().text(1, 64, true, true, true)
```

Should all return a String of 1..64, wouldn't that make a lot of sense? 
Not 1..64, 2..64, 3..64, just because you might also allow numbers and lower case. 
I would assume that the flags indicate that the string MAY contain one of the characters, not that it MUST contain. 

I wouldn't know what this would do though:

```
faker.text().text(1, 64, false, false, false)
```

asolntsev:
@bodiam No, this method works differently.  :)

1. `faker.text().text(1, 64, false, false, false)` returns 1..64, all lower case Latin letters
2. `faker.text().text(1, 64, true, false, false)` returns 1..64, lower and upper case; at least one upper guaranteed.
3. `faker.text().text(1, 64, false, false, true)` returns 1..64, lower case and digits; at least one digit guaranteed.
4. etc.

This is the logic usually needed for password generation.

src/main/java/net/datafaker/providers/base/Text.java
```diff
@@ -66,25 +66,50 @@ public String text(int minimumLength, int maximumLength, boolean includeUppercas
...
+     * For example, expression {@code text(5, 15, true, false, true} generates a random string of length 5 to 15
+     * containing at least one upper case letter and at least one digit.
+     *
+     * @param minimumLength The generated text will not be shorter than this length
+     * @param maximumLength The generated text will not be longer than this length
+     * @param includeUppercase is at least one uppercase letter required
+     * @param includeSpecial  is at least one special symbol required
+     * @param includeDigit  is at least one digit required
+     * @return a random string withing given length interval
+     * @see Text#DEFAULT_SPECIAL
```
snuyanzin, line 79: seems need to add a comment that it could throw exception

asolntsev: Thanks, added `@throws` javadoc.

PR #1668, opened 2025-09-23
Add ability to use whitelist conditions

snuyanzin:
Sometimes especially for prod systems it is required to have kind of limitations like whitelist or blacklist for providers.

Instead of having both, just add predicate where condition could be supplied

without predicate it behaves as before

### DIFF
--- a/src/main/java/net/datafaker/providers/base/BaseFaker.java
+++ b/src/main/java/net/datafaker/providers/base/BaseFaker.java
@@ -29,6 +29,7 @@ import java.util.function.Supplier;
  * @author ren
  */
 public class BaseFaker implements BaseProviders {
+    private static final Predicate<Class<?>> EVERY_PROVIDER_ALLOWED = t -> true;
     private final FakerContext context;
     private final FakeValuesService fakeValuesService;
     private final Map<Class<?>, AbstractProvider<?>> providersCache = new IdentityHashMap<>();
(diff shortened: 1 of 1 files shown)


asolntsev:
To be honest, I don't understand why it's needed.

Do I correctly understand that
1. `new Faker.expression("#{Name.fullName}")` return a random name
2. `Faker(..., c -> c != Name.class).expression("#{Name.fullName}")` throws an exception

But if you don't need names, just don't call `*.expression("#{Name.fullName}")`... What's the problem?

src/main/java/net/datafaker/providers/base/BaseFaker.java
```diff
@@ -53,8 +55,13 @@ public BaseFaker(Locale locale, RandomService randomService) {
     }
 
     public BaseFaker(FakeValuesService fakeValuesService, FakerContext context) {
+        this(fakeValuesService, context, null);
```
asolntsev: A minor thing: maybe instead of null, it's better to pass a Predicate that always returns true? (to avoid potential NPEs and simplify the logic)

snuyanzin:
>Do I correctly understand that
    new Faker.expression("#{Name.fullName}") return a random name
    Faker(..., c -> c != Name.class).expression("#{Name.fullName}") throws an exception

yes

>But if you don't need names, just don't call *.expression("#{Name.fullName}")... What's the problem?

it works if you control the input

however if input comes from a third party then you can not say just don't call smth... or even if you say you can not be sure in it

PR #1671, opened 2025-09-24
Stop using Internet#username() interally

kingthorin:
Since Internet#usernam() was already changed to use Credentials#username() all internal use may as well do the same thing so that DF itself isn't using deprecated methods.


asolntsev (approved):

kingthorin:
Thanks @asolntsev !

PR #1681, opened 2025-09-26
Personal Identification Number provider for Norway (NO)

asolntsev:
This PR allows to generate a random national ID number for Norway:

```java
Faker faker = new Faker(new Locale("no", "NO"));
String pin1 = faker.idNumber().valid();
PersonIdNumber pin2 = faker.idNumber().valid(new IdNumberRequest(21, 65, FEMALE));
String pin3 = faker.idNumber().invalid();
```

### DIFF
--- a/src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
+++ b/src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
@@ -7,11 +7,7 @@ import net.datafaker.providers.base.PersonIdNumber.Gender;
 
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Stream;
 
-import static java.lang.Integer.parseInt;
 import static net.datafaker.idnumbers.Utils.birthday;
 import static net.datafaker.idnumbers.Utils.gender;
 
@@ -29,21 +25,13 @@ import static net.datafaker.idnumbers.Utils.gender;
  * <p>
  *     Example: 11077941012
  * </p>
+ *
  * @see <a href="https://www.skatteetaten.no/en/person/national-registry/identitetsnummer-og-elektronisk-id/fodselsnummer/">Overview</a>
  * @see <a href="https://taxid.pro/">Online generator</a>
  * @see <a href="https://www.oecd.org/content/dam/oecd/en/topics/policy-issue-focus/aeoi/norway-tin.pdf">More on centuries</a>
  */
 public class NorwegianIdNumber implements IdNumberGenerator {
     private static final DateTimeFormatter BIRTHDAY_FORMAT = DateTimeFormatter.ofPattern("ddMMyy");
-    private static final List<Map.Entry<Interval, Interval>> CENTURIES = Stream.of(
-            "1854-1899, 74-50",
-            "1940-1999, 99-90",
-            "1900-1999, 49-0",
-            "2000-2039, 99-50"
-        )
-        .map(s -> s.split(", "))
-        .map(array -> Map.entry(Interval.of(array[0]), Interval.of(array[1]).reverse()))
-        .toList();
 
     private static final int[] CHECKSUM_COEFFICIENTS_K1 = {3, 7, 6, 1, 8, 9, 4, 5, 2};
     private static final int[] CHECKSUM_COEFFICIENTS_K2 = {5, 4, 3, 2, 7, 6, 5, 4, 3, 2};
@@ -73,20 +61,41 @@ public class NorwegianIdNumber implements IdNumberGenerator {
 
     String basePart(BaseProviders faker, LocalDate birthday, Gender gender) {
         String birthdayDigits = BIRTHDAY_FORMAT.format(birthday);
-
-        int sequenceNumber = CENTURIES
-            .stream()
-            .filter(e -> e.getKey().contains(birthday.getYear()))
-            .map(Map.Entry::getValue)
-            .map(interval -> faker.random().nextInt(interval.from, interval.to))
-            .findFirst()
-            .orElseThrow(() -> new IllegalArgumentException("Birthday is not in range of supported in Norway"));
-
-        int genderDigit = gender == Gender.FEMALE
-            ? faker.options().option(0, 2, 4, 6, 8)
-            : faker.options().option(1, 3, 5, 7, 9);
+        int sequenceNumber = generateSequenceNumber(faker, birthday.getYear());
+        int genderDigit = genderDigit(faker, gender);
 
         return "%s%02d%s".formatted(birthdayDigits, sequenceNumber, genderDigit);
+    }
+
+    /**
+     * <ul>
+     * <li>born 1854-1899: allocated from series 749-500</li>
+     * <li>born 1900-1999: allocated from series 499-000</li>
+     * <li>born 1940-1999: also allocated from series 999-900</li>
+     * <li>born 2000-2039: allocated from series 999-500</li>
+     * </ul>
+     *
+     * @see <a href="https://www.oecd.org/content/dam/oecd/en/topics/policy-issue-focus/aeoi/norway-tin.pdf">TIN Structure</a>
+     */
+    private int generateSequenceNumber(BaseProviders faker, int year) {
+        if (year >= 1854 && year <= 1899) {
+            return faker.random().nextInt(50, 74);
+        } else if (year >= 1940 && year <= 1999) {
+            return faker.random().nextInt(90, 99);
+        } else if (year >= 1900 && year <= 1999) {
+            return faker.random().nextInt(0, 49);
+        } else if (year >= 2000 && year <= 2039) {
+            return faker.random().nextInt(50, 99);
+        } else {
+            throw new IllegalArgumentException("Birthday is not in range of supported in Norway");
+        }
+    }
+
+    private int genderDigit(BaseProviders faker, Gender gender) {
+        return switch (gender) {
+            case FEMALE -> faker.options().option(0, 2, 4, 6, 8);
+            case MALE -> faker.options().option(1, 3, 5, 7, 9);
+        };
     }
 
     /**
@@ -109,18 +118,4 @@ public class NorwegianIdNumber implements IdNumberGenerator {
         return (11 - checkSum % 11) % 10;
     }
 
-    private record Interval(int from, int to) {
-        public static Interval of(String string) {
-            String[] split = string.split("-");
-            return new Interval(parseInt(split[0]), parseInt(split[1]));
-        }
-
-        public boolean contains(int value) {
-            return value >= from && value <= to;
-        }
-
-        public Interval reverse() {
-            return new Interval(to, from);
-        }
-    }
 }


src/test/java/net/datafaker/idnumbers/NorwegianIdNumberTest.java
```diff
@@ -0,0 +1,129 @@
...
+            .hasSize(11)
+            .describedAs("Invalid ID number should consist from numbers only")
+            .matches("\\d+");
+        assertThat(checksum)
+            .describedAs("Checksum should be broken")
+            .isNotEqualTo(validChecksum);
+    }
+
+    //private static class ZOPA extends ArrayList<Integer> implements Collection<Integer>,  {
+    private static class ZOPA implements Future<Integer>, Iterable<String> {
```
pokebadgerswithspoon: 😱

snuyanzin: what is `ZOPA` and why do we need it here?

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
```diff
@@ -0,0 +1,126 @@
...
+    private static final DateTimeFormatter BIRTHDAY_FORMAT = DateTimeFormatter.ofPattern("ddMMyy");
+    private static final List<Map.Entry<Interval, Interval>> CENTURIES = Stream.of(
+            "1854-1899, 74-50",
+            "1940-1999, 99-90",
+            "1900-1999, 49-0",
+            "2000-2039, 99-50"
+        )
+        .map(s -> s.split(", "))
+        .map(array -> Map.entry(Interval.of(array[0]), Interval.of(array[1]).reverse()))
+        .toList();
```
kingthorin: I think this reads a bit more clearly:
```java
record Century(Interval from, Interval to) {}

private static final List<Century> CENTURIES = List.of(
    new Century(Interval.of("1854-1899"), Interval.of("74-50").reverse()),
    new Century(Interval.of("1940-1999"), Interval.of("99-90").reverse()),
    new Century(Interval.of("1900-1999"), Interval.of("49-0").reverse()),
    new Century(Interval.of("2000-2039"), Interval.of("99-50").reverse())
);
```

Ignore the names from and to, you get the idea

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Override
+    public String generateInvalid(BaseProviders faker) {
+        String valid = generateValid(faker);
+        String basePart = valid.substring(0, valid.length() - 2);
+        int invalidChecksum = (checksum(basePart) + 1) % 100;
+        return "%s%02d".formatted(basePart, invalidChecksum);
+    }
+
+    String basePart(BaseProviders faker, LocalDate birthday, Gender gender) {
```
ruzzland, line 62: Why is this method not private?

src/test/java/net/datafaker/idnumbers/NorwegianIdNumberTest.java
asolntsev: Oops, leftovers of debug code... Fixed.

asolntsev: Oops, leftovers of debug code... :)
Fixed.

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
asolntsev: Because it's called from unit-tests.

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
asolntsev: I agree that in pure Java code, your style looks better.

But our initial idea was to keep this description as close as possible to the original table in documentation:

| Period    | Sequence |
| -------- | ------- |
| 1854-1899 | 74x-50x |
| 1940-1999 | 99x-90x |
| 1900-1999 | 49x-0x |
| 2000-2039 | 99x-50x |

(where `x` is even or odd depending on the gender).

Also, the term "Century" is not the best here because "1854-1899" is only a half of a century. :) 
Maybe "Period" or something similar...

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
ruzzland: It's obvious, but why is it called from there? Is it a part of the public interface?

kingthorin: @asolntsev just add some javadoc saying that it's only exposed for testing purposes and isn't part of the public API

For my two cents 🪙

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
```diff
@@ -0,0 +1,126 @@
...
+            .stream()
+            .filter(e -> e.getKey().contains(birthday.getYear()))
+            .map(Map.Entry::getValue)
+            .map(interval -> faker.random().nextInt(interval.from, interval.to))
+            .findFirst()
+            .orElseThrow(() -> new IllegalArgumentException("Birthday is not in range of supported in Norway"));
+
+        int genderDigit = gender == Gender.FEMALE
+            ? faker.options().option(0, 2, 4, 6, 8)
+            : faker.options().option(1, 3, 5, 7, 9);
```
snuyanzin:
```suggestion
        int genderDigit = faker.options().option(0, 2, 4, 6, 8);
        genderDigit = gender == Gender.FEMALE ? genderDigit: genderDigit + 1;
```

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
snuyanzin: should we introduce then `@VisibleForTesting` ?

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
```diff
@@ -0,0 +1,126 @@
...
+            return new Interval(parseInt(split[0]), parseInt(split[1]));
+        }
+
+        public boolean contains(int value) {
+            return value >= from && value <= to;
+        }
+
+        public Interval reverse() {
+            return new Interval(to, from);
+        }
```
snuyanzin: these 2 seems to be leading to confusing results especially if we call like 
```
new Interval(3, 5)
contains(4)
reverse()
contains(4)
```

src/test/java/net/datafaker/idnumbers/NorwegianIdNumberTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.jupiter.params.provider.CsvSource;
+
+import java.time.LocalDate;
+
+import static java.lang.Integer.parseInt;
+import static org.assertj.core.api.Assertions.assertThat;
+
+class NorwegianIdNumberTest {
+    private final NorwegianIdNumber generator = new NorwegianIdNumber();
+    private final Faker faker = new Faker();
```
snuyanzin, line 16: I would expect at least one test for NorweianIdNumber with corresponding locale

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
kingthorin: I think javadoc is simpler and more clear. If an annotation is used then others just have to lookup the meaning of a nonstandard annotation 🤷‍♂️

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
snuyanzin: probably better to have something in constructor like 
```java
public static Interval of(String string) {
  final int[] split = Arrays.stream(string.split("-")).mapToInt(Integer::parseInt).toArray();
  return new Interval(Math.min(split[0], split[1]), Math.max(split[0], split[1]));
}
```

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
asolntsev: @ruzzland @kingthorin @snuyanzin We don't need any javadoc or annotations. This method **is not public**, it's _package-private_. It can be called only from the same package. Meaning, only from tests.

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
asolntsev: Also possible. But IMHO this way is a bit harder to read.

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
snuyanzin: it means that nobody prevents and nothing warns users from having same package in their code base and call it right?
And then if in future because of refactoring we decide to rename/change visibility, remove it the complaining user will be right because there is nothing telling it can not be used and it is for internal use only.

ruzzland: @asolntsev Definitely NOT ONLY from the tests. They are meant to be called from the neighbour classes. They are for the code decomposition, when the whole package incapsulates some logic spread over multiple classes. But marking some of the inherently private methods as package private only for test introduces some kind of inconsistency. Of course IMHO.

pokebadgerswithspoon: > t means that nobody prevents and nothing warns users from having same package in their code

Two observations:

1) If someone creates "his" code in "our" package outside of this repo - he should be prepared to bear the consequences. As he is ~~operating~~ messing guts of "our" package unchecked.
However should he create a PR  in this repo he would be supported on the review.

2) This provider implements `IdNumberGenerator` and that should be source of contract for the user. I think this is what L in SOLID stands for.

There are many ways to misuse the code, I'd say these 2 are on the list.

TLDR: `package-private` is okey, because user is warned by package naming and L of Solid.

snuyanzin: please be aware that this is open source project which implies everyone can use it and we should allow something being used only by a strong reason because to close it back will be a much trickier task

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
asolntsev: Anyway, in the end I simplified this code: converted Map to plain old "if-else-if". :)

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
asolntsev: > It's obvious, but why is it called from there? Is it a part of the public interface?

Grrrr... If it's obvious, why you are asking it? 

@snuyanzin @kingthorin We have a very strong reason for having this method non-public. We need to cover its logic by unit-tests. And it's impossible to cover this logic through the public API because the public API **generates random data**.   

@ruzzland Isn't this obvious?

snuyanzin: I didn't get , what is the issue with putting extra javadoc explaining the reason or annotation?
Or make it private and no need for that, however something should be done with tests

asolntsev: C'mon, it's a widely used pattern to test private methods! 
In DataFaker project, we already have a lot of package-private methods. It has never been a problem. Nobody is every expected to create `new NorwegianIdNumber()`. End-users are expected to call `new Faker().idNumber().valid()` - this is the actual public API, and it doesn't allow calling package-private methods.

@snuyanzin 
Sure, we can add javadoc saying `Private API, allowed to call only from tests.`.
But we have too many such places.
In that sense, the annotation would be better. It's much shorter. 
@snuyanzin Can we use some existing annotation from Guava or "Error prone" library, or should we introduce a custom annotation?

snuyanzin: the initial idea was to have the project lightweight without dependencies (probably except for snakeyaml) so the custom one is preferred then or make sure that thirdparty annotation doesn't appear in runtime

kingthorin: I think we are getting caught on a small detail here, I'm sure there are other examples of this in the project already.

If we want them all addressed I suggest someone open an issue and we come to consensus and address them all.

Let's not let it derail this PR

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
```diff
@@ -0,0 +1,120 @@
...
+     */
+    private int generateSequenceNumber(BaseProviders faker, int year) {
+        if (year >= 1854 && year <= 1899) {
+            return faker.random().nextInt(50, 74);
+        } else if (year >= 1940 && year <= 1999) {
+            return faker.random().nextInt(90, 99);
+        } else if (year >= 1900 && year <= 1999) {
+            return faker.random().nextInt(0, 49);
+        } else if (year >= 2000 && year <= 2039) {
+            return faker.random().nextInt(50, 99);
```
snuyanzin, line 88: can you elaborate why javadoc and the doc (https://www.oecd.org/content/dam/oecd/en/topics/policy-issue-focus/aeoi/norway-tin.pdf) says
>The three digits comprising the individual number are allocated sequentially within the
specific date of birth. Individual digits are allocated as follows

and this method returns 1 or 2 digits?

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
asolntsev: In the end, I've simplified all this map and removed using Intervals at all.

src/test/java/net/datafaker/idnumbers/NorwegianIdNumberTest.java
asolntsev: done

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
asolntsev: I agree that it's a bit tricky to track :(
The point is that the third digit depends on the gender. 
E.g. for year 1888, we pick a random number between `50` and `74`, and append one more digit (the third digit) to it (even for women and odd for men).

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
asolntsev: simplified the code

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
ruzzland: @asolntsev Hmm... For me it sounds strange that "it's **_impossible_** to cover this logic through the public API". But I got the reason for now, thanks.

asolntsev: @kingthorin I've create a ticket: https://github.com/datafaker-net/datafaker/issues/1682
@ruzzland All volunteers are welcome.

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * The Norwegian Individual ID Number is 11 digits log
+ * Format: ddMMyyCCGkk
+ * <ul>
+ *     <li>dd - day of birth</li>
+ *     <li>MM - month of birth</li>
+ *     <li>yy - year of birth</li>
+ *     <li>CC - sequence (range depends on century)</li>
+ *     <li>kk - checksum</li>
+ * </ul>
```
snuyanzin, line 24: `G` - seems to be missed

asolntsev: thanks, fixed.

kingthorin (approved):

pokebadgerswithspoon (approved):
table of requirements is gone, sad this idea did not work, but handwritten if..else are not that scary in the end.
LTGM and approve if that counts

Issue #1682, opened 2025-09-26
Mark non-private methods intended to be called only from tests


asolntsev:
**Is your feature request related to a problem? Please describe.**
DataFaker has multiple package-private (and maybe even public) methods that are not intended to be used by end users. Mostly, they were made non-private only to be called from unit-tests.

**Describe the solution you'd like**
We need to explicitly mark such methods with some annotation like `@VisibleOnlyForTests` or some javadoc.
Maybe it's also a good idea to create a special test (probably using [ArchUnit](https://www.archunit.org/) library) that would verify that all package-private methods are annotated.

**Additional context**
Born in discussion https://github.com/datafaker-net/datafaker/pull/1681#discussion_r2382692009

kingthorin:
I don't think we should introduce a dependency for this. It's always been a big deal to the team that dependencies are kept minimal. To the point that it was added to the contributing guidelines earlier this year. https://github.com/datafaker-net/datafaker/blob/main/CONTRIBUTING.md#dependencies

To me the three practical options are.
- We leave it as-is (no change).
- We add JavaDoc.
- We create and use our own annotation.

As you've pointed out in your PR an annotation isn't terribly effective.

asolntsev:
@kingthorin I vote for option #1: 

Just leave package-private methods as is. :)

kingthorin:
GitHub needs to add polls 😀

kingthorin:
From @bodiam 
> Hi all,I'm not sure what we're trying to solve here, but I dislike every bit of this, and this looks like a solution looking for a problem.Erik On 30 Sep 2025, at 05:35

bodiam:
thanks @kingthorin , I saw that, but I don't see what problem we're trying to solve. There's a package protected method, so what? I see no problem with having them. 

> Datafaker has multiple package-private methods that are not intended to be used by end users. 

How can one call a package private method even? Using reflection or so? The annotations wouldn't stop that either.

kingthorin:
Since no-one else has chimed back in on the topic I'm going to say that's three votes for leave it as-is (and none for any other option).

asolntsev:
> How can one call a package private method even?

Technically, it's not hard at all. Any user can just create a package "net.datafaker.idnumbers" in his project, and he can call any package-private methods. 

See discussion in [comments](https://github.com/datafaker-net/datafaker/pull/1681#discussion_r2383120711)

PR #1683, opened 2025-09-27
#1682 mark as @InternalAPI methods and classes not intended for end-users

asolntsev:
@snuyanzin @ruzzland Is this what you meant?
For me, this biggest question is, if DataFaker users will pay attention to this annotation. It's quite invisible for them. IDE would not show any warnings about it.

### DIFF
--- a/src/test/java/net/datafaker/ArchitectureTest.java
+++ b/src/test/java/net/datafaker/ArchitectureTest.java
@@ -1,0 +1,45 @@
+package net.datafaker;
+
+import com.tngtech.archunit.core.domain.JavaClasses;
+import com.tngtech.archunit.core.importer.ClassFileImporter;
+import com.tngtech.archunit.lang.ArchRule;
+import net.datafaker.annotations.InternalApi;
+import net.datafaker.idnumbers.IdNumberGenerator;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
+import static com.tngtech.archunit.base.DescribedPredicate.not;
+import static com.tngtech.archunit.core.domain.JavaClass.Predicates.type;
+import static com.tngtech.archunit.core.importer.ImportOption.Predefined.DO_NOT_INCLUDE_TESTS;
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.methods;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+
+@TestInstance(PER_CLASS)
+public class ArchitectureTest {
+    /*
+     * 1. Don't hold these in static fields - otherwise they will consume memory till the end of all tests
+     * 2. Use `@TestInstance(PER_CLASS)` to reuse they between tests
+     */
+    private final JavaClasses PRODUCTION_CLASSES = new ClassFileImporter().withImportOption(DO_NOT_INCLUDE_TESTS).importPackages("net.datafaker");
+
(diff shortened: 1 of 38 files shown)


src/test/java/net/datafaker/ArchitectureTest.java
```diff
@@ -0,0 +1,45 @@
...
+     * 1. Don't hold these in static fields - otherwise they will consume memory till the end of all tests
+     * 2. Use `@TestInstance(PER_CLASS)` to reuse they between tests
+     */
+    private final JavaClasses PRODUCTION_CLASSES = new ClassFileImporter().withImportOption(DO_NOT_INCLUDE_TESTS).importPackages("net.datafaker");
+
+    @Test
+    public void packagePrivateMethodsShouldBeMarkedAsInternalAPI() {
+        ArchRule rule = methods()
+            .that().arePackagePrivate()
+            .should().beAnnotatedWith(InternalApi.class);
```
pokebadgerswithspoon, line 30: It was not a question to me, but still I feel urge to share that it might be an overkill. Apologies 😎 

1) now we do this to enforce users to treat `package-private` as a pattern to narrow down a unit to test
2) then (I'm speculating now) someone will want a true `package-private` feature and he'll need to introduce a new annotation `@PackagePrivateApi`
3) a new contributor comes and gets scared away

So, I'd think this steeps up learning curve for a newcomer. 
Even if implemented the code still could be misused,
- previously one would misuse API and package naming
- after this PR one would misuse annotations

asolntsev: Totally agree. 
So it's easier to just leave package-private methods as is?

ruzzland: > Is that what you meant?

No, my point is if you use _package private_ solely for testing, it's probably a sign of not the very best overall design. It is IMHO, I admit it may be a moot point and if it's a widely used pattern as you say (at least in this project) I suggest not introducing any annotations nor using javadoc at all. Apologies for the distraction.

bodiam:
Hi all,I'm not sure what we're trying to solve here, but I dislike every bit of this, and this looks like a solution looking for a problem.Erik On 30 Sep 2025, at 05:35, Ruslan ***@***.***> wrote:﻿
@ruzzland commented on this pull request.



In src/test/java/net/datafaker/ArchitectureTest.java:
> +import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.methods;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+
***@***.***(PER_CLASS)
+public class ArchitectureTest {
+    /*
+     * 1. Don't hold these in static fields - otherwise they will consume memory till the end of all tests
+     * 2. Use ***@***.***(PER_CLASS)` to reuse they between tests
+     */
+    private final JavaClasses PRODUCTION_CLASSES = new ClassFileImporter().withImportOption(DO_NOT_INCLUDE_TESTS).importPackages("net.datafaker");
+
+    @Test
+    public void packagePrivateMethodsShouldBeMarkedAsInternalAPI() {
+        ArchRule rule = methods()
+            .that().arePackagePrivate()
+            .should().beAnnotatedWith(InternalApi.class);


Is that what you meant?

No, my point is if you use package private solely for testing, it's probably a sign of not the very best overall design. It is IMHO, I admit it may be a moot point and if it's a widely used pattern as you say (at least in this project) I suggest not introducing any annotations nor using javadoc at all. Apologies for the distraction.

kingthorin:
I believe this can be closed.

https://github.com/datafaker-net/datafaker/issues/1682

PR #1684, opened 2025-09-27
Error prone

asolntsev:


src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
```diff
@@ -102,6 +107,10 @@ private int genderDigit(BaseProviders faker, Gender gender) {
      * k1 = 11 - ((3 × d1 + 7 × d2 + 6 × m1 + 1 × m2 + 8 × å1 + 9 × å2 + 4 × i1 + 5 × i2 + 2 × i3) mod 11),
      * k2 = 11 - ((5 × d1 + 4 × d2 + 3 × m1 + 2 × m2 + 7 × å1 + 6 × å2 + 5 × i1 + 4 × i2 + 3 × i3 + 2 × k1) mod 11).
      */
+    @RestrictedApi(
```
bodiam, line 110: Why is this stuff here? It's a package private method, there's no way this can be called by anyone.

asolntsev: Because in our previous discussion, there was [comment](https://github.com/datafaker-net/datafaker/pull/1681#discussion_r2383120711) a'la "if someone creates package 'net.datafaker.idnumbers' in his project then he can call these methods".

**I don't like this solution either,** I just played with different ways how it's potentially possible to mark methods "not intended for use".

kingthorin:
I believe this can be closed.

https://github.com/datafaker-net/datafaker/issues/1682

src/main/java/net/datafaker/idnumbers/NorwegianIdNumber.java
bodiam: Let's assume that people who want to use `net.datafaker` as their package have a really good reason to do so :)

Issue #1685, opened 2025-09-27
2.5.1 Outstanding tasks


kingthorin:
- [ ] Also, it might be good to update this: https://www.datafaker.net/releases/2.5.1-SNAPSHOT/ (it's the mkdocs page)

_Originally posted by @erikpragt-connectid in https://github.com/datafaker-net/datafaker/issues/1649#issuecomment-3332345441_

- [ ] POM update workflow needs to adjust current and future versions more specifically 

_Also mentioned in #1649_

asolntsev:
The task can be considered done: release 2.5.2 was automatically made without any problems.

Issue #1649, opened 2025-09-11
Datafaker 2.5.0 Release


bodiam:
Hi all,

My proposal is to release Datafaker 2.5.0 on September 16, 2025, for no other reason than that it is the planned release data of Java 25 as well. 

If you want to get anything in, you still have a few days, but release incoming so to say!

Cheers,
Erik

kingthorin:
I’ve updated the milestone that had been created as 2.4.5 ➡️ 2.5.0 https://github.com/datafaker-net/datafaker/milestone/3

bodiam:
Thanks @kingthorin. We might have make a small change to some providers, maybe the Credentials one, to change the `since` field to 2.5.0 instead of 2.4.5, if it even has a since right now. (I checked, it has one, and it's 2.4.5 at this moment)

Another link here:
* https://github.com/datafaker-net/datafaker/blob/2bb19f9c55954872f321cedc1acdc9e21d61ed07/src/main/java/net/datafaker/providers/base/Internet.java#L44
* https://github.com/datafaker-net/datafaker/blob/2bb19f9c55954872f321cedc1acdc9e21d61ed07/src/main/java/net/datafaker/providers/base/Credentials.java#L9
* https://github.com/datafaker-net/datafaker/blob/2bb19f9c55954872f321cedc1acdc9e21d61ed07/src/main/java/net/datafaker/providers/base/Name.java#L154
* https://github.com/datafaker-net/datafaker/blob/2bb19f9c55954872f321cedc1acdc9e21d61ed07/src/main/java/net/datafaker/providers/base/BaseProviders.java#L157
* https://github.com/datafaker-net/datafaker/blob/2bb19f9c55954872f321cedc1acdc9e21d61ed07/src/main/java/net/datafaker/providers/base/BaseProviders.java#L166

kingthorin:
👍 want me to tackle those?
(https://github.com/search?q=repo%3Adatafaker-net%2Fdatafaker+%222.4.5%22+-language%3ASVG&type=code)

kingthorin:
Looks like you're on it 😀 https://github.com/datafaker-net/datafaker/pull/1654

bodiam:
<img width="1536" height="1024" alt="Image" src="https://github.com/user-attachments/assets/fb57251c-046e-45d5-9fe4-8e07ee17f1be" />

https://x.com/epragt/status/1967957135024853386

[...]

PR #1686, opened 2025-09-28
Tweak post-release wf to handle version identifiers more specifically

kingthorin:
This should address concerns raised in #1649/#1672 and one of the tasks in #1685.

@asolntsev I'm requesting your review because you're the one that ended up tweaking/fixing the PR for 2.5.1.


.github/workflows/pom_update.yml
```diff
@@ -26,12 +26,20 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
         git checkout -b $BRANCH
+        # Get the current version
+        CURR_VERS=$(./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout | sed 's/-SNAPSHOT$//')
```
asolntsev: Maybe use the normal full name `CURR_VERSION` instead of `CURR_VERS`?

.github/workflows/pom_update.yml
```diff
@@ -26,12 +26,20 @@ jobs:
...
+        echo $CURR_VERS
+        # Increment the version
         ./mvnw build-helper:parse-version versions:set -DnewVersion=\${parsedVersion.majorVersion}.\${parsedVersion.minorVersion}.\${parsedVersion.nextIncrementalVersion}-SNAPSHOT
         git add pom.xml
-        VERS=$(./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout|tr -d SNAPSHOT|tr -d "-")
-        # Replace version in README.md look for >, :, ` (greater than, colon, backtick) followed by 2 period number period number and replace it with the leading char and new version
-        sed -i -E 's/([\>`\:]{1})2\.[0-9]\.[0-9]/\1'"$VERS"'/g' README.md
+        # Get the new version
+        VERS=$(./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout | sed 's/-SNAPSHOT$//')
```
asolntsev: Do I correctly understand that `$VERS` is always equal to `$CURR_VERS`? 
Why do we need two variables then?

.github/workflows/pom_update.yml
```diff
@@ -26,12 +26,20 @@ jobs:
...
+        echo $VERS
+        # Replace version in README.md look for >, : (greater than, colon) followed by "2" period number period number (ex: 2.4.2) and replace it with the same leading char and current version
+        sed -i -E 's/([\>\:]{1})2\.[0-9]\.[0-9]/\1'"$CURR_VERS"'/g' README.md
```
asolntsev: It's hard for me to understand what this line should do.
But it fails on my Mac:

```
> sed -i -E 's/([\>\:]{1})2\.[0-9]\.[0-9]/\1'"$CURR_VERS"'/g' README.md
sed: 1: "s/([\>\:]{1})2\.[0-9]\. ...": \1 not defined in the RE
```

And README doesn't contains combination `>:`.

asolntsev (changes requested):

.github/workflows/pom_update.yml
kingthorin: No VERS is set after it's been incremented.

.github/workflows/pom_update.yml
kingthorin: sed behaves differently on Mac (bsd).

I ran into that the other day too. You need to `brew install gnu-sed` then use `gsed`.

That bit of the regex is: one item from the character set colon or greater than.

.github/workflows/pom_update.yml
kingthorin: I'll rename them both more clearly, there's really no need to keep them short.

kingthorin: Done

bodiam: > "full name" CURR_VERSION. 

Made me smile. How about calling it `CURRENT_VERSION`, since that's what it is?

.github/workflows/pom_update.yml
bodiam: how about calling it `NEW_VERSION` ?

kingthorin: I committed a change after asolntsev's comment, it's currently `NEXT_VERSION`. You good with that?

bodiam: yeah, perfect!

bodiam: (but the comment says: get the new version, instead of get the next version....)

.github/workflows/pom_update.yml
kingthorin: Same, that's also been changed already

kingthorin:
https://www.githubstatus.com/incidents/071h21gptcp0
> Degraded Performance for GitHub Actions MacOS Runners

kingthorin:
Sure, will adjust that ~~later this evening.~~ Monday.

.github/workflows/pom_update.yml
bodiam: Sorry, this GitHub thing is new for me :) 

(I think I didn't see the outdated part)

bodiam:
> The script is working, all right! :) The only suggestion is to change commit message: instead of generic "Update README and pom versions", let's write the specific version number:
> 
> > Release DataFaker $RELEASED_VERSION

Datafaker, not DataFaker!

asolntsev (approved):

kingthorin:
I did fix the name before merging :wink:

asolntsev:
> Datafaker, not DataFaker!

May I ask why Datafaker, not DataFaker?
I was thinking the opposite: something that does fake date is "data faker", isn't it?

bodiam:
> May I ask why Datafaker, not DataFaker?
> I was thinking the opposite: something that does fake date is "data faker", isn't it?

I don't think there's a great reason for it. I can absolutely see your point, but I think I flipped a coin, maybe I looked at both, and thought Datafaker looks more normal casing than DataFaker. It's a bit strange to have the F uppercased. 

But more importantly, and why I'm objecting to DataFaker vs Datafaker, is that all our documentation is using "Datafaker", including every link here: https://www.datafaker.net/in-the-media/links/, so either way, it's a bit hard to change.

Maybe next framework I'll call DataFaker ;)

bodiam:
PS: `updts` ??? What's this, 1980 where we can use max 8 characters? At least be consistent and call it `pdts` if you want to remove all vowels.....

PS2: nice work on this! We should release a 2.5.2 just to test this workflow!

kingthorin:
Hahahaha, I dunno.... just a weird habit.

asolntsev:
> We should release a 2.5.2 just to test this workflow!

Why not. We have Norway ID number and the fix with no-argument constructors in 2.5.2:
https://github.com/datafaker-net/datafaker/milestone/5?closed=1

asolntsev:
@bodiam I've tried to release 2.5.2 - seems that everything worked well. 
The automatically generated PR looks good: https://github.com/datafaker-net/datafaker/pull/1696/files

PR #1690, opened 2025-10-01
Retrieve the no-argument constructor

### LINKED ISSUE #1689: The JavaObjectTransformer sometimes finds a constructor other than the no parameter one
**Describe the bug**
When calling the apply() method of the JavaObjectTransformer on a class that has multiple constructors, including a no-argument one. Sometimes, when attempting to locate the no-argument constructor, it finds one of the others instead.

Here is a stack trace illustrating the exception that occurs:

`java.lang.IllegalArgumentException: wrong number of arguments: 0 expected: 1
	at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:59)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486)
	at net.datafaker.transformations.JavaObjectTransformer.apply(JavaObjectTransformer.java:61)`

As can be seen, the problem occurs in line 61 of the JavaObjectTransformer. It calls `c.getDeclaredConstructors()[0]`, but there is no guarantee that the 0th array element is the no-argument constructor.

**To Reproduce**
To reproduce, create a class with a no-argument constructor as well as a second constructor that takes an argument.

Then create a JavaObjectTransformer and attempt to apply a schema to the class you have created. Sometimes it will work fine, other times it will throw the exception.

**Expected behavior**
What should happen is the no-argument constructor should always be used in this section of the code in JavaObjectTransformer.

**Versions:**
 - OS: Windows 11
 - JDK: Oracle Java 21
 - Faker Version 2.5.1

**Additional context**

The Java documentation is clear that the `c.getDeclaredConstructors()[0]` will not sort the array and can be returned in any order

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Class.html#getDeclaredConstructors()

mt657-jpg:
…using a method called `getParameterlessPublicConstructor`.

This fixes #1689 where the no-argument constructor is not always the one you get

### DIFF
--- a/src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
+++ b/src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
@@ -57,8 +57,12 @@ public class JavaObjectTransformer implements Transformer<Object, Object> {
         } else {
             if (result == null) {
                 try {
-                    Constructor<?> primaryConstructor = CLASS2CONSTRUCTOR.computeIfAbsent(clazz, c -> c.getDeclaredConstructors()[0]);
-                    result = primaryConstructor.newInstance();
+                    Constructor<?> primaryConstructor = CLASS2CONSTRUCTOR.computeIfAbsent(clazz, c -> getParameterlessPublicConstructor(c).orElse(null));
+                    if (primaryConstructor != null) {
+                        result = primaryConstructor.newInstance();
+                    } else {
+                        throw new RuntimeException("Failed to local a parameterless constructor for class " + clazz.getName());
+                    }
                 } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
                     throw new RuntimeException(e);
                 }
@@ -167,11 +171,12 @@ public class JavaObjectTransformer implements Transformer<Object, Object> {
     }
 
     private boolean hasParameterlessPublicConstructor(Class<?> clazz) {
-        for (Constructor<?> constructor : clazz.getConstructors()) {
-            if (constructor.getParameterCount() == 0) {
-                return true;
-            }
-        }
-        return false;
+        return getParameterlessPublicConstructor(clazz).isPresent();
+    }
+
+    private Optional<Constructor<?>> getParameterlessPublicConstructor(Class<?> clazz) {
+        return Arrays.stream(clazz.getConstructors())
+            .filter(constructor -> constructor.getParameterCount() == 0)
+            .findFirst();
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
```diff
@@ -57,8 +57,12 @@ public Object apply(Object input, Schema<Object, ?> schema) {
         } else {
             if (result == null) {
                 try {
-                    Constructor<?> primaryConstructor = CLASS2CONSTRUCTOR.computeIfAbsent(clazz, c -> c.getDeclaredConstructors()[0]);
-                    result = primaryConstructor.newInstance();
+                    Constructor<?> primaryConstructor = CLASS2CONSTRUCTOR.computeIfAbsent(clazz, c -> getParameterlessPublicConstructor(c).orElse(null));
+                    if (primaryConstructor != null) {
+                        result = primaryConstructor.newInstance();
+                    } else {
+                        throw new RuntimeException("Failed to local a parameterless constructor for class " + clazz.getName());
```
kingthorin:
```suggestion
                        throw new RuntimeException("Failed to locate a parameterless public constructor for class " + clazz.getName());
```

src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
```diff
@@ -167,11 +171,12 @@ private Object getObject(Schema<Object, ?> schema, Object result, Constructor<?>
...
-            if (constructor.getParameterCount() == 0) {
-                return true;
-            }
-        }
-        return false;
+        return getParameterlessPublicConstructor(clazz).isPresent();
+    }
+
+    private Optional<Constructor<?>> getParameterlessPublicConstructor(Class<?> clazz) {
+        return Arrays.stream(clazz.getConstructors())
```
kingthorin, line 183: Does clazz.getConstructors() restrict to public? If not then there's a condition missing to conform to the function name

asolntsev: 1. `clazz.getDeclaredConstructors()` returns ALL constructors
2. `clazz.getConstructors()` returns only PUBLIC constructors.

bodiam:
Would it be possible to add a few testcases which demonstrate some scenarios? (like only a no arg (private|package|public) constructor, only a constructor with parameters, a class with both a no-arg and parameter constructor, etc?)

mt657-jpg:
I have added some tests and filtered for public constructors only

bodiam:
Looks great. If you're happy with it, happy to merge it! Thanks for your work on this!

asolntsev (approved):

kingthorin (approved):

PR #1694, opened 2025-10-09
improve image asserts

asolntsev:
in addition to just "data:*base64" prefix, also

1. check image size
2. verify that the base64 part is a valid parseable  image data
3. validate XML format in case of SVG images

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Image.java
+++ b/src/main/java/net/datafaker/providers/base/Image.java
@@ -2,8 +2,8 @@ package net.datafaker.providers.base;
 
 
 import javax.imageio.ImageIO;
-import java.awt.*;
 import java.awt.Color;
+import java.awt.Graphics2D;
 import java.awt.image.BufferedImage;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
--- a/src/test/java/net/datafaker/assertions/ImageAssert.java
+++ b/src/test/java/net/datafaker/assertions/ImageAssert.java
@@ -26,6 +26,10 @@ public class ImageAssert extends AbstractAssert<ImageAssert, ImageData> {
         super(actual, ImageAssert.class);
     }
 
+    public ImageAssert is(ImageType type) {
+        return is(type, 256, 256);
+    }
+
     public ImageAssert is(ImageType type, int width, int height) {
         assertImageDataUrl(type);
         return switch (type) {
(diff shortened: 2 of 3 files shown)


src/main/java/net/datafaker/providers/base/Image.java
```diff
@@ -2,14 +2,15 @@
 
 
 import javax.imageio.ImageIO;
-import java.awt.Color;
 import java.awt.*;
+import java.awt.Color;
```
kingthorin, line 5: I'm not sure why we have one specific import for colour and one for the entire package. Could you do away with the wildcard and use all specific imports?

src/test/java/net/datafaker/assertions/ImageAssert.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+        return this;
+    }
+
+    private ImageAssert verifySvg(ImageData image, int width, int height) {
+        String content = RE.matcher(image.url()).replaceFirst("$1");
+        byte[] bytes = Base64.getDecoder().decode(content);
+        try {
+            String xml = new String(bytes);
+            assertThat(xml).matches("<svg xmlns=\".+\" width=\"256\" height=\"256\">.*");
```
kingthorin: Ends with a closing svg tag?

kingthorin: Or will the assert at 76 catch this if there isn't a closing tag?

src/test/java/net/datafaker/providers/base/ImageTest.java
```diff
@@ -6,83 +6,85 @@
...
     @Test
     void svg() {
-        assertThat(faker.image().base64SVG()).startsWith("data:image/svg+xml;base64,");
+        assertThatImage(faker.image().base64SVG()).is(SVG, 256, 256);
     }
 
     @Test
     void tiff() {
-        assertThat(faker.image().base64TIFF()).startsWith("data:image/tiff;base64,");
+        assertThatImage(faker.image().base64TIFF()).is(TIFF, 256, 256);
```
kingthorin: Since these are all using 256x256 make a method that just takes the image type as a parameter?

src/test/java/net/datafaker/assertions/ImageAssert.java
asolntsev: good point, added `</svg>` (though - yes, 76 line also checks the same thing).

src/test/java/net/datafaker/providers/base/ImageTest.java
asolntsev: thanks. Created method `is(type)` which by default assumes 256x256.

erikpragt-connectid: Wouldn't it be better to use a custom assert here? Like:

```
assertThat(image).is(TIFF, 256, 256); 
```

See https://joel-costigliola.github.io/assertj/assertj-core-custom-assertions.html

asolntsev: Yes, and this is the custom assert :)
Method `assertThatImage` returns the custom assert.

PR #1698, opened 2025-10-11
Add an ice-cream specification

sergejsvisockis:
Contains different sorts of ice cream properties.


asolntsev:
@sergejsvisockis DataFaker 2.5.3 has been released. Thank you for your contributions!

PR #1708, opened 2025-10-27
Enhance Commerce Provider testing with Ukrainian locale support

vicky-iv:
- Added CommerceLocaleTest.java to provide testing for Ukrainian commerce data
- Added individual validation for product name parts:
  - Adjective component (commerce.product_name.adjective)
  - Material component (commerce.product_name.material)
  - Product component (commerce.product_name.product)

### DIFF
--- a/src/test/java/net/datafaker/providers/base/CommerceLocaleTest.java
+++ b/src/test/java/net/datafaker/providers/base/CommerceLocaleTest.java
@@ -13,12 +13,10 @@ public class CommerceLocaleTest extends BaseFakerLocaleTest {
 
     @Override
     protected Stream<Arguments> localeProviderListTest() {
-
         return Stream.of(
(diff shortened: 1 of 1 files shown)


asolntsev (approved):

kingthorin:
Splitting on space has lead to failures:
https://github.com/datafaker-net/datafaker/actions/runs/18859794216/job/53815604879

vicky-iv:
I see - this is because of
```
product_name:
    adjective:
        - "Heavy Duty"
```
It's the only parameter with a space in the middle. There could be more cases like this in the future, so the proposed approach definitely isn't going to work. 
I could at least remove these changes from `CommerceTest` - what do you think?

kingthorin:
Ya unless you can think of another alternative

PR #1710, opened 2025-10-28
Fix test for commerce product name

vicky-iv:
Fix for #1708 
@kingthorin I had an idea that I implemented in this PR. The implementation turned out a bit bulky, but it should now be more flexible.


asolntsev:
@vicky-iv @kingthorin  I've submitted an alternative solution with test locale: https://github.com/datafaker-net/datafaker/pull/1711

Maybe it's simpler than logic with spaces, and better than using Mockito.

asolntsev:
Implemented in https://github.com/datafaker-net/datafaker/pull/1711

PR #1711, opened 2025-10-28
#1708 fix test failing because some product might have space in name

asolntsev:
Alternative fix for https://github.com/datafaker-net/datafaker/pull/1708

@vicky-iv


kingthorin (approved):
I think this does what we need. Thanks!

vicky-iv:
@asolntsev That looks much better than my bulky attempt - thank you!

Issue #1707, opened 2025-10-26
Release versions


bodiam:
Hi all, 

I was trying to use Semver (https://semver.org/) for releases. We (well, I.....) haven't been always accurate with this, but the general thinking was:

MAJOR (x.0.0) -> really big changes, like a major set of API breakage, or having Java 25 as a minimum or so.
MINOR (0.x.0) -> most features. Every time we release a new version, ideally this should be the one, where we introduce new features and users can safely upgrade.
PATCH (0.0.x) -> version when you make backward compatible bug fixes

If there's 1 new feature and 6 bugfixes, is this a bugfix or a feature release? I don't know. 
If there's 1 hardly used method which we remove after having communicated that for a few releases, which breaks the API, is that a MAJOR release? I don't know.

But I just saw the release of 2.5.3 (thank you for that!), which actually only has new features, so to me, this would be a MINOR release, not a bugfix release, so I think the version should have been 2.6.0, not 2.5.3 (yes, I know it's a minor release, but that's exactly why I think 2.6.0 would be the right version).

Either way, no big deal, but just wanted to share my thinking here and ideally get some feedback/thoughts on this

Cheers!
Erik

asolntsev:
I agree, this should be 2.6.0 because it contains new features. :)

In my head, I was rather thinking that "it doesn't break anything, so it's minor". But yes, this is not semver. :)

kingthorin:
I honestly haven't been thinking of it much for datafaker. But yeah I guess the last one should have been 2.6.0.

Shall we always open an issue before doing a release so that version number(s) are agreed upon?
(Or do you want to enable Discussions for the repo, and it can be done there?)

bodiam:
I don't mind that much really. I think we're all aligned in theory, and if we sometimes deviate from that, no big deal for me. Let's not have a meeting about the right version number :)On 28 Oct 2025, at 08:10, Rick M ***@***.***> wrote:﻿kingthorin left a comment (datafaker-net/datafaker#1707)
I honestly haven't been thinking of it much for datafaker. But yeah I guess the last one should have been 2.6.0.
Shall we always open an issue before doing a release so that version number(s) are agreed upon?
(Or do you want to enable Discussions for the repo, and it can be done there?)

kingthorin:
Shall I update the post-release process to increment MIN instead of PATCH?

kingthorin:
Either way we'll have to tweak things manually sometimes, but that isn't a huge deal.

[...]

PR #1712, opened 2025-10-29
Version Update workflow now increments MINOR  instead of PATCH

kingthorin:
Related to datafaker-net/datafaker#1707


kingthorin:
We'll just have to adjust things manually.

kingthorin:
I guess I could add a workflow run trigger to this workflow and require an input of the version number.

Or a separate workflow.

I'll put some thought into it, it can be tackled in another PR.

PR #1704, opened 2025-10-23
Ukrainian Locale Enhancements

vicky-iv:
This pull request expands and improves the Ukrainian locale by adding more country-specific data:
- Added Ukrainian universities
- Vehicle data
- Minor adjustments and formatting changes

Also normalized all generated vehicle license plates to uppercase for consistency.

PR #1715, opened 2025-11-03
Add Color locale tests; migrate color lists from commerce.color to color.name

vicky-iv:
I noticed an inconsistency in how color values are stored across different languages. 
In some files, color was part of the `сommerce` entity, but I couldn’t find any usages of `commerce.color` in the code, and the `Commerce` class doesn’t have a `color()` method.

Therefore, I propose to normalize the color data and always store color as a separate entity in the `color.name` format.

**PR description**
- Introduces `ColorLocaleTest` to verify `Color.name()` localization across many locales.
- Extends `BaseFakerLocaleTest` with additional `BaseFaker` instances 
- Moves color data from `commerce.color` to `color.name` for several locales to normalize where the Color provider reads from and to remove duplication.
- Locales updated in resources: `be.yml`, `by.yml`, `da-DK.yml`, `ru.yml`, `sv.yml`, `uk.yml` (add `color.name` blocks; remove `commerce.color` where applicable), and `uz.yml` (remove `commerce.color` as duplicate).

### DIFF
--- a/src/test/java/net/datafaker/providers/base/ColorLocaleTest.java
+++ b/src/test/java/net/datafaker/providers/base/ColorLocaleTest.java
@@ -2,56 +2,22 @@ package net.datafaker.providers.base;
 
 import org.junit.jupiter.params.provider.Arguments;
 
+import java.util.List;
 import java.util.stream.Stream;
 
 import static org.junit.jupiter.params.provider.Arguments.arguments;
 
 public class ColorLocaleTest extends BaseFakerLocaleTest {
 
-    private final Color colorUA = fakerUA.color();
-    private final Color colorUZ = fakerUZ.color();
-    private final Color colorCH = fakerCH.color();
-    private final Color colorBR = fakerBR.color();
-    private final Color colorDK = fakerDK.color();
-    private final Color colorJP = fakerJP.color();
-    private final Color colorBY = fakerBY.color();
-    private final Color colorArabic = fakerArabic.color();
-    private final Color colorCA = fakerCA.color();
-    private final Color colorCZ = fakerCZ.color();
-    private final Color colorDE = fakerDE.color();
-    private final Color colorFR = fakerFR.color();
-    private final Color colorHU = fakerHU.color();
-    private final Color colorAM = fakerAM.color();
-    private final Color colorKR = fakerKR.color();
-    private final Color colorPT = fakerPT.color();
-    private final Color colorRU = fakerRU.color();
-    private final Color colorSE = fakerSE.color();
-    private final Color colorTH = fakerTH.color();
-    private final Color colorTR = fakerTR.color();
+    private final List<BaseFaker> fakers = List.of(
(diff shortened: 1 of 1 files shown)


asolntsev (approved):
Nice job, thank you @vicky-iv !

P.S. I have been thinking about all these `localeProviderListTest` (like `ColorLocaleTest` in this case). What exactly do they test? What type of error can they catch? 
They seems useless for me...
At least,
* instead of listing all locales manually, we can generate the locales list automatically.
* instead of listing alll yml keys like "color.name" manunally, we could generate this list automatically (e.g. using annotations).

vicky-iv:
@asolntsev Thanks for your comment! 
These tests use the method `net.datafaker.providers.base.AbstractProviderListTest#testProviderList`, which is used across all provider tests. In #1704, I added the ability to use the same method in the locale tests.

> instead of listing all locales manually, we can generate the locales list automatically.

Some locales don't have a list of colors, so I thought it would make sense to add locales manually.

> instead of listing all yml keys like "color.name" manunally, we could generate this list automatically (e.g. using annotations).

Automatically generating the list sounds like a good idea, but it won’t work in all cases. In some tests we might need to cover more complex entities with multiple parameters. A good example is `net.datafaker.providers.base.VehicleLocaleTest#localeProviderListTest`

snuyanzin:
@vicky-iv 
>Automatically generating the list sounds like a good idea, but it won’t work in all cases. In some tests we might need to cover more complex entities with multiple parameters. A good example is 

I added a PR to your branch generalizing this
WDYT about this?
https://github.com/vicky-iv/datafaker/pull/1

vicky-iv:
@asolntsev I slightly improved `ColorLocaleTest`, it still requires manually adding locales because not every locale includes `color`, but the implementation is no longer bulky.

Thanks @snuyanzin for the idea!

asolntsev (approved):
All good, thank you for the PR.

My doubt is more general. I don't understand what exactly does this test verify?
What type of error can it potentially catch? For me, it seems a useless test.
Not only this one, but all the previous similar tests.

```java
    @Override
    protected Stream<Arguments> localeProviderListTest() {
        return fakers.stream()
            .map(faker -> arguments(TestSpec.of(() -> faker.color().name(), "color.name"), faker));
    }
```

kingthorin:
We’ve been through that discussion before.

asolntsev:
... And what was the answer?

kingthorin:
It was a few things:

- It reduces the chances of someone making a copy/paste mistake and having the wrong key resolved for a given method
- It replaced previous regex tests that were simply "this string has content"
- It is meant to be extended further so that more and more content is covered with decent regex patterns (which the TestSpec supports)

If you need more you'll have to go looking. I don't want to rehash it.

asolntsev:
@kingthorin Sorry, I should not waste your time by asking this.
That's right, we already discussed it. Sorry. 

I will try to prepare PR if I get some idea how to improve these tests.

kingthorin:
No worries, sorry if my response seemed negative or harsh; things weren't going well yesterday.

PR #1718, opened 2025-11-07
make locale-related provider tests more generic

asolntsev:
Instead of listing some specific locales, use method `internationalFakers()` that returns all supported locales.

And only in very specific cases (e.g. `UniversityLocaleTest`) run the test for some specific locales (because university names in other locales don't use this pattern).

@vicky-iv What do you think?


vicky-iv:
@asolntsev looks great, thank you!

PR #1730, opened 2025-12-09
Adjective faker based on existing YML file. Generated by Grok Fast 1.

bodiam:


kingthorin (approved):
LGTM

asolntsev (approved):
Looks good, but arises few questions:
1. Why did we need the file `adjective.yml` if we didn't use it (before this PR)?
2. How this "adjective" differs from `words.adjective`?
3. Don't we have duplication in files `adjective.yml` and `words.yml/adjective`?

kingthorin:
Ha! I didn't bother to check, good catch.

erikpragt-connectid:
We don't have a faker class for all YML files. There's a small gap. I knew we had a adjective, but I couldn't find it, so there might be overlap. 

I'll investigate which one we need.

Issue #1737, opened 2025-12-17
GraalVM native-image: Missing inner classes in reachability-metadata.json for expression resolution


felipekj:
## Description

The `reachability-metadata.json` in DataFaker 2.5.3 is incomplete for GraalVM native-image support. Expression resolution like `#{IDNumber.ssn_valid}` fails in native images because several inner classes of `FakeValuesService` are not registered.

## Error

```
java.lang.RuntimeException: Unable to resolve #{IDNumber.ssn_valid} directive for FakerContext
```

## Missing Classes

The following inner classes are essential for expression resolution but are not included in `META-INF/native-image/reachability-metadata.json`:

- `net.datafaker.service.FakeValuesService$MethodResolver`
- `net.datafaker.service.FakeValuesService$MethodAndCoercedArgs`
- `net.datafaker.service.FakeValuesService$MethodAndCoercedArgsResolver`
- `net.datafaker.service.FakeValuesService$ValueResolver`
- `net.datafaker.service.FakeValuesService$SafeFetchResolver`
- `net.datafaker.service.FakeValuesService$ConstantResolver`
- `net.datafaker.service.FakeValuesService$RegExpContext`

## Reproduction

1. Use DataFaker 2.5.3 in a Spring Boot 3.x application
2. Call `faker.idNumber().valid()` which uses `#{IDNumber.ssn_valid}` expression
3. Compile to native image with GraalVM
4. Run native tests - expression resolution fails

## Workaround

Add Spring RuntimeHintsRegistrar with the missing classes:

```java
String[] serviceClasses = {
    "net.datafaker.service.FakeValuesService$MethodResolver",
    "net.datafaker.service.FakeValuesService$MethodAndCoercedArgs",
    "net.datafaker.service.FakeValuesService$MethodAndCoercedArgsResolver",
    "net.datafaker.service.FakeValuesService$ValueResolver",
    "net.datafaker.service.FakeValuesService$SafeFetchResolver",
    "net.datafaker.service.FakeValuesService$ConstantResolver",
    "net.datafaker.service.FakeValuesService$RegExpContext"
};

for (String className : serviceClasses) {
    hints.reflection().registerType(
        classLoader.loadClass(className),
        MemberCategory.INVOKE_DECLARED_CONSTRUCTORS,
        MemberCategory.INVOKE_PUBLIC_CONSTRUCTORS,
        MemberCategory.INVOKE_DECLARED_METHODS,
[...]

PR #1744, opened 2026-01-10
Added config generator for GraalVM and updated config.

bodiam:
This may address https://github.com/datafaker-net/datafaker/issues/1737


asolntsev (changes requested):
Seems that Copilot comments are reasonable this time :)

PR #1757, opened 2026-02-04
avoid '&' in company domain names

fhueter:
Company domain names are created based on the company name and therefore might contain '&' (e.g. for German company names). So we have to remove them from the according domain name.

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Company.java
+++ b/src/main/java/net/datafaker/providers/base/Company.java
@@ -80,7 +80,7 @@ public class Company extends AbstractProvider<BaseProviders> {
         for (int i = 0; i < res.length; i++) {
             final char c = res[i];
             switch (c) {
-                case '.', ',', '\'', ' ', ']' -> offset++;
+                case '.', ',', '\'', ' ', ']', '&' -> offset++;
                 default -> res[i - offset] = res[i];
             }
         }


asolntsev:
@fhueter thank you! Don't you want to add some unit-test with company name containing &?

fhueter:
> @fhueter thank you! Don't you want to add some unit-test with company name containing &?

Was thinking about that as well but since there exists no "hard coded" unit test so far I honestly did not want to start with that single one.

kingthorin:
I'm not sure this is the right way to go (deny list certain characters). I'm sure you intended for this to be a small change (and it does align with the current implementation). 

But I'm pretty sure if we aren't handling/allowing punny code domains then the allow list for DNS is LDH (letter-digit-hyphen). 
[Each segment (splitting on periods) is 1-63 char, hyphen can't be first or last. Entire value including periods must be <=253 char]

- RFC 1035 — original DNS specification
- RFC 1123 — relaxed hostname rules (allowed digits at start)

Maybe that should be a separate piece of work?

fhueter:
Exactly what I was thinking when I found that special classes for domain name creation already. I guess the clean approach would be to use that logic to create a domain name based on a company name or anything else.

But because of the fact that '&'s in company names are killing the build in the project I am working on right now I wanted to provide a small and quick fix for that one without opening a can of worms. 😬

kingthorin:
That's fair. I'll add further changes to my TODO list 😀 

@asolntsev I'll add tests in the next iteration.

asolntsev:
Well, seems that current algorithm for generating domain name doesn't work at all (for non-EN locales).

I slightly modified `CompanyTest.java` to try all locales:

```java
    @RepeatedTest(100)
    void testUrl() throws MalformedURLException, URISyntaxException {
        for (Locale locale : Locale.getAvailableLocales()) {
            String url = new Faker(locale).company().url();
            assertThat(url).matches(URL_PATTERN);
            new URL("https://" + url).toURI();
        }
    }
```

And it fails like crazy :)

```java
Expecting actual:
  "www.xn--80aaaecb3b4bbeeb3b.рф"
to match pattern:
  "(([a-z&0-9]|[a-z&0-9][a-z&0-9\-]*[a-z&0-9])\.)*([a-z&0-9]|[a-z&0-9][a-z&0-9\-]*[a-z&0-9])"
```

```java
Expecting actual:
  "www.xn--()-lqia8ao2eer1ac9fgt4gvb6fgocbg3ag24d.net"
to match pattern:
  "(([a-z&0-9]|[a-z&0-9][a-z&0-9\-]*[a-z&0-9])\.)*([a-z&0-9]|[a-z&0-9][a-z&0-9\-]*[a-z&0-9])"
```

```java
Expecting actual:
  "www.xn--()-lqi1ajd2euayc2ffik4graz1hhqcg3a32d.io"
to match pattern:
  "(([a-z&0-9]|[a-z&0-9][a-z&0-9\-]*[a-z&0-9])\.)*([a-z&0-9]|[a-z&0-9][a-z&0-9\-]*[a-z&0-9])"
```

```java
Expecting actual:
  "www.greenfelderwiegandandheller.бг"
to match pattern:
  "(([a-z&0-9]|[a-z&0-9][a-z&0-9\-]*[a-z&0-9])\.)*([a-z&0-9]|[a-z&0-9][a-z&0-9\-]*[a-z&0-9])"
```

etc.

asolntsev:
@kingthorin @fhueter Even if we remove the check for regex (which might be incorrect) and use only the standard Java's `URL` constructor for validating the generated URL, we will see some invalid URLs:

* `https://www.xn--"--wldubvay2a6exa.gov.il`
* `https://www.xn--jebxn--bebxn--4db"xn--8dbxn--uebxn--pebxn--nebandxn--cebxn--tebxn--9dbxn--vebxn--eeb.com`
* `https://www.xn--5db"xn--sebxn--4dbxn--hebxn--7dbxn--7dbandxn--nebxn--5dbxn--tebxn--9dbxn--keb.net.il`
* `https://www.xn--5dbxn--4dbxn--5db"xn--7dbandsons.net.il`

fhueter:
@asolntsev Is there any chance we might get this fix released in a bugfix version 2.5.4 of datafaker?
Maybe together with other things which got fixed recently? These '&'s in company domain names keep killing our build pretty often and the team is thinking about totally removing datafaker from the project once and for all... 😔

kingthorin:
I can start one.

kingthorin:
https://central.sonatype.com/artifact/net.datafaker/datafaker
image

fhueter:
Thank you very much. That was quick! 
I've already updated the datafaker dependency in our project to 2.5.4!! 👍

PR #1764, opened 2026-02-20
mark all APIs nullable/non-nullable

asolntsev:
JSpecify is de-facto standard annotations for marking nullability of variables, parameters, return values and fields. Supported by all IDEs.

All remarkable libraries have migrated to JSpecify: Spring, JUnit etc.

1. By default, we mark everything as non-nullable (see @NullMarked in package-info.java)
2. When needed, we mark the nullable parameters/fields as @Nullable.

### DIFF
--- a/src/main/java/net/datafaker/internal/helper/SingletonLocale.java
+++ b/src/main/java/net/datafaker/internal/helper/SingletonLocale.java
@@ -1,15 +1,17 @@
 package net.datafaker.internal.helper;
 
-import java.util.HashMap;
+import org.jspecify.annotations.Nullable;
+
 import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * This class allows to use {@link java.util.IdentityHashMap} for locales.
  * To do that it guarantees one instance of {@link SingletonLocale} per {@link Locale}.
  */
 public class SingletonLocale {
-    private static final Map<Locale, SingletonLocale> LOCALE2SINGLETON_LOCALE = new HashMap<>();
+    private static final Map<Locale, SingletonLocale> LOCALE2SINGLETON_LOCALE = new ConcurrentHashMap<>();
     private final Locale locale;
 
     // Hash code is required for FakerContext where SingletonLocale is a field
@@ -19,23 +21,16 @@ public class SingletonLocale {
         this.locale = locale;
     }
 
-    public static SingletonLocale get(Locale locale) {
+    @Nullable
+    public static SingletonLocale get(@Nullable Locale locale) {
         if (locale == null) {
             return null;
         }
-        SingletonLocale res = LOCALE2SINGLETON_LOCALE.get(locale);
-        if (res != null) {
-            return res;
-        }
-        synchronized (SingletonLocale.class) {
-            res = LOCALE2SINGLETON_LOCALE.get(locale);
-            if (res != null) {
-                return res;
-            }
-            res = new SingletonLocale(locale);
-            LOCALE2SINGLETON_LOCALE.put(locale, res);
-            return res;
-        }
+        return getRequired(locale);
+    }
+
+    public static SingletonLocale getRequired(Locale locale) {
+        return LOCALE2SINGLETON_LOCALE.computeIfAbsent(locale, (__) -> new SingletonLocale(locale));
     }
 
     public Locale getLocale() {
--- a/src/main/java/net/datafaker/internal/helper/SingletonLocale.java
+++ b/src/main/java/net/datafaker/internal/helper/SingletonLocale.java
@@ -30,7 +30,19 @@ public class SingletonLocale {
     }
 
     public static SingletonLocale getRequired(Locale locale) {
-        return LOCALE2SINGLETON_LOCALE.computeIfAbsent(locale, (__) -> new SingletonLocale(locale));
+        SingletonLocale res = LOCALE2SINGLETON_LOCALE.get(locale);
+        if (res != null) {
+            return res;
+        }
+        synchronized (SingletonLocale.class) {
+            res = LOCALE2SINGLETON_LOCALE.get(locale);
+            if (res != null) {
+                return res;
+            }
+            res = new SingletonLocale(locale);
+            LOCALE2SINGLETON_LOCALE.put(locale, res);
+            return res;
+        }
     }
 
     public Locale getLocale() {
--- a/src/main/java/net/datafaker/providers/base/Credentials.java
+++ b/src/main/java/net/datafaker/providers/base/Credentials.java
@@ -144,7 +144,6 @@ public class Credentials extends AbstractProvider<BaseProviders> {
 
     /**
      * Generates a user ID based on the regex pattern defined in the resource file.
-     * If the regex is null or invalid, it returns null.
      *
      * @return A randomly generated user ID
      */
--- a/src/main/java/net/datafaker/providers/base/Credentials.java
+++ b/src/main/java/net/datafaker/providers/base/Credentials.java
@@ -149,8 +149,7 @@ public class Credentials extends AbstractProvider<BaseProviders> {
      * @return A randomly generated user ID
      */
     public String userId() {
-        String regex = resolve("credentials.uid_pattern");
-        return faker.regexify(regex);
+        return faker.regexify(resolve("credentials.uid_pattern"));
     }
 
     /**
--- a/src/main/java/net/datafaker/providers/base/Credentials.java
+++ b/src/main/java/net/datafaker/providers/base/Credentials.java
@@ -1,7 +1,12 @@
 package net.datafaker.providers.base;
 
+import org.jspecify.annotations.Nullable;
+
+import java.util.logging.Logger;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
+
+import static java.util.logging.Level.WARNING;
 
 /**
  * Generates credentials such as usernames, uids and passwords.
@@ -12,6 +17,7 @@ public class Credentials extends AbstractProvider<BaseProviders> {
 
     public static final int MIN_PASSWORD_LENGTH = 8;
     public static final int MAX_PASSWORD_LENGTH = 16;
+    private static final Logger log = Logger.getLogger(Credentials.class.getName());
 
     protected Credentials(BaseProviders faker) {
         super(faker);
@@ -142,6 +148,7 @@ public class Credentials extends AbstractProvider<BaseProviders> {
      *
      * @return A randomly generated user ID based on the regex or null if the regex is null or invalid
      */
+    @Nullable
     public String userId() {
         return userId(resolve("credentials.uid_pattern"));
     }
@@ -153,7 +160,8 @@ public class Credentials extends AbstractProvider<BaseProviders> {
      * @param regex The regex pattern to generate the user ID
      * @return A randomly generated user ID based on the regex or null if the regex is null or invalid
      */
-    public String userId(String regex) {
+    @Nullable
+    public String userId(@Nullable String regex) {
         if(regex == null) {
             return null;
         }
@@ -161,6 +169,7 @@ public class Credentials extends AbstractProvider<BaseProviders> {
         try {
             Pattern.compile(regex);
         } catch (PatternSyntaxException e) {
+            log.log(WARNING, e, () -> "Invalid regex pattern: " + regex);
             return null;
         }
 
--- a/src/main/java/net/datafaker/providers/base/Credentials.java
+++ b/src/main/java/net/datafaker/providers/base/Credentials.java
@@ -146,11 +146,11 @@ public class Credentials extends AbstractProvider<BaseProviders> {
      * Generates a user ID based on the regex pattern defined in the resource file.
      * If the regex is null or invalid, it returns null.
      *
-     * @return A randomly generated user ID based on the regex or null if the regex is null or invalid
+     * @return A randomly generated user ID
      */
-    @Nullable
     public String userId() {
-        return userId(resolve("credentials.uid_pattern"));
+        String regex = resolve("credentials.uid_pattern");
+        return faker.regexify(regex);
     }
 
     /**
--- a/src/main/java/net/datafaker/providers/base/Lorem.java
+++ b/src/main/java/net/datafaker/providers/base/Lorem.java
@@ -1,9 +1,9 @@
 package net.datafaker.providers.base;
 
-import net.datafaker.internal.helper.WordUtils;
-
 import java.util.ArrayList;
 import java.util.List;
+
+import static net.datafaker.internal.helper.WordUtils.capitalizeWords;
 
 
 /**
@@ -113,9 +113,9 @@ public class Lorem extends AbstractProvider<BaseProviders> {
     public String sentence(int wordCount, int randomWordsToAdd) {
         int numberOfWordsToAdd = randomWordsToAdd == 0 ? 0 : faker.random().nextInt(randomWordsToAdd);
         final int totalWordCount = wordCount + numberOfWordsToAdd;
-        StringBuilder sb = new StringBuilder();
+        StringBuilder sb = new StringBuilder(wordCount * 9);
         if (totalWordCount > 0) {
-            sb.append(WordUtils.capitalize(word()));
+            sb.append(capitalizeWords(word()));
         }
         for (int i = 1; i < totalWordCount; i++) {
             sb.append(" ").append(word());
--- a/src/main/java/net/datafaker/providers/base/Lorem.java
+++ b/src/main/java/net/datafaker/providers/base/Lorem.java
@@ -113,7 +113,7 @@ public class Lorem extends AbstractProvider<BaseProviders> {
     public String sentence(int wordCount, int randomWordsToAdd) {
         int numberOfWordsToAdd = randomWordsToAdd == 0 ? 0 : faker.random().nextInt(randomWordsToAdd);
         final int totalWordCount = wordCount + numberOfWordsToAdd;
-        StringBuilder sb = new StringBuilder(wordCount * 9);
+        StringBuilder sb = new StringBuilder();
         if (totalWordCount > 0) {
             sb.append(capitalizeWords(word()));
         }
--- a/src/main/java/net/datafaker/service/FakeValuesService.java
+++ b/src/main/java/net/datafaker/service/FakeValuesService.java
@@ -106,8 +106,9 @@ public class FakeValuesService {
      * @param path   path to a file with YAML structure
      * @throws IllegalArgumentException in case of invalid path
      */
-    public void addPath(Locale locale, @Nullable Path path) {
+    public void addPath(Locale locale, Path path) {
         requireNonNull(locale);
+        //noinspection ConstantValue
         if (path == null || Files.notExists(path) || Files.isDirectory(path) || !Files.isReadable(path)) {
             throw new IllegalArgumentException("Path should be an existing readable file: \"%s\"".formatted(path));
         }
--- a/src/main/java/net/datafaker/service/FakeValuesService.java
+++ b/src/main/java/net/datafaker/service/FakeValuesService.java
@@ -3,7 +3,6 @@ package net.datafaker.service;
 import com.github.curiousoddman.rgxgen.RgxGen;
 import net.datafaker.internal.helper.CopyOnWriteMap;
 import net.datafaker.internal.helper.SingletonLocale;
-import net.datafaker.internal.helper.WordUtils;
 import net.datafaker.providers.base.AbstractProvider;
 import net.datafaker.providers.base.Address;
 import net.datafaker.providers.base.BaseFaker;
@@ -15,6 +14,7 @@ import net.datafaker.transformations.Field;
 import net.datafaker.transformations.JsonTransformer;
 import net.datafaker.transformations.Schema;
 import net.datafaker.transformations.SimpleField;
+import org.jspecify.annotations.Nullable;
 
 import java.io.IOException;
 import java.lang.reflect.Array;
@@ -50,6 +50,7 @@ import static java.util.Objects.requireNonNull;
 import static java.util.logging.Level.FINE;
 import static java.util.logging.Level.SEVERE;
 import static net.datafaker.internal.helper.JavaNames.toJavaNames;
+import static net.datafaker.internal.helper.WordUtils.capitalizeWords;
 import static net.datafaker.transformations.Field.field;
 
 public class FakeValuesService {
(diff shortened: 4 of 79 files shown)


bodiam:
Awesome! Just wondering, could this start breaking things for libraries which depend on us, or Kotlin projects? Just wonder if we release this, should this be 2.x release or 3.x ?

asolntsev:
No, it doesn't need to be 3.0
These are just annotations that fix current state of things. They don't change the behavior, they will just help IDE to highlight problematic places in code.

bodiam:
Okay. My understanding was that before the code is annotated, Kotlin will treat this as platform types. So, the following code would work:

```
var name: String? = faker.name().firstName()
var name: String = faker.name().firstName()
```

Both would work. But if I look here: https://resources.jetbrains.com/storage/products/kotlinconf-2025/may-23/JSpecify_%20Java%20Nullness%20Annotations%20and%20Kotlin%20_%20David%20Baker.pdf, they say: 

> The Kotlin compiler treats JSpecify-annotated Java code as null-safe types as of version 2.1.0.

So, if we'd add @Nullable to something Kotlin clients assumed was non-null, they'll get compile errors if they're using it in a non-null context without null checks.

So, something like this might break Kotlin consumers, I think:

```
    @Nullable
    public String userId() {
```

Also, I think the code is perhaps incorrect, I don't think it should ever return a nullable userId. (and yes, I know what the documentation says, but it's also a bit sus) Wdyt?

asolntsev:
> The Kotlin compiler treats JSpecify-annotated Java code as null-safe types as of version 2.1.0.
> So, something like this might break Kotlin consumers, I think:

@bodiam Hm... Yes, that's right: the new annotations can break compilation of existing Kotlin code.
For example:
```kotlin
  val userId1: String = faker.credentials().userId(null);
```

* Previously: it compiled because Kotlin doesn't know if `userId(null)` can return null.
* Now: this **code doesn't compile** because Kotlin knows that `userId(null)` can return null.

So yes, we need to release it as DataFaker 3.0.0 :)


P.S. And yes, good point: method `faker.credentials.userId()` never returns null. I've removed `@Nullable` annotation from this method.

src/main/java/net/datafaker/providers/base/Credentials.java
```diff
@@ -140,10 +146,11 @@ public String weakPassword() {
      * Generates a user ID based on the regex pattern defined in the resource file.
      * If the regex is null or invalid, it returns null.
      *
-     * @return A randomly generated user ID based on the regex or null if the regex is null or invalid
+     * @return A randomly generated user ID
      */
     public String userId() {
-        return userId(resolve("credentials.uid_pattern"));
+        String regex = resolve("credentials.uid_pattern");
+        return faker.regexify(regex);
```
kingthorin: Why not inline it and avoid the intermediate storage?

src/test/java/net/datafaker/formats/YamlTest.java
```diff
@@ -41,7 +42,7 @@ private static Stream<Arguments> generateTestSchema() {
             of(Schema.of(field("number", () -> BigDecimal.valueOf(123.123))), "number: 123.123" + System.lineSeparator()),
             of(Schema.of(field("boolean", () -> true)), "boolean: true" + System.lineSeparator()),
             of(Schema.of(field("nullValue", () -> null)), "nullValue: null" + System.lineSeparator()),
-            of(Schema.of(field("array", () -> new String[]{null, "test", "123"})),
+            of(Schema.of(field("array", () -> new @Nullable String[]{null, "test", "123"})),
```
kingthorin: Are users of datafaker going to have to do this?

Why is it needed here but not the one just before?

kingthorin (commented):
It's kind of a lot to review, especially with the unrelated typo fixes etc.

These are the only things I noticed.

src/main/java/net/datafaker/internal/helper/SingletonLocale.java
```diff
@@ -19,23 +21,16 @@ private SingletonLocale(Locale locale) {
...
-            }
-            res = new SingletonLocale(locale);
-            LOCALE2SINGLETON_LOCALE.put(locale, res);
-            return res;
-        }
+        return getRequired(locale);
+    }
+
+    public static SingletonLocale getRequired(Locale locale) {
+        return LOCALE2SINGLETON_LOCALE.computeIfAbsent(locale, (__) -> new SingletonLocale(locale));
```
snuyanzin: doesn't match PR description

snuyanzin:
is there any enforcer that could fail the build if missed?
Or otherwise it always be in a phase of "eventual consistency"

bodiam:
> So yes, we need to release it as DataFaker 3.0.0 :)
> 
> P.S. And yes, good point: method `faker.credentials.userId()` never returns null. I've removed `@Nullable` annotation from this method.

Just to be clear on this: I do really like the PR, I think we should absolutely do it, just may need a bit of checking on the impact, and how we're communicating this, that's all. I have no issue with releasing DF 3.x either.

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
```diff
@@ -210,6 +210,7 @@ void testLocalesChainGetterRu() {
     }
 
     @Test
+    @SuppressWarnings("deprecation")
```
bodiam: Why is this deprecation warning suppressed, instead of using the `Locale.of` method?

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
```diff
@@ -191,10 +192,10 @@ private enum Column {
         SINCE("Since", 7, Pattern.compile("@since\\s+\\d\\.\\d+\\.\\d"), comment -> comment.group().substring("@since".length()).trim());
         private final String columnName;
         private final int length;
-        private final Pattern pattern2extract;
-        private final Function<Matcher, String> extractor;
+        private final @Nullable Pattern pattern2extract;
+        private final @Nullable Function<Matcher, String> extractor;
 
-        Column(String columnName, int length, Pattern pattern2extract, Function<Matcher, String> extractor) {
+        Column(String columnName, int length, @Nullable Pattern pattern2extract, @Nullable Function<Matcher, String> extractor) {
```
bodiam, line 202: I don't think this is correct. `pattern2extract` cannot be null, it would throw an NPE at line 246.

bodiam, line 202: same for `extractor`, if it's null, it would throw an exception at line 249/250. There's no guards/optional stuff at all on this.

src/test/java/net/datafaker/integration/FakerIntegrationTest.java
```diff
@@ -88,7 +89,7 @@ private Faker init(Locale locale, Random random) {
 
     @ParameterizedTest
     @MethodSource("dataParameters")
-    void testAllFakerMethodsThatReturnStrings(Locale locale, Random random) throws Exception {
+    void testAllFakerMethodsThatReturnStrings(@Nullable Locale locale, @Nullable Random random) throws Exception {
```
bodiam, line 92: I have no idea if these should be nullable or not. Cannot image they should be.

src/test/java/net/datafaker/formats/YamlTest.java
bodiam: I don't know either, but a nullable string[] which is not null? looks weird. I think this could be removed?

src/main/java/net/datafaker/providers/base/Lorem.java
```diff
@@ -113,9 +113,9 @@ public String sentence(int wordCount) {
     public String sentence(int wordCount, int randomWordsToAdd) {
         int numberOfWordsToAdd = randomWordsToAdd == 0 ? 0 : faker.random().nextInt(randomWordsToAdd);
         final int totalWordCount = wordCount + numberOfWordsToAdd;
-        StringBuilder sb = new StringBuilder();
+        StringBuilder sb = new StringBuilder(wordCount * 9);
```
snuyanzin: what is this magic 9 ?

src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
```diff
@@ -163,7 +164,7 @@ public String getEndStream() {
 
     private Object getObject(Schema<Object, ?> schema, Object result, Constructor<?> recordConstructor) {
         final Field<Object, ?>[] fields = schema.getFields();
-        final Object[] values = new Object[fields.length];
+        final @Nullable Object[] values = new Object[fields.length];
```
bodiam, line 170: It's not really clear to me here wheat the intention is. Why can a final reference to values which is filled by an object be nullable? I don't think it can?

src/main/java/net/datafaker/transformations/CompositeField.java
```diff
@@ -1,24 +1,28 @@
...
+import org.jspecify.annotations.Nullable;
 
 import java.util.Objects;
 
 public class CompositeField<MyObject extends AbstractProvider<?>, MyType> extends Schema<MyObject, MyType> implements Field<MyObject, MyType> {
+    @Nullable
     private final String name;
 
-    public CompositeField(String name, Field<MyObject, MyType>[] fields) {
+    public CompositeField(@Nullable String name, Field<MyObject, MyType>[] fields) {
```
bodiam, line 12: I don't understand why a compositeField can have a nullable name.

src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
```diff
@@ -375,7 +376,7 @@ public String getEndStream() {
         throw new UnsupportedOperationException("Not supported for SQL transformer. Use generate or generateStream instead to create SQL statements.");
     }
 
-    private String generateBatchModeStatements(Schema<IN, ?> schema, List<IN> inputs, int limit) {
+    private String generateBatchModeStatements(Schema<IN, ?> schema, @Nullable List<IN> inputs, int limit) {
```
bodiam, line 379: I don't think collection types should ever be nullable.

src/main/java/net/datafaker/transformations/TomlTransformer.java
```diff
@@ -55,9 +61,9 @@ public String getEndStream() {
         return "";
     }
 
-    private void writeSchema(StringBuilder sb, IN input, Schema<IN, ?> schema, String pathPrefix) {
+    private void writeSchema(StringBuilder sb, @Nullable IN input, Schema<IN, ?> schema, String pathPrefix) {
         Field<IN, ?>[] fields = schema.getFields();
-        Set<String> seen = new HashSet<>();
+        Set<@Nullable String> seen = new HashSet<>();
```
snuyanzin, line 67: do you have a case when it might be null?

It looks like annotations sometimes were placed very randomly

snuyanzin:
@asolntsev can you explain how you put annotations?
It looks like in some cases it makes some in some cases there are placed very randomly

src/main/java/net/datafaker/service/FakeValuesService.java
```diff
@@ -105,7 +106,7 @@ private FakeValuesInterface getCachedFakeValue(SingletonLocale locale) {
      * @param path   path to a file with YAML structure
      * @throws IllegalArgumentException in case of invalid path
      */
-    public void addPath(Locale locale, Path path) {
+    public void addPath(Locale locale, @Nullable Path path) {
```
bodiam: this doesn't seem to be correct. It's nullable, but when it is null, we thorw an IllegalArgumentException. Don't think that's correct?

bodiam:
I used Claude Code to review it as well:

## Summary

  This branch introduces JSpecify (@NullMarked, @Nullable) and Error Prone (@CheckReturnValue) annotations across the entire codebase. It adds package-info.java files to 22 packages, making non-null the default for
   all parameters and return types, and explicitly marks nullable locations with @Nullable. Along the way, it includes several cleanup improvements: thread-safety fixes, dead-code removal, raw type elimination, and
   typo corrections.

## Findings

### Critical

  None found.

### Improvements

  1. Vehicle.licensePlate(String) - Javadoc still references old behavior (Vehicle.java:147)
  The Javadoc at line 147 still says "If the regex is null or invalid, it returns null" - wait, that's actually on Credentials. For Vehicle.licensePlate, the method now returns @Nullable but the null check on
  licensePlatesByState was removed. This is correct since resolve() throws rather than returning null - but this is a subtle contract dependency. A brief comment explaining why the null check was removed would help
   future readers.
  2. Credentials.userId() - stale Javadoc (Credentials.java:146-149)
  The no-arg userId() Javadoc still says "If the regex is null or invalid, it returns null" (line 147), but the method no longer delegates to the nullable overload and will now throw if resolve() fails. The Javadoc
   should be updated to match.
  3. SingletonLocale.get() still returns @Nullable (SingletonLocale.java:22-27)
  The get(Locale) method accepts @Nullable and returns @Nullable. Several call sites now wrap it with requireNonNull(SingletonLocale.get(locale), ...). Consider whether the @Nullable accepting overload is still
  needed. Currently FakerContext has a private singletonLocale() helper, and FakeValuesContext has an inline requireNonNull. Since get(null) is only useful for the one place that truly passes a nullable locale, it
  might be cleaner to have callers handle the null check themselves and make get() non-null-in-non-null-out - but this is a judgment call given this is already the chosen direction.
  4. FakeValuesService.addPath accepts @Nullable Path (FakeValuesService.java:212)
  The @Nullable on the path parameter is odd for a public API since the next line immediately throws on null. Consider keeping it non-null in the signature and letting NullMarked enforce it at compile time, rather
  than accepting null only to immediately reject it. Same goes for addUrl where url is now enforced via requireNonNull but the parameter isn't marked @Nullable (which is actually the better approach).
  5. TomlTransformer.writeSchema - Set<@Nullable String> (TomlTransformer.java:68)
  The Set<@Nullable String> for seen looks intentional (since f.getName() can return null), but adding null to a HashSet and then calling add(null) again to detect duplicates could lead to confusing "duplicate key"
   errors for unnamed fields. This seems pre-existing but worth a thought.
  6. @SuppressWarnings("ConstantValue") in WeightedRandomSelector (WeightedRandomSelector.java:59,67,75)
  These are appropriate since the validation methods guard against null at the API boundary. Just confirming this was intentional and correct.
  7. LazyEvaluated.get() uses requireNonNull (LazyEvaluated.java:26)
  The requireNonNull(value, "Null value not allowed") is correct for the contract (supplier shouldn't return null), but the message "Null value not allowed" could be more descriptive, e.g., "Supplier returned null"
   or "LazyEvaluated value was null after evaluation".

### Nitpicks

  - Lorem.java:116: new StringBuilder(wordCount * 9) - the initial capacity heuristic (9 chars per word) is a reasonable optimization. Minor: a comment explaining the magic number would be nice.
  - FakeValuesGrouping.java: Nice cleanup removing raw types and adding the null check before putAll.
  - Internet.BotUserAgent.any: Good simplification using nextEnum().
  - Typo fixes ("estonian" -> "Estonian", "summ" -> "sum", "criterias" -> "criteria", "a algorithm" -> "an algorithm") are all welcome.
  - The CsvTest exception type change from IllegalArgumentException to NullPointerException correctly tracks the behavioral change in SimpleField.transform().

## Conclusion

  Approved. This is a high-quality, thorough nullability annotation pass. The @NullMarked + @Nullable approach with JSpecify is the modern best practice for Java, the provided scope is correct, and the various code
   cleanups (thread-safety in SingletonLocale, dead code removal in Vehicle, raw type cleanup in FakeValuesGrouping) are all improvements. The setLocale being made final to avoid virtual dispatch from the
  constructor is a particularly nice detail.

  The only item I'd suggest addressing before merge is the stale Javadoc on Credentials.userId() (finding #2).

asolntsev:
> @asolntsev can you explain how you put annotations?
> It looks like in some cases it makes some in some cases there are placed very randomly

As I explained in the PR description:

1. By default, we mark everything as non-nullable (see @NullMarked in package-info.java)
2. When needed, we mark the nullable parameters/fields as @nullable.

So I added @Nullable in places where i see that null is actually expected. Of course, I might make a mistake in some of these places (but I hope very few).

src/main/java/net/datafaker/providers/base/Credentials.java
asolntsev: Yep, done.

src/main/java/net/datafaker/internal/helper/SingletonLocale.java
asolntsev: Ok, I've reverted the unrelated changes.

src/main/java/net/datafaker/providers/base/Lorem.java
asolntsev: It's a heuristic number to predict the resulting string length. Good for performance for long strings.

Anyway, I've reverted this change in this PR.

src/test/java/net/datafaker/service/FakeValuesServiceTest.java
asolntsev: Hm... I don't remember why this suppression was needed... :( 

I've removed it, and don't see any deprecation warnings... 
Sorry for the hassle.

bodiam: It's a good puzzle, I like it :)

src/main/java/net/datafaker/service/FakeValuesService.java
asolntsev: Good point. Now I see that `path` cannot be null.

Previously, I saw this piece and decided that path is nullable:

```java
if (path == null ... ) 
```

src/main/java/net/datafaker/transformations/sql/SqlTransformer.java
asolntsev: @bodiam I don't know the initial intention, but there is code that passes `null` value to `inputs` parameter:

Image

src/main/java/net/datafaker/transformations/CompositeField.java
asolntsev: @bodiam Me too. :)

But this is reality: there are multiple usages with `null` name:

Image

NB! All these usages are in tests. Perharps these are just incorrect tests? If you are sure name should never be null, I can make it non-nullable and fix those tests.

src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
asolntsev: This annotation doesn't mean "values can be null". It means "values is an array that can hold null values".

E.g. 
```java
final @Nullable Object[] values = new Object[fields.length];
values[0] = null;
values[1] = null;
```

src/main/java/net/datafaker/transformations/TomlTransformer.java
asolntsev: See my comment https://github.com/datafaker-net/datafaker/pull/1764/changes#r2847618881

This `seen` can hold null values because `Field.getName()` is nullable. If we decide to fix https://github.com/datafaker-net/datafaker/pull/1764/changes#r2847618881, then I will remove this `@Nullable` from `seen`.

src/test/java/net/datafaker/formats/YamlTest.java
asolntsev: @bodiam Expression `new @Nullable String[]` means "create an array that can hold null values".

Without `@Nullable`, we would see a warning:

Image


@kingthorin No, users of datafaker don't always need to do it. 
Only if they enable nullability warnings in IDE and want to fix them all.

src/test/java/net/datafaker/integration/FakerIntegrationTest.java
asolntsev: Just check the usages.
Method `dataParameters` sends some "null" values:

Image

My annotation just reflects the reality,

src/test/java/net/datafaker/script/ProvidersDocsGenerator.java
asolntsev: Yes, it's a bit tricky code.
This is nullable - there is `Column` constructor that passes null values:

Image

Enum `Column` has 4 values, and two of them (`NAME` and `GROUP`) have null values:

Image

The trick is that method `getValue()` never gets called for `NAME` and `GROUP`, thus never causing NPE. 

I would like to make this code easily trackable, but it's out of scope for this PR.

asolntsev: same as https://github.com/datafaker-net/datafaker/pull/1764/changes#r2832698599

asolntsev:
> is there any enforcer that could fail the build if missed?

Yes, there is library "NullAway" (probably some others as well).
I think Spotbugs can also be configured to fail the build on nullability problems. 
I just thought it would be too much for one PR. We can enable such enforcer in a separate PR.

src/main/java/net/datafaker/transformations/CompositeField.java
asolntsev: @bodiam @snuyanzin I've found the case where `field.name` can actually be `null` - in DF tests.

Here is a simplified version:

```java
@Test
void xmlSchemaTest() {
  Schema<String, String> schema = Schema.of(
    compositeField("input", new Field[]{
      field("type", () -> "button"), 
      field(null, () -> "Click me")  // NAME IS NULL!
    })
  );
  String expected = "<input type=\"button\">Click me</input>";

  XmlTransformer<String> xmlTransformer = new XmlTransformer.XmlTransformerBuilder<String>().build();
  assertThat(xmlTransformer.generate(schema, 1)).isEqualTo(expected);
    }
```

If I try to set non-null field name (`field("SOME-FIELD-NAME", () -> "Click me")`) then I get a different XML:

```xml
<input type="button" SOME-FIELD-NAME="Click me"/>
```

kingthorin:
I don't feel like my skills/knowledge give me a good place from which to review this PR.

As a contributor how would I know when/if I should be annotating something?

asolntsev:
@kingthorin 

> I don't feel like my skills/knowledge give me a good place from which to review this PR.

Yes, unfortunately, such PR is hard to review. 
If you really want, you can checkout this branch in review IDE warnings about potential nullability problems. 


> As a contributor how would I know when/if I should be annotating something?
What I did:
1. Marked everything as non-nullable by default (added `@NullMarked` in `package-info.java` files in all packages).
2. Went through the codebase (reviewed every single file one by one) and looked for IDEA warnings. In every single place where I saw that `null` is actually passed to variable/parameter/field, I marked this specific var/param/field as `@Nullable`.

Looks massive, but actually simple.

kingthorin:
Seems reasonable to me

kingthorin (approved):

asolntsev:
@kingthorin @bodiam @snuyanzin What about releasing [2.6.0](https://github.com/datafaker-net/datafaker/milestone/8) with all current developments, and then [3.0.0](https://github.com/datafaker-net/datafaker/milestone/9) with nullability annotations?

^^^

bodiam:
Hi @asolntsev , we've release 2.7.0, I have no objection with preparing for a 3.0.0 release, what do you think?

asolntsev:
Yes, I would be happy if we release 3.0.0

Otherwise we have a risk that all the nullability changes will get more and more merge conflicts to resolve.

kingthorin:
I can switch all the artifacts for an upcoming 3.0.0 release, if we are sure that's the plan. (The snapshot references etc)

We are sticking with targeting Java 17, right?

asolntsev:
It would be great!
I've rebased the PR on the latest main. There were not too many conflicts, fortunately. :) 

Yes, we are  sticking to Java 17. I don't see any good reason to require Java18+ from our users.

kingthorin:
Okay, I'll do a PR prepping for 3.0.0 later today.

asolntsev:
Ok, can I then merge this PR?
I think I prefer to squash it because it contains one essential commit and several "fix-ups".

kingthorin:
That's fine with me.

bodiam:
I think it would be more like Java 21/Java 25, but I think as a general rule of thumb we should stay aligned with the minimum version used by Spring, it think that would be a safe minium version, and that is still Java 17, so let's stick to that for now?

bodiam:
@asolntsev if it's just 3.0.0-snapshot, I think it's fine to merge, let's see how we go!

PR #1803, opened 2026-04-26
Introduce `iban-commons` for IBAN generation and validation

spannm:
### Summary

Replaces hand-written IBAN logic in `Finance.java` with open-source library [`de.speedbanking:iban-commons`](https://www.speedbanking.de/), significantly increasing country coverage and removing ~130 lines of maintenance-heavy boilerplate.

### Changes

**`pom.xml`**
- Add `de.speedbanking:iban-commons:1.8.5` as a runtime dependency

**`Finance.java`**
- Remove hand-maintained `Map<String, String>` of 89 country regex patterns
- Remove manual checksum calculation (`calculateIbanChecksum`) and zero-padding helper
- `ibanSupportedCountries()` now delegates to `IbanRegistry.values()`
- `iban(countryCode)` now delegates to `RandomIban.of(countryCode).toString()`

**`FinanceTest.java`**
- Strengthen IBAN assertion for all country codes: `isNotBlank()` → `Iban.isValid()`

### Notes

The set of supported countries is now authoritative from the library — no manual updates needed when new countries are added to the IBAN standard.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@
         <rgxgen.version>3.1</rgxgen.version>
         <google.javaformat.version>1.17.0</google.javaformat.version>
         <jacoco-maven-plugin.version>0.8.14</jacoco-maven-plugin.version>
-        <jbanking.version>4.3.0</jbanking.version>
+        <iban-commons.version>1.8.5</iban-commons.version>
         <junit.version>6.0.3</junit.version>
         <libphonenumber.version>9.0.29</libphonenumber.version>
         <maven.compiler.source>17</maven.compiler.source>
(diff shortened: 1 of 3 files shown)


kingthorin:
https://github.com/datafaker-net/datafaker/blob/main/CONTRIBUTING.md#dependencies

~215kb

https://mvnrepository.com/artifact/de.speedbanking/iban-commons/1.8.5

kingthorin:
Copilot's comments seem correct. Also needs a spotlessApply

`23:11:16.385 [ERROR] Run 'mvn spotless:apply' to fix these violations.`

bodiam:
I don't think this is aligned with our "contributing.md" guide. Closing.

spannm:
hi - Thanks for the quick review and getting back to me.

I understand the concern, your dependency policy is clear about keeping Datafaker slim. With 215 KB, `iban-commons` is technically within the <1 MB limit, but I realize that the new dep does not add enough value to justify its addition.

I will submit a leaner PR that extends the existing regex map with the missing countries to gain broader country coverage without adding any dependency.

bodiam:
I don't mind the size of the library much, but I mind the "Fills a need we can't do "simply"" part. 

Also, it would have been good manners to indicate you're the author of the library, it could have changed the reception on my end a bit. Now I'm more keen on using https://github.com/arturmkrtchyan/iban4j, which seems to address the same issues, and seems to be more widely accepted.

spannm:
Thanks Erik @bodiam, I have only good intentions.

PR #1816, opened 2026-05-15
fix: refine Spanish localization entries

ferclager:
Updated various entries in the Spanish localization file, including city prefixes, country names, street suffixes, and payment methods to correct accents and improve consistency.


bodiam (approved):
Looks good to me, gracias!!

PR #1817, opened 2026-05-15
fix: formatting and update examples in README

ferclager:
Updated README to correct formatting and clarify examples.


bodiam:
thanks! Approved!

PR #1835, opened 2026-06-08
Add dedicated Http faker with status codes, headers, user agents and more

bodiam:
Introduces net.datafaker.providers.base.Http with:
- Status codes by category: statusCode(), informational(), successful(),
  redirect(), clientError(), serverError() — all return int
- Reason phrases: statusMessage() and per-category *Message() variants
- Combined form: statusCodeWithReason() → "200 OK"
- Headers: requestHeader(), responseHeader()
- MIME types: contentType()
- User agents: userAgent(), userAgent(Browser), mobileUserAgent()
  covering Chrome, Firefox, Safari, Edge, and mobile browsers
- Protocol: httpMethod(), httpVersion(), encoding()
- Response bodies: responseBody() with JSON success/error templates

Registers http.yml in EnFile and Http in BaseProviders.
Includes HttpTest with 168 assertions covering all methods.

https://claude.ai/code/session_01WM2gJRU7t1i8sXu6wLx6ak


src/main/resources/en/http.yml
```diff
@@ -0,0 +1,335 @@
...
+      - "audio/mpeg"
+      - "audio/ogg"
+      - "audio/wav"
+      - "video/mp4"
+      - "video/webm"
+      - "video/ogg"
+      - "font/ttf"
+      - "font/woff"
+      - "font/woff2"
+      user_agent:
```
bodiam, line 201: There's a bit of duplication here with `internet()`, I know, but maybe internet is not the best name anymore.

src/main/java/net/datafaker/providers/base/Http.java
```diff
@@ -0,0 +1,250 @@
+package net.datafaker.providers.base;
+
+import java.util.List;
+
+/**
+ * Faker for HTTP-related data: status codes and reason phrases, request/response headers,
+ * content types, user agents, HTTP methods and versions, transfer encodings, and response bodies.
+ *
+ * @since 2.4.0
```
bodiam:
```suggestion
 * @since 2.6.0
```

src/main/java/net/datafaker/providers/base/Http.java
```diff
@@ -0,0 +1,250 @@
...
+ *
+ * @since 2.6.0
+ */
+public class Http extends AbstractProvider<BaseProviders> {
+
+    private static final List<String> ALL_STATUS_CATEGORIES = List.of(
+        "informational", "successful", "redirect", "client_error", "server_error"
+    );
+
+    private static final List<String> ALL_BODY_TYPES = List.of(
```
kingthorin: I'd suggest you call this TEXT_BODY_TYPES or something like that. It definitely isn't "all" and it doesn't cover any non-text based types as far as I can tell. (image etc)

src/main/java/net/datafaker/providers/base/Http.java
```diff
@@ -0,0 +1,250 @@
...
+        "json", "xml", "html", "plain", "csv", "javascript", "css", "graphql", "form", "markdown"
+    );
+
+    protected Http(BaseProviders faker) {
+        super(faker);
+    }
+
+    // -------------------------------------------------------------------------
+    // Status codes — return int
+    // -------------------------------------------------------------------------
```
kingthorin: This seems kinda redundant

src/main/java/net/datafaker/providers/base/Http.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    // -------------------------------------------------------------------------
+    // Combined code + reason — return String like "200 OK"
+    // -------------------------------------------------------------------------
+
+    /** Returns a random status code and reason phrase, e.g. {@code "200 OK"} or {@code "404 Not Found"}. */
+    public String statusCodeWithReason() {
+        return randomStatusEntry();
+    }
```
kingthorin: Why multi-line single line comment? Why not just make proper JavaDoc? (Esp in a single method like this)

src/main/resources/en/http.yml
```diff
@@ -0,0 +1,335 @@
...
+        - "{\"data\":{\"users\":[{\"id\":\"1\",\"name\":\"Alice\"},{\"id\":\"2\",\"name\":\"Bob\"}],\"totalCount\":2}}"
+        - "{\"data\":{\"createUser\":{\"id\":\"42\",\"name\":\"New User\",\"createdAt\":\"2024-01-15T10:00:00Z\"}}}"
+        - "{\"data\":null,\"errors\":[{\"message\":\"Not found\",\"locations\":[{\"line\":1,\"column\":1}],\"path\":[\"user\"]}]}"
+        - "{\"data\":null,\"errors\":[{\"message\":\"Unauthorized\",\"extensions\":{\"code\":\"UNAUTHENTICATED\"}}]}"
+        form:
+        - "name=John+Doe&email=john%40example.com&status=active"
+        - "username=admin&password=secret&remember=true"
+        - "id=42&action=update&value=hello+world"
+        - "grant_type=authorization_code&code=abc123&redirect_uri=https%3A%2F%2Fexample.com%2Fcallback"
+        - "query=search+term&page=1&per_page=20&sort=asc"
```
kingthorin: This doesn't make sense. When have you seen a response like this?

(I'm not saying it's impossible, but form url encoding is generally associated with requests)

src/main/resources/en/http.yml
```diff
@@ -0,0 +1,335 @@
...
+      - "DELETE"
+      - "PATCH"
+      - "HEAD"
+      - "OPTIONS"
+      - "TRACE"
+      http_version:
+      - "HTTP/1.0"
+      - "HTTP/1.1"
+      - "HTTP/2"
+      - "HTTP/3"
```
kingthorin, line 245: Add 0.9?

src/main/java/net/datafaker/providers/base/Http.java
```diff
@@ -0,0 +1,250 @@
...
+
+    private static String parseReason(String statusEntry) {
+        return statusEntry.substring(statusEntry.indexOf(' ') + 1);
+    }
+
+    // -------------------------------------------------------------------------
+    // Enums
+    // -------------------------------------------------------------------------
+
+    public enum Browser {
```
kingthorin: Wow Claude, enums are enums? Insightful

src/main/java/net/datafaker/providers/base/Http.java
bodiam: it's all within the scope of the change, it's all all in the scope of the universe.

src/main/java/net/datafaker/providers/base/Http.java
bodiam: how so? you mean the int? or the whole comment?

src/main/java/net/datafaker/providers/base/Http.java
bodiam: Idk :) Ask Claude. But also, I wasn't sure if we actually enforced the formatting on this. I don't think we do, but I'm happy to change it.

src/main/resources/en/http.yml
bodiam: why do you want 0.9?

src/main/java/net/datafaker/providers/base/Http.java
bodiam: I used Mythos. Maybe that's why ;)

src/main/java/net/datafaker/providers/base/Http.java
kingthorin: The whole comment. The methods declare their return type.

src/main/java/net/datafaker/providers/base/Http.java
kingthorin: I don't think spotless has anything for this. I'd rather the muitiple single line comment ummm "section breaks" were removed, if there's anywhere they're actually valuable then merge with the JavaDoc.

src/main/resources/en/http.yml
kingthorin: Because it's a valid value.

src/main/resources/en/http.yml
bodiam: Yes, you're right. Maybe I'll just remove form from the list.

src/main/resources/en/http.yml
bodiam: I don't think that's a good enough reason. I think it's a very niche usecase, I'm not going to include it just because it's valid.

kingthorin: Alright.

bodiam:
@kingthorin addressed PR feedback. Wdyt?

kingthorin (approved):
LGTM!

PR #1849, opened 2026-06-21
Do not create RegExpContext each time

snuyanzin:
`RegExpContext` is created each time making allocation.
This could be avoided.

Even more small benchmark with methods with expressions
```java
 @Benchmark
    public void one(Blackhole blackhole) {
        blackhole.consume(DATA_FAKER.expression("#{Name.firstName}"));
    }

    @Benchmark
    public void two(Blackhole blackhole) {
        blackhole.consume(DATA_FAKER.expression("#{Name.firstName} #{Address.city}"));
    }

    @Benchmark
    public void three(Blackhole blackhole) {
        blackhole.consume(DATA_FAKER.expression("#{Name.firstName} #{Address.city} #{Name.lastName}"));
    }
```
shows that it can improve throughput about 10%

like 
before
```
Benchmark                    Mode  Cnt     Score     Error   Units
DatafakerExpressions.one    thrpt   10  7778.462 ± 272.241  ops/ms
DatafakerExpressions.three  thrpt   10  2447.599 ±  69.891  ops/ms
DatafakerExpressions.two    thrpt   10  2910.791 ±  58.837  ops/ms

```
after
```
 Benchmark                    Mode  Cnt     Score     Error   Units
 DatafakerExpressions.one    thrpt   10  9001.782 ± 608.820  ops/ms
 DatafakerExpressions.three  thrpt   10  2746.053 ± 136.003  ops/ms
 DatafakerExpressions.two    thrpt   10  3260.148 ±  93.052  ops/ms
```

### DIFF
--- a/src/main/java/net/datafaker/service/FakeValuesService.java
+++ b/src/main/java/net/datafaker/service/FakeValuesService.java
@@ -81,7 +81,7 @@ public class FakeValuesService {
 
     private static final Map<String, String[]> EXPRESSION_2_SPLITTED = new CopyOnWriteMap<>(WeakHashMap::new);
 
-    private final Map<String, RegExpContext> REGEXP2SUPPLIER_MAP = new CopyOnWriteMap<>(HashMap::new);
+    private final Map<String, RegExpContext> regexp2SupplierMap = new CopyOnWriteMap<>(HashMap::new);
     public void updateFakeValuesInterfaceMap(List<SingletonLocale> locales) {
         for (final SingletonLocale l : locales) {
             fakeValuesInterfaceMap.computeIfAbsent(l, this::getCachedFakeValue);
@@ -593,7 +593,7 @@ public class FakeValuesService {
                 }
                 continue;
             }
-            final RegExpContext cachedRegExpContext = REGEXP2SUPPLIER_MAP.get(expr);
+            final RegExpContext cachedRegExpContext = regexp2SupplierMap.get(expr);
             final Object resolved;
             if (cachedRegExpContext != null && cachedRegExpContext.root() == root && cachedRegExpContext.context() == context) {
                 resolved = cachedRegExpContext.resolver().resolve();
@@ -685,7 +685,7 @@ public class FakeValuesService {
         LOG.fine(() -> "resExp(%s [%s]) current: %s, root: %s, context: %s, expr: %s -> res: %s".formatted(directive, Arrays.toString(args), current, root, context, expr, res));
         if (res instanceof CharSequence) {
             if (((CharSequence) res).isEmpty()) {
-                REGEXP2SUPPLIER_MAP.put(expr, new RegExpContext(root, context, EMPTY_STRING));
+                regexp2SupplierMap.put(expr, new RegExpContext(root, context, EMPTY_STRING));
             }
             return res;
         }
@@ -699,7 +699,7 @@ public class FakeValuesService {
                     if (value == null) {
                         it.remove();
                     } else {
-                        REGEXP2SUPPLIER_MAP.put(expr, new RegExpContext(root, context, resolver));
+                        regexp2SupplierMap.put(expr, new RegExpContext(root, context, resolver));
                         return value;
                     }
                 }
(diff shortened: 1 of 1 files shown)


asolntsev:
I don't understand how a simple record `RegExpContext` can have a significant effect on performance. 
It's just a Java record.
Java can easily create and remove records. Should not be a problem.

Also, is the new code effectively the same?
* Previously, `RegExpContext` contained different records for different Locales and RandomServices.
* Now, `RegExpContext` contains only one record per regex. Wouldn't it mix different locales and RegExpContexts?

kingthorin:
Seems like these benchmarks could be added to our test suite and that they should be expanded for multi-locale scenarios.

snuyanzin:
>Also, is the new code effectively the same?

>    Previously, RegExpContext contained different records for different Locales and RandomServices.
    Now, RegExpContext contains only one record per regex. Wouldn't it mix different locales and RegExpContexts?

there is at least one missing piece here
previously it was static and these could be actual points.

Now it is non static. It means for every instance of `Faker` there is its own `RegExpContext`.
The only place where `Locale` or `seed` might play a role is `doWith` however there is a force set/reset of them.

Also there is `FakerTest#testDeterministicAndNonDeterministicProvidersReturnValues` checking it is different output.

Another test
```java
        System.out.println(faker.expression("#{Name.name}#{Name.name}"));
        System.out.println(faker.expression("#{Name.name}"));
        System.out.println(faker.doWith(() -> faker.expression("#{Name.name}"), Locale.JAPANESE));
        System.out.println(faker.expression("#{Name.name}#{Name.name}"));
        Faker faker2 = new Faker();
        System.out.println(faker2.expression("#{Name.name}#{Name.name}"));
        System.out.println(faker2.expression("#{Name.name}"));
        System.out.println(faker2.doWith(() -> faker2.expression("#{Name.name}"), Locale.JAPANESE));
        System.out.println(faker2.expression("#{Name.name}#{Name.name}"));
```

shows
```
Eusebio HickleBetsy Block
Hal Wilderman DVM
西沢 恵子
Brian HansenAraceli Haley
Roscoe BorerShon Senger
Valentine Franecki
関口 萌
Mr. Garfield WillmsWalker Little
```

snuyanzin:
> Java can easily create and remove records. Should not be a problem.

even if it is fast it doesn't mean it is free

```java
    @Benchmark
    public void one(Blackhole blackhole) {

        blackhole.consume(0);
    }
```
vs 
```java
    @Benchmark
    public void one(Blackhole blackhole) {

        blackhole.consume(new myRecord(reg, context, resolver)); // all arg are consts here
    }
```

shows
```
Benchmark                  Mode  Cnt        Score       Error   Units
DatafakerExpressions.one  thrpt   10  4508344.705 ± 96119.272  ops/ms
```
vs 
```
Benchmark                  Mode  Cnt       Score      Error   Units
DatafakerExpressions.one  thrpt   10  709334.454 ± 9181.791  ops/ms
```

snuyanzin:
>Seems like these benchmarks could be added to our test suite and that they should be expanded for multi-locale scenarios.

IIRC the license might be an issue since there is another license for jmh

that's why we have a separate project https://github.com/datafaker-net/datafaker-benchmark

src/main/java/net/datafaker/service/FakeValuesService.java
```diff
@@ -81,7 +81,7 @@ public class FakeValuesService {
 
     private static final Map<String, String[]> EXPRESSION_2_SPLITTED = new CopyOnWriteMap<>(WeakHashMap::new);
 
-    private final Map<RegExpContext, ValueResolver> REGEXP2SUPPLIER_MAP = new CopyOnWriteMap<>(HashMap::new);
+    private final Map<String, RegExpContext> REGEXP2SUPPLIER_MAP = new CopyOnWriteMap<>(HashMap::new);
```
kingthorin: Name should no longer be FULL CAPS

snuyanzin: makes sense, renamed

kingthorin (approved):

asolntsev (approved):

PR #1855, opened 2026-06-24
Fix lowercase country code in Finance.bic()

### LINKED ISSUE #1853: BIC faker broken in R2.6
**Describe the bug**
ISO 9362 (BIC) char set = uppercase A–Z + digits only. 
Country segment = ISO 3166-1 alpha-2, and ISO 3166-1 mandates uppercase alpha-2 codes. 
Means lowercase BIC = non-conformant. 

References:
- ISO 9362:2022 — https://www.iso.org/standard/60390.html
- ISO 3166-1 alpha-2 (uppercase) — https://www.iso.org/iso-3166-country-codes.html
- SWIFT BIC structure — https://www.swift.com/standards/data-standards/bic-business-identifier-code

**To Reproduce**
```
import java.util.regex.Pattern;
import net.datafaker.Faker;
public class BicCaseReproduction {
  public static void main(String[] args) {
    // ISO 9362: 8 or 11 uppercase alphanumerics.
    Pattern ISO_9362 = Pattern.compile("^[A-Z0-9]{8,11}$");
    Faker faker = new Faker(new Random(0));
    System.out.println("ISO9362_OK="+ISO_9362.matcher(faker.finance().bic()).matches());
  }
}
```

**Expected behavior**
an ISO 9362:2022 compliant string

**Versions:**
 - DataFaker Version 2.6.0

**Additional context**
I think the issue has been introduced in [this commit](https://github.com/datafaker-net/datafaker/commit/d4267942d61314017cba303f98cd607d071409f4) which is part of *PR* #1805 . The previous version was producing `return faker.regexify("([A-Z]){4}([A-Z]){2}([0-9A-Z]){2}([0-9A-Z]{3})?");` while the new version is a little more complicated : 
```
String countryCode = faker.country().countryCode2();
String regex = String.format("[A-Z]{4}%s[0-9A-Z]{2}(?:[0-9A-Z]{3})?", countryCode);
return faker.regexify(regex);
```

`faker.country().countryCode2()` reads country.code2 from yaml which is stored lowercase. The test masks this as it has also been changed from regex to iban-commons-junit which appears to tolerates lowercase.
I only got this because I have a [test](https://github.com/mferretti/SeedStream/blob/main/generators/src/test/java/com/datagenerator/generators/semantic/DatafakerNewTypesTest.java#126) in [SeedStream](https://github.com/mferretti/SeedStream) that checks for conformant to the ISO produced BIC
```
  @Test
  void shouldGenerateBIC() {
    String bic = (String) generateWithContext("germany", "bic");
    // BIC/SWIFT code format: 8 or 11 alphanumeric characters
    assertThat(bic).isNotNull().isNotEmpty().matches("^[A-Z0-9]{8,11}$");
  }
```

spannm:
## Problem
`countryCode2()` returns lowercase ISO 3166-1 alpha-2 codes (e.g. `de`).
The BIC regex `[A-Z]{4}%s[0-9A-Z]{2}` requires uppercase, so generated BICs silently violated ISO 9362.

## Fix
Apply `.toUpperCase(Locale.ROOT)` before interpolating the country code into the regex pattern.

## Also included
- Javadoc for `Finance.bic()` and `Country` provider methods
- `iban-commons` bumped to 1.8.8
- Deprecated `assertThatIbanOf` replaced with `assertThatIban` in `FinanceTest`

Fixes #1853

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
         <commons-validator.version>1.10.1</commons-validator.version>
         <rgxgen.version>3.1</rgxgen.version>
         <google.javaformat.version>1.17.0</google.javaformat.version>
-        <iban-commons.version>1.8.7</iban-commons.version>
+        <iban-commons.version>1.8.8</iban-commons.version>
         <jacoco-maven-plugin.version>0.8.15</jacoco-maven-plugin.version>
         <junit.version>6.1.0</junit.version>
         <libphonenumber.version>9.0.33</libphonenumber.version>
--- a/src/main/java/net/datafaker/providers/base/Country.java
+++ b/src/main/java/net/datafaker/providers/base/Country.java
@@ -1,6 +1,10 @@
 package net.datafaker.providers.base;
 
 /**
+ * Provides methods for generating country-related data such as names, codes, capitals, and flags.
+ * <p>
+ * This provider is backed by the data defined in {@code country.yml}.
+ *
  * @since 0.8.0
  */
 public class Country extends AbstractProvider<BaseProviders> {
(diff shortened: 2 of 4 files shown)


kingthorin (approved):
Thanks!

PR #1858, opened 2026-06-25
Tweak release processing

kingthorin:
- Give the version updates workflow a proper name.
- Adjust release documentation to be up-to-date.
- Update release page generator to trim "What's Changed" heading from release description if present (the template already contains the heading in with the capitalization we want).


bodiam:
Approved ;)

PR #1860, opened 2026-06-27
Enable spotless index

snuyanzin:
Also more info about it
https://github.com/diffplug/spotless/tree/main/plugin-maven#incremental-up-to-date-checking-and-formatting

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,14 +29,27 @@ jobs:
     runs-on: ${{ matrix.runs-on }}
     steps:
       - uses: actions/checkout@v7
+        with:
+          # ratchetFrom needs history reachable to the base/before commit
+          fetch-depth: 0
       - name: Set up JDK ${{ matrix.java-version }}
         uses: actions/setup-java@v5
         with:
           java-version: ${{ matrix.java-version }}
           distribution: 'temurin'
           cache: maven
+      - name: Determine Spotless ratchet ref
+        shell: bash
+        run: |
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            ref="${{ github.event.pull_request.base.sha }}"
+          else
+            ref="${{ github.event.before }}"
+            [ "$ref" = "0000000000000000000000000000000000000000" ] && ref=""
+          fi
+          [ -n "$ref" ] && echo "RATCHET=-DratchetFrom=$ref" >> "$GITHUB_ENV"
       - name: Build with Maven
-        run: ./mvnw $mvn_options install -PnoGpg --file pom.xml
+        run: ./mvnw $mvn_options install $RATCHET -PnoGpg --file pom.xml
       - name: Run ProvidersDocsGenerator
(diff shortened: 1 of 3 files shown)


asolntsev:
@snuyanzin Do  understand correctly that it will help during local development, but would not be used in GitHub Actions?
Maybe we want to share the index between GitHub Actions builds as well?

snuyanzin:
yes, it is for local only

my current fear about ci: I don't know how to review it since it will appear in every PR after that

asolntsev:
We can create a separate branch like "main2" with the shared index, and create few PRs to "main2".
And we'll see how it works.

But this is optional, of course. 
Improving the local development is already a step forward.

snuyanzin:
looks like `spotless.ratchFrom` can do for ci what we need
```
    <ratchetFrom>origin/main</ratchetFrom>
```

https://github.com/diffplug/spotless/tree/main/plugin-maven#ratchet

.github/workflows/build.yml
```diff
@@ -29,14 +29,27 @@ jobs:
...
+      - name: Determine Spotless ratchet ref
+        shell: bash
+        run: |
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            ref="${{ github.event.pull_request.base.sha }}"
+          else
+            ref="${{ github.event.before }}"
+            [ "$ref" = "0000000000000000000000000000000000000000" ] && ref=""
+          fi
+          [ -n "$ref" ] && echo "RATCHET=-DratchetFrom=$ref" >> "$GITHUB_ENV"
```
snuyanzin, line 50: to make generic in case of multiple branches

snuyanzin:
test commit shows spotless keeps working https://github.com/datafaker-net/datafaker/actions/runs/28367702772/job/84037564425?pr=1860

bodiam:
Anything needed to merge this?

kingthorin:
Aren't all the Java files already formatted? In which case we shouldn't need the ratchet setup?

snuyanzin:
>Aren't all the Java files already formatted? In which case we shouldn't need the ratchet setup?

1. it is not only about java/also yaml which is configured in this project
this exactly the edge case.

In ideal world in main/master/other branch everything is formatted and having ratchet will run spotless only against files from the PR and fails if anything violates

However there might be unideal world: someone merges a commit with violated spotless rules files.
It means this build fails, however all others(non touching problem files) will pass even without a fix
Then there are 2 options
1.  another PR comes (the one which does not touch files violating spotless rules), only PR's files will be checked against spotless and nothing from previous violations will be detected
2. another PR comes (this time touching files already violating spotless). In this case it will fail on spotless

kingthorin:
Why wouldn't we simply fail PRs that aren't formatted properly?

Then there's only one condition that matters, things are always proper.

kingthorin:
Actually ignore that, I shouldn't be debating this. There's nothing wrong with it per se. it isn't what I'd do but that doesn't make it bad or wrong 😉

kingthorin (approved):

snuyanzin:
>Why wouldn't we simply fail PRs that aren't formatted properly?

this is satisfied, no problem with this

PR #1861, opened 2026-06-27
feat(stock): add ISIN generation and consolidate Luhn logic

spannm:
## What

Adds `Stock.isin()` for generating **ISO 6166-compliant ISINs** (country code + NSIN + Luhn check digit).

While implementing this, I noticed duplicated Luhn logic in `Finance` and `Code` and extracted it into a shared `LuhnCheckDigit` utility in `internal.helper`.

Class `Stock` was preferred over class `Finance` since an ISIN is a securities identifier, not a payment instrument.

## Changes

- `Stock.isin()` – new method, validated against `ISINValidator` in repeated test runs
- `LuhnCheckDigit` – new internal helper with `CharSequence` and `int[]` overloads, replaces inline Luhn loops in `Finance.creditCard()` and `Code.imei()`
- `FinanceTest` – uses `CreditCardValidator` instead of manual Luhn check; `creditCardWithType` converted to `@ParameterizedTest`

### DIFF
--- a/src/main/java/net/datafaker/internal/helper/LuhnCheckDigit.java
+++ b/src/main/java/net/datafaker/internal/helper/LuhnCheckDigit.java
@@ -1,0 +1,58 @@
+package net.datafaker.internal.helper;
+
+/**
+ * Calculates the check digit using the Luhn algorithm (Modulus 10).
+ *
+ * @since 2.8.0
+ */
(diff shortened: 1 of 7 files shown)


bodiam:
Nice PR, thanks! Approved!

PR #1863, opened 2026-06-28
Further Java Modernization

kingthorin:
- collect(Collectors.toList()) ➡️  toList()
- Use List/Set/Map.of where immutables are okay
- Leverage Records
- Use updated instanceof form to reduce assignments/intermediate storage
- Leverage var + local variable type inference
- Use text blocks to avoid concatenation, etc.

### DIFF
--- a/src/main/java/net/datafaker/service/FakeValuesContext.java
+++ b/src/main/java/net/datafaker/service/FakeValuesContext.java
@@ -7,18 +7,37 @@ import java.net.URL;
 import java.util.Locale;
 import java.util.Objects;
 
-record FakeValuesContext(SingletonLocale singletonLocale, String filename, String path, URL url) {
+class FakeValuesContext {
+    private final SingletonLocale singletonLocale;
+    private final String filename;
+    private final int filenameHashCode;
+    private final String path;
+    private final URL url;
+    private final int urlHashCode;
 
     private FakeValuesContext(Locale locale) {
-        this(SingletonLocale.get(locale), getFilename(locale), getFilename(locale), null);
+        this(locale, getFilename(locale), getFilename(locale), null);
     }
 
     private FakeValuesContext(Locale locale, URL url) {
-        this(SingletonLocale.get(locale), getFilename(locale), null, url);
+        this(locale, getFilename(locale), null, url);
     }
 
     private FakeValuesContext(Locale locale, String filename, String path) {
-        this(SingletonLocale.get(locale), filename, path, null);
+        this(locale, filename, path, null);
+    }
+
+    private FakeValuesContext(Locale locale, String filename, String path, URL url) {
+        this.singletonLocale = SingletonLocale.get(locale);
+        this.filename = filename;
+        this.path = path;
+        this.url = url;
+        this.filenameHashCode = filename == null ? 0 : filename.hashCode();
+        try {
+            this.urlHashCode = url == null ? 0 : url.toURI().hashCode();
+        } catch (URISyntaxException e) {
+            throw new RuntimeException("Invalid url: \"%s\"".formatted(url), e);
+        }
     }
 
     public static FakeValuesContext of(Locale locale) {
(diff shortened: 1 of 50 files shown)


src/main/java/net/datafaker/service/files/EnFile.java
```diff
@@ -3,20 +3,13 @@
 import java.util.List;
 import java.util.stream.Stream;
 
-public class EnFile {
+public record EnFile(String file, String path) {
```
snuyanzin: is it java 17?

src/main/java/net/datafaker/service/FakeValuesContext.java
```diff
@@ -7,37 +7,18 @@
...
 import java.util.Objects;
 
-class FakeValuesContext {
-    private final SingletonLocale singletonLocale;
-    private final String filename;
-    private final int filenameHashCode;
-    private final String path;
-    private final URL url;
-    private final int urlHashCode;
+record FakeValuesContext(SingletonLocale singletonLocale, String filename, String path, URL url) {
```
snuyanzin: is it java 17?

kingthorin: I may have been overly general in saying 17. 

I guess really it's 12-17.

I think records were 16.

snuyanzin: yes, double checked as well, thanks

src/main/java/net/datafaker/transformations/XmlTransformer.java
```diff
@@ -34,7 +34,7 @@ public String generate(Iterable<IN> input, Schema<IN, ?> schema) {
         }
 
         StringJoiner data = new StringJoiner(LINE_SEPARATOR);
-        for (IN in : input) {
+        for (var in : input) {
```
snuyanzin, line 37: one symbol longer?

src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
```diff
@@ -78,7 +78,7 @@ public Object apply(Object input, Schema<Object, ?> schema) {
             Consumer<Object> consumer = SCHEMA2CONSUMER.get(schema);
             if (consumer == null) {
                 final Field<Object, ?>[] fields = schema.getFields();
-                final Map<String, java.lang.reflect.Field> name2ClassField = Stream.of(clazz.getDeclaredFields()).collect(
+                var name2ClassField = Stream.of(clazz.getDeclaredFields()).collect(
```
snuyanzin: before that the var was `final`, after that is not...
not sure we should make such change

src/main/java/net/datafaker/transformations/XmlTransformer.java
kingthorin: Either way it's dealt with at compile time not runtime.

src/main/java/net/datafaker/transformations/JavaObjectTransformer.java
kingthorin: Good point, I'll check and revert those.

src/test/java/net/datafaker/TestStrings.java
```diff
@@ -0,0 +1,19 @@
...
+    /** Maps text-block newlines to the platform line separator used by transformers. */
+    public static String lines(String textBlock) {
+        return textBlock.replace("\n", System.lineSeparator());
+    }
+
+    /** Like {@link #lines(String)}, but strips one trailing platform line separator. */
+    public static String linesTrimTrailing(String textBlock) {
+        String result = lines(textBlock);
+        String sep = System.lineSeparator();
+        return result.endsWith(sep) ? result.substring(0, result.length() - sep.length()) : result;
```
snuyanzin: This might be dangerous especially for SQL containing `\n` inside values 
```sql
INSERT INTO "MyTable" ("N\nu\nm\nb\ne\nr\n", "Password") VALUES (1, '5');
```

kingthorin: Okay, will revert.

asolntsev:
@kingthorin While I generally like records, I must warn that after conversion a class to a recorder, all fields become publicly accessible.

E.g. in class `FakeValuesContext`, fields `singletonLocale`, `filenameHashCode`, `path`, `urlHashCode` didn't have a public getter. Now they all have public getters. Users might start using them, though we didn't plan it.

kingthorin:
@asolntsev & @snuyanzin got your comments all addressed I believe.

snuyanzin (approved):

PR #1869, opened 2026-07-01
Add `@FakerSource` annotation and `FakerSourceProvider` for dynamic test data generation

spannm:
**RFC — looking for feedback, not asking for a merge yet**

Right now, "generate N fake values and check a property" means `@RepeatedTest` plus a manual `faker.x().y()` call, and locale variants each get their own static `Faker` field. `@FakerSource` replaces that with one annotation:

@ParameterizedTest(name = "[{index}] {0}")
@FakerSource(code = "address#longitude", repeat = 10)
void testLongitude(String longitude) { ... }

It resolves a provider method via reflection (also handles locales, whole-provider injection, and combinatorial params). Test-only, lives under `src/test` — this is tooling for our own suite, not public API. I migrated `AddressTest`, `CodeTest`, `FinanceTest`, `HttpTest` as a proof of concept.

**Why it's nice:** the standout win is much better test feedback. Instead of one aggregate `@RepeatedTest` result where generated values are buried in a loop, each value gets its own named, inspectable row in the report — you can see exactly which input triggered a failure. Compare `irishAddress` before/after:

Before (`@RepeatedTest` — values hidden):
irishaddress RepeatedTest

After (`@FakerSource` — values visible by name):
irishaddress with FakerSource

Same story for `creditCardWithType` — before you only see the enum branch that ran, after you see the actual generated card number per case:

Before:
creditcard before

After:
creditcard with FakerSource

It also kills off a bunch of one-off static locale-`Faker` fields and `@MethodSource` generator methods.

**Tradeoff to weigh:** resolution is string-based reflection, so a rename like `Address.longitude()` isn't caught by the IDE while editing. That said, it *is* caught immediately at build time — an incorrectly annotated `@FakerSource` simply fails the test, same as any other broken test, so nothing wrong slips through CI unnoticed. There's also a fallback that re-instantiates the test class if the instance isn't ready yet during early JUnit lifecycle, and method resolution grabs the first match by name/type — worth watching if we ever get overloaded provider methods, though it hasn't been an issue in the proof-of-concept migration.

Before I migrate more tests: given the much better test feedback, is this a direction we want — with the understanding that any misconfiguration surfaces as a failing test, not a silent runtime issue?

### DIFF
--- a/src/test/java/net/datafaker/junit/FakerSource.java
+++ b/src/test/java/net/datafaker/junit/FakerSource.java
@@ -1,0 +1,177 @@
+package net.datafaker.junit;
+
+import org.junit.jupiter.params.provider.ArgumentsSource;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * {@code @FakerSource} is an {@link ArgumentsSource} for
+ * {@link org.junit.jupiter.params.ParameterizedTest} that generates test arguments
+ * by invoking a Datafaker provider method via reflection.
+ *
+ * <h3>Descriptor syntax</h3>
+ * <p>
+ * Full form with parameters:
+ * <pre>{@code
+ * code = "fieldName#methodName(ParamType1, ParamType2, ...)"
+ * }</pre>
+ * <p>
+ * Short form for no-argument methods (parentheses optional):
+ * <pre>{@code
+ * code = "fieldName#methodName"
+ * }</pre>
+ * <ul>
+ *   <li>{@code fieldName}  – field in the test class holding a Datafaker provider instance,
+ *       or a dot-separated path starting from a field (e.g. {@code "NL_FAKER.address"})</li>
+ *   <li>{@code methodName} – method to invoke on the resolved provider</li>
+ *   <li>{@code ParamType}  – simple or fully-qualified name of an {@code Enum},
+ *       primitive ({@code int}, {@code long}, {@code boolean}), or {@code String}</li>
+ * </ul>
+ *
+ * <h3>Parameter values</h3>
+ * <p>
+ * Use {@link #params()} for a single value per parameter position,
+ * or {@link #multiParams()} for multiple alternative values per position.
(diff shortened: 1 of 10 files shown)


asolntsev:
I would say that the public API for end users is `new Faker().address().longitude()`. And this API should be called from tests.

The fact that DF uses YML and cryptic strings like `address#longitude` is rather an implementation details. Most users should be aware of it.

spannm:
> I would say that the public API for end users is `new Faker().address().longitude()`. And this API should be called from tests.
> 
> The fact that DF uses YML and cryptic strings like `address#longitude` is rather an implementation details. Most users should be aware of it.

@asolntsev fair point — but this is test-only tooling under `src/test`, not a public API change. The `code` string just points at the same public call (`address().longitude()`), the same way `@MethodSource("someGenerator")` already points at a method by name today. Under the hood, `@FakerSource` still exercises the public API — it doesn't replace or hide it.

**The real payoff is the test report**: every generated value now shows up individually and by name (see before/after screenshots above), instead of being buried in a `@RepeatedTest` loop. And if the code string is wrong, the test just fails at build — no silent runtime surprise.

Happy to rename `code` if it undersells that it's just referencing existing public API.

P.S.: Updated PR description for clarity and added screenshots.

cc: @bodiam

bodiam:
Sorry, I'm closing this because of AI messages.

spannm:
That's disappointing - I put in a considerable amount of work and genuinely believe it adds value to the test suite. The screenshots speak for themselves.

bodiam:
The fact that you put a lot of work in it or not is no reason to add it to the code. Besides, you said you were looking for feedback, and now you have feedback. 

I don't think it's very useful at this stage, @asolntsev doesn't think it's super useful, and if you want, you can always create a small side library, datafaker-junit5 or so, which has this functionality. If it's used by other people, we can always reconsider to add it to the core, and make it part of the "public API" or so, but at this stage, I'm not convinced about this.

spannm:
> Sorry, I'm closing this because of AI messages.

Closing a PR for such reason seems unjustified.

Where did you spot "AI messages"?

You are a native English speaker (I assume), I'm not. For me English is a 3rd language.

How does the level of English in a PR even matter, as long as the intent is clear?

PR #1873, opened 2026-07-11
Replace iban-commons with jbanking

bodiam:
## Summary

- replace the test-only `iban-commons-junit` dependency with `jbanking`
- validate generated BICs and IBANs using jbanking APIs
- retain check-digit validation for Datafaker-supported countries that jbanking does not model
- remove the stale iban-commons source reference from `Finance`

## Why

The Finance tests depended on iban-commons-specific assertions and registry data. Moving to jbanking keeps independent validation coverage while avoiding coupling Datafaker's supported-country list to the validator library's registry.

## Impact

This changes test validation only; Finance's public API and generated values remain unchanged.

## Validation

- `./mvnw -Dtest=FinanceTest test` — 726 tests passed
- Maven Spotless checks passed as part of the test run


asolntsev (approved):
All fine, but I did't understand why `jbanking` dependency is preferred to `iban-commons-junit`.

bodiam:
iban-commons-junit is maintained by 1 person, doesn't have many stars, and caused our 2.6.0 to be broken.

asolntsev:
@bodiam Thanks. Just - it could be mentioned in commit message and PR description. The most important question is always "why?"

spannm:
I am saddened to say it feels that @bodiam is on some personal vendetta against me and my work.

spannm:
I honestly don't see a real reason to drop **iban-commons** - @asolntsev thank you for asking "why"

The GitHub Stars comparison isn't fair - jbanking's 13 years old, mine's 7 months. Both are niche libraries. Of course jbanking has more stars, ask again in 13 years.

> @bodiam: iban-commons-junit is maintained by 1 person

It's 2 actually. jbanking is maintained by 1 - if that even matters for a test-only dependency.

Also worth remembering: datafaker's own IBAN coverage went from 89 to 104 countries thanks to iban-commons (#1805)

Re a bug #1853 - yes, BIC generation broke in 2.6.0, after @kingthorin pulled me in I provided a fix very quickly and naturally I took the blame.

And for what it's worth, iban-commons beats jbanking on speed and memory, too (see [here for details](https://github.com/SpeedBankingDe/iban-commons-benchmarks))

Honestly, a unilateral move like this - made without a heads-up - is exactly what makes people stop wanting to contribute.

Kind regards, Markus

PR #1880, opened 2026-07-16
Add a Collection option

### LINKED ISSUE #1879: Add a Collection option
I use DataFaker `Option.nextElement(List<E>)` to get a random option from a list.
I'd like to get the same random option from a `java.util.Collection`.

I'd like to have a `Option.nextElement(Collection<E>`) available

My alternative solution is a project helper class converting the collection to a list before calling DataFaker

deluxe:
Add a Collection option to address #1879

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Options.java
+++ b/src/main/java/net/datafaker/providers/base/Options.java
@@ -1,5 +1,6 @@
 package net.datafaker.providers.base;
 
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -167,5 +168,20 @@ public class Options extends AbstractProvider<BaseProviders> {
      */
     public <E> E nextElement(List<E> list) {
         return list.get(faker.random().nextInt(list.size()));
+    }
+
+    /**
+     * Returns a random element from a collection.
+     *
+     * @param collection The collection to take a random element from.
+     * @param <E>        The type of the elements in the collection.
+     * @return A randomly selected element from the collection.
+     * @throws IllegalArgumentException if the collection is empty.
+     * @since 3.0.0
+     */
+    public <E> E nextElement(Collection<E> collection) throws IllegalArgumentException {
+        return collection.stream()
+        .skip(faker.random().nextInt(collection.size()))
+        .findFirst().orElseThrow(() -> new IllegalArgumentException("Collection is empty"));
     }
 }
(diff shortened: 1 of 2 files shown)


deluxe:
Hi there, first time contributor here! Please feel free ask any question.

src/main/java/net/datafaker/providers/base/Options.java
deluxe: I named this method `nextElement` to overload `nextElement(List<E>)` but both methods could be named `option` like others in this class (with `nextElement(List<E>)` kept for backward compatibility).

asolntsev: Generally it's totally fine.

But in case of huge collections, this is not the most effective solution. `List.copyOf(collection)` will create a copy of the whole collection every time.

A more effective solution would be using Iterator, something like this:

```java
int index = faker.random().nextInt(collection.size());
Iterator<T> it = collection.iterator();
for (int i = 0; i < index; i++) {
  iterator.next();
}
return iterator.next();
```

asolntsev (approved):
Approved - with one suggestion to improve performance.

src/main/java/net/datafaker/providers/base/Options.java
kingthorin: How about:

```java
// Use streams, skip to random location in collection, return the first item there
return collection.stream()
            .skip(faker.random().nextInt(collection.size()))
            .findFirst();
```

deluxe: > How about:
> 
> ```java
> return collection.stream()
>             .skip(faker.random().nextInt(collection.size()))
>             .findFirst();
> ```

Would it randomize the output in case of an ordered collection ?

kingthorin: Yes, similar to the other approaches it would return a random _n_ th element.

@asolntsev any feedback, stream vs iterator?

deluxe: Oh yes it's not a filter but a skip. My bad!

asolntsev: @kingthorin Sure, your solution with `stream().skip()` is more concise.

kingthorin (approved):
Thanks!

asolntsev (commented):
@deluxe Something is wrong with formatting in this PR.

Look at indentation here:

Image

This is how usually formatting looks like in Java code:

Image

kingthorin:
That's on me, I did run spotless:apply but I may have still broken something after.
Will check/fix.

deluxe:
You are too fast for me! If you prepare a v3, wouldn't it be the perfect time to harmonize the API by renaming `Options.nextElement(...)` to `Options.option(...)` ?

kingthorin:
We could. I'd suggest opening a new issue in which to discuss. I don't think anyone has brought it up before.

PR #1885, opened 2026-07-17
Refactor Options as Selection

deluxe:
Address #1883 by creating a whole new Selection class

### DIFF
--- a/src/main/java/net/datafaker/providers/base/Options.java
+++ b/src/main/java/net/datafaker/providers/base/Options.java
@@ -1,11 +1,7 @@
 package net.datafaker.providers.base;
 
-import java.util.Collection;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 /**
  * @since 0.8.0
@@ -19,7 +15,7 @@ public class Options extends AbstractProvider<BaseProviders> {
     }
 
     /**
-     * Returns a random element from an varargs.
+     * Returns a random element from a varargs.
      *
      * @param options The varargs to take a random element from.
      * @param <E>     The type of the elements in the varargs.
@@ -63,7 +59,7 @@ public class Options extends AbstractProvider<BaseProviders> {
     }
 
     /**
-     * Returns a random unique subset of elements from an varargs.
+     * Returns a random unique subset of elements from a varargs.
      *
      * @param size    The size of subset to return.
      * @param options The varargs to take a random element from.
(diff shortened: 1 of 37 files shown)


deluxe:
I first renamed `subset` as `manyOf` but settled for `setOf` to keep the `Set` reference.

deluxe:
I also chose to deprecate `random().nextEnum(Enum)` to direct users to `selection().oneOf(Enum)` and help `selection` discoverability.

With this much of a change, i did not mark any class or method as `forRemoval`.

asolntsev (approved):
If we all like the name `faker.selection().oneOf(...)`, then this PR is totally fine.

src/main/java/net/datafaker/providers/base/Options.java
```diff
@@ -1,67 +1,65 @@
...
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 /**
  * @since 0.8.0
+ * @deprecated Use {@link Selection} instead
  */
+@Deprecated(since = "3.0.0")
```
bodiam: Should this have the `forRemoval=true` flag on it? And possibly on a few other spots as well?

bodiam:
I left a tiny comment, not 100% sure about it, but I'd think it would make sense to mark it for removal?

deluxe:
> I left a tiny comment, not 100% sure about it, but I'd think it would make sense to mark it for removal?

I'm ok with it. Will do.

bodiam (approved):

bodiam:
Thanks! Looks good to me. Anyone else feedback?

src/main/java/net/datafaker/providers/base/ProviderRegistration.java
```diff
@@ -159,7 +159,8 @@ <PR extends ProviderRegistration, AP extends AbstractProvider<PR>> AP getProvide
     /**
      * @deprecated Use {@link #selection()} instead
      */
-    @Deprecated(since = "3.0.0")
+    @SuppressWarnings("removal")
+    @Deprecated(since = "3.0.0", forRemoval = true)
     default Options options() {
         return getProvider(Options.class, Options::new);
```
kingthorin, line 165: Instead of suppressing the removal warning why don't we just actually use selection internally?

deluxe: Because it will not return the same class (and Selection.class does not extend Options.class).

kingthorin: Right okay, sorry. Coffee needed 🙄

kingthorin (approved):

deluxe:
Is there something left for me to do for this PR?

PR #1887, opened 2026-07-21
Clean and automate native-image metadata

bodiam:
## Summary

- regenerate GraalVM reachability metadata without test classes, test dependencies, or test-only resources
- add a repeatable tracing-agent profile and generation script
- derive test-class exclusions from compiled test output and include resources from Datafaker plus runtime dependencies
- validate generated metadata before replacing the published file

## Root cause

The checked-in metadata was generated from the full test process without sufficient filtering. That allowed JUnit, Mockito, Byte Buddy, IDE, and test fixture metadata to leak into the library artifact. Resource discovery was also dependent on which paths happened to execute during tests.

## Validation

- `JAVA_HOME=/Users/erikpragt/.sdkman/candidates/java/current ./scripts/update-native-image-metadata.sh`
  - 20,445 tests; 0 failures; metadata validation passed
- GraalVM 25 native REST build with `--no-fallback`
- native executable smoke test: `GET /persons` returned 10 generated people, including phone and address data
- `git diff --check` and JSON syntax validation


kingthorin:
I have zero insight/knowledge on this

bodiam:
> I have zero insight/knowledge on this

yeah, same, but I made an native build using Codex, and it mentioned that our image is too big. This was the context Codex gave me:

```
  Datafaker worked natively, but its reachability metadata contained test-runner, JUnit,
  Mockito, Byte Buddy, and IDE entries. I cleaned that metadata and added src/test/java/net/
  datafaker/NativeImageMetadataTest.java:15 to prevent regression.

  Results:

  - Reachability metadata: 470 KB → 259 KB
  - Reflection entries: 856 → 372
  - Native-image warnings: removed
  - Native executable: 48.27 MB → 41.47 MB
  - Metadata regression test and formatting checks pass
  ```
  
I've built an new app now using  Datafaker as a native build with the updated version of Datafaker, and it works fine, so I think it's fine.

bodiam:
I'm just gonna assume you approved it. Thanks @kingthorin  ;-)

PR #1901, opened 2026-08-19
Fix domain/hostname generation to comply with RFC 1123

kingthorin:
Follow-up to PR #1757 where invalid domain names were being generated, particularly affecting non-ASCII locales and special characters in company names.

Problems fixed:
- Ampersands and special characters in company names now properly sanitized
- Non-ASCII locales now generate valid ASCII hostnames (via Punycode)
- Raw Unicode characters no longer leak into URLs
- Invalid Punycode output eliminated

Solution:
- New HostnameHelper class enforces RFC 1123 LDH (Letter-Digit-Hyphen) rules
- HostnameHelper applies IDN.toASCII() for Unicode input, enabling Punycode conversion
- Punycode prefix ("xn--") is preserved during sanitization
- Domain suffixes now route through toAsciiHostname() to preserve multi-label TLDs (e.g., "co.uk")
- All domain generation methods sanitize output to valid RFC 1123 hostnames
- FakerIDN guarantees valid ASCII output, never throws exceptions
- Test patterns updated to accept 2+ label domains

All generated domains are now valid for use with java.net.URL/URI constructors, support multi-label TLDs, and properly handle international domain names via Punycode.

This is a hardening fix with no breaking changes to the public API.

## Full disclosure

Claude code was leveraged in part during preparation of this PR.

### DIFF
--- a/src/main/java/net/datafaker/internal/helper/FakerIDN.java
+++ b/src/main/java/net/datafaker/internal/helper/FakerIDN.java
@@ -1,32 +1,109 @@
 package net.datafaker.internal.helper;
 
 import java.net.IDN;
+import java.util.Locale;
+import org.jspecify.annotations.Nullable;
 
 /**
- * Created by tshick on 10/30/16.
+ * Helper for converting domain names to ASCII using Punycode (IDN).
+ * <p>
+ * This class wraps {@link IDN#toASCII(String)} with fallback handling for edge cases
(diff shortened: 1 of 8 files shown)


kingthorin:
Sorry about the churn, it's now ready for review.

asolntsev (approved):

PR #1912, opened 2026-09-12
Add TemporalUnit overloads to timeAndDate(), deprecate TimeUnit variants

### LINKED ISSUE #1911: timeAndDate() should use java.time.temporal.TemporalUnit rather than java.util.concurrent.TimeUnit
**Is your feature request related to a problem? Please describe.**

I want to use: ```timeAndDate().past(10, TimeUnit.of(ChronoUnit.YEARS))```. However it's not possible to convert this:

```java.lang.IllegalArgumentException: No TimeUnit equivalent for Years```

The issue is: ```java.util.concurrent.TimeUnit``` is inappropriate for date operations, because the largest unit is "days".

**Describe the solution you'd like**

I propose to not use ```java.util.concurrent.TimeUnit``` here at all. Instead use ```java.time.temporal.TemporalUnit```, so that we can use the API like this:

```timeAndDate().past(10, ChronoUnit.YEARS)```

```timeAndDate().past(java.time.Duration duration)``` might also be a good idea.

**Additional context**

This proposal also affects other APIs, at least ```timeAndDate().future()```

LouisDeconinck:
Closes #1911

## What

`timeAndDate().past(10, TimeUnit.of(ChronoUnit.YEARS))` fails today because `java.util.concurrent.TimeUnit` maxes out at days. Per the approach agreed in the issue: the twelve `future`/`past` overloads taking `TimeUnit` are deprecated (`@Deprecated(since = "3.0.0", forRemoval = true)`, delegating via `unit.toChronoUnit()`), and twelve equivalents taking `java.time.temporal.TemporalUnit` are added — so `ChronoUnit.YEARS`, `MONTHS`, etc. now work.

Bounds are computed via `referenceDate.atZone(ZoneId.systemDefault()).plus/minus(atMost, unit)` so calendar units are handled correctly (`Instant.plus` would reject units above days); the random instant still comes from the existing `between(from, to)` helper, preserving the uniform distribution and strict-before/after invariants.

## Testing

- `./mvnw test -Dtest=TimeAndDateTest` — 1930 tests pass, including new `ChronoUnit.YEARS` bounds cases and TemporalUnit variants of the bounds/minimum/reference-date tests
- `FakerTest` + `DateAndTimeTest` pass — existing `TimeUnit` behavior and `#{date.past '15','SECONDS'}` expressions unaffected
- `./mvnw spotless:check` clean

Docs (`date-format.md`) examples updated to `ChronoUnit`.

### DIFF
--- a/docs/documentation/date-format.md
+++ b/docs/documentation/date-format.md
@@ -7,8 +7,8 @@ Since 1.2.0 Datafaker supports specifying of date formats for dates and timestam
 
     ```java
     Faker faker = new Faker();
-    System.out.println(faker.timeAndDate().future(1, TimeUnit.HOURS, "yyyy MM.dd mm:hh:ss"));
-    System.out.println(faker.timeAndDate().past(1, TimeUnit.HOURS, "yyyy-MM-dd mm:hh:ss"));
+    System.out.println(faker.timeAndDate().future(1, ChronoUnit.HOURS, "yyyy MM.dd mm:hh:ss"));
+    System.out.println(faker.timeAndDate().past(1, ChronoUnit.HOURS, "yyyy-MM-dd mm:hh:ss"));
     System.out.println(faker.timeAndDate().birthday(1, 99, "yyyy/MM/dd"));
     ```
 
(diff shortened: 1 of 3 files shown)


bodiam:
Looks good to me! Thanks for the contribution! Anyone else feedback?

kingthorin (approved):
LGTM

asolntsev (approved):
Looks good!

PR #1916, opened 2026-09-19
Keep float and byte values numeric in JSON

jakezwang:
`JsonTransformer` quotes `Float` and `Byte` values, unlike the other built-in numeric types. For example, a float price of `1.25f` becomes `"1.25"` instead of `1.25`.

Include bytes and finite floats in numeric output. Preserve the existing quoted representation of non-finite floats. Tests cover ordinary fields and collection elements, with numeric-looking strings and non-finite values as controls.

Validation: both new numeric regressions fail before the fix. All 21,355 unit tests pass on Java 17 and 21; the full Java 17 `install -PnoGpg` also passes integration tests, formatting, coverage reporting, and packaging. An independent public-API check parses generated JSON across numeric limits, signed zero, nested values, and unchanged scalar types.


asolntsev (approved):

src/main/java/net/datafaker/transformations/JsonTransformer.java
```diff
@@ -125,9 +125,11 @@ private static void value2String(@Nullable Object value, StringBuilder sb) {
         } else if (value instanceof Integer
                 || value instanceof Long
                 || value instanceof Short
+                || value instanceof Byte
                 || value instanceof BigInteger
                 || value instanceof Boolean
                 || value instanceof Double
```
snuyanzin, line 131: since you are here, can you please also add same for Double?
Inf/-Inf/NaN?

asolntsev: Sure, I will add.

asolntsev: @snuyanzin https://github.com/datafaker-net/datafaker/pull/1917

PR #1917, opened 2026-09-19
#1916 support infinite Doubles in JSON

asolntsev:
This is a small addition to https://github.com/datafaker-net/datafaker/pull/1916

* add quotes to Double values "NaN", "Infinity" and "-Infinity" in json
* add tests for other non-string values


snuyanzin (approved):
LGTM
thanks for fixing this

